PR #2, opened 2017-07-26
Middleware trait.

withoutboats:


nayato:
@carllerche Finagle does it with a couple of things: `Filter.andThen(factory: ServiceFactory): ServiceFactory` and `LoadBalancerFactory.newEndpoint(addr: Address): ServiceFactory` which is used to compose Service Factories and not just Services.

dmexe:
what is the current state of the middleware trait, it could able to merge?

carllerche:
I have been experimenting with this trait here: https://github.com/carllerche/tower-web/tree/master/src/middleware. There are some differences.

I will be working to upstream `Middleware` soonish.

carllerche:
Given the above comment, I am going to close this PR.

PR #27, opened 2017-11-08
add a CloneService helper for sharing a Service

seanmonstar:
I haven't spent much time optimizing, as I wanted to open this and see how you felt about the concept in general.

The idea is after having wrapped up a bunch of services, you could want to use the same service in a couple of places (such as different gRPC methods). This provides a safe way of being able to clone a `Service`, by tracking a `Task` for each clone.

### DIFF
--- a/tower-util/src/clone.rs
+++ b/tower-util/src/clone.rs
@@ -1,0 +1,86 @@
+//! Contains a `CloneService` type.
+
+use std::sync::{Arc, Mutex};
+use futures::{Async, Poll};
+use futures::task::{self, Task};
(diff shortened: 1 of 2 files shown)


carllerche:
Do you think this is still needed / a good idea? I have doubts, but I don't want to dig in more unless you think it is worth pursuing.

seanmonstar:
I suspect needing to clone a service will come up eventually, but I no longer need this for Conduit.

carllerche:
Closing for now, we can revisit the design of such a feature when a concrete need comes up.

PR #35, opened 2018-01-16
Make tower-balance load-aware

olix0r:
Previously, tower-balance used a fixed round-robin strategy for load
distribution.

This change makes `Balance` generic over its load metric and selection
strategy. The following new traits have been introduced to satisfy this:

- `tower_balance::Load` provides a concrete load metric (i.e. for a service);
- `tower_balance::Choose` provides a strategy for selecting a node;

There are two load balancing configurations supported out-of-the-box:

- `tower_balance::round_robin` provides a load-ignorant round-robin balancer.
- `tower_balance::power_of_two_choices` uses the Power of Two Choices to distribute requests to the least-loaded node. This should be used in conjunction with `tower_balance::load::WithPendingRequests` to decorate a `Discover` instance so that all services it produces implement `Load`

### DIFF
--- a/tower-balance/src/choose/mod.rs
+++ b/tower-balance/src/choose/mod.rs
@@ -1,6 +1,6 @@
 use ordermap::OrderMap;
 
-use Loaded;
+use Load;
 
 mod p2c;
 mod round_robin;
@@ -10,11 +10,14 @@ pub use self::round_robin::RoundRobin;
 
 /// A strategy for choosing nodes.
 pub trait Choose {
+    type Metric: PartialOrd;
 
     /// Returns the index of a ready endpoint.
     ///
     /// ## Panics
     ///
     /// If `nodes` is empty.
-    fn call<K, L: Loaded>(&mut self, nodes: &OrderMap<K, L>) -> usize;
+    fn call<K, L>(&mut self, nodes: &OrderMap<K, L>) -> usize
+    where
+        L: Load<Metric = Self::Metric>;
 }
--- a/tower-balance/src/choose/mod.rs
+++ b/tower-balance/src/choose/mod.rs
@@ -1,7 +1,5 @@
 use ordermap::OrderMap;
 
-use Load;
-
 mod p2c;
 mod round_robin;
 
@@ -9,15 +7,34 @@ pub use self::p2c::PowerOfTwoChoices;
 pub use self::round_robin::RoundRobin;
 
 /// A strategy for choosing nodes.
-pub trait Choose {
-    type Metric: PartialOrd;
-
+pub trait Choose<K, N> {
     /// Returns the index of a ready endpoint.
     ///
     /// ## Panics
     ///
     /// If `nodes` has fewer than 2 entries.
-    fn call<K, L>(&mut self, nodes: &OrderMap<K, L>) -> usize
-    where
-        L: Load<Metric = Self::Metric>;
+    fn choose(&mut self, nodes: Nodes<K, N>) -> usize;
+}
+
+/// Holds two or more loaded nodes.
+pub struct Nodes<'a, K: 'a, N: 'a>(&'a OrderMap<K, N>);
+
+impl<'a, K: 'a, N: 'a> Nodes<'a, K, N> {
+    pub(crate) fn new(inner: &'a OrderMap<K, N>) -> Self {
+        assert!(2 <= inner.len(), "Nodes must have 2 or more items");
+        Nodes(inner)
+    }
+
+    pub fn len(&self) -> usize {
+        self.0.len()
+    }
+}
+
+impl<'a, K: 'a, N: 'a> ::std::ops::Index<usize> for Nodes<'a, K, N> {
+    type Output = N;
+
+    fn index(&self, idx: usize) -> &Self::Output {
+        let (_, node) = self.0.get_index(idx).expect("out of bounds");
+        node
+    }
 }
--- a/tower-balance/src/choose/mod.rs
+++ b/tower-balance/src/choose/mod.rs
@@ -1,5 +1,4 @@
 use ordermap::OrderMap;
-use std::hash::Hash;
 
 use Loaded;
 
@@ -11,13 +10,10 @@ pub use self::round_robin::RoundRobin;
 
 /// A strategy for selecting nodes.
 pub trait Choose {
-    type Key: Hash + Eq;
-    type Loaded: Loaded;
-
     /// Returns the index of a ready endpoint.
     ///
     /// ## Panics
     ///
     /// If `ready` is empty.
-    fn call(&mut self, ready: &OrderMap<Self::Key, Self::Loaded>) -> usize;
+    fn call<K, L: Loaded>(&mut self, ready: &OrderMap<K, L>) -> usize;
 }
--- a/tower-balance/src/choose/mod.rs
+++ b/tower-balance/src/choose/mod.rs
@@ -10,7 +10,7 @@ pub use self::p2c::PowerOfTwoChoices;
 pub use self::round_robin::RoundRobin;
 
 /// A strategy for selecting nodes.
-pub trait Select {
+pub trait Choose {
     type Key: Hash + Eq;
     type Loaded: Loaded;
 
--- a/tower-balance/src/choose/mod.rs
+++ b/tower-balance/src/choose/mod.rs
@@ -16,7 +16,7 @@ pub trait Choose {
     ///
     /// ## Panics
     ///
-    /// If `nodes` is empty.
+    /// If `nodes` has fewer than 2 entries.
     fn call<K, L>(&mut self, nodes: &OrderMap<K, L>) -> usize
     where
         L: Load<Metric = Self::Metric>;
--- a/tower-balance/src/choose/mod.rs
+++ b/tower-balance/src/choose/mod.rs
@@ -7,34 +7,44 @@ pub use self::p2c::PowerOfTwoChoices;
 pub use self::round_robin::RoundRobin;
 
 /// A strategy for choosing nodes.
+// TODO hide `K`
 pub trait Choose<K, N> {
-    /// Returns the index of a ready endpoint.
+    /// Returns the index of a replica to be used next.
     ///
-    /// ## Panics
-    ///
-    /// If `nodes` has fewer than 2 entries.
-    fn choose(&mut self, nodes: Nodes<K, N>) -> usize;
+    /// `replicas` cannot be empty, so this function must always return a valid index on
+    /// [0, replicas.len()-1].
+    fn choose(&mut self, replicas: Replicas<K, N>) -> usize;
 }
 
-/// Holds two or more loaded nodes.
-pub struct Nodes<'a, K: 'a, N: 'a>(&'a OrderMap<K, N>);
-
-impl<'a, K: 'a, N: 'a> Nodes<'a, K, N> {
-    pub(crate) fn new(inner: &'a OrderMap<K, N>) -> Self {
-        assert!(2 <= inner.len(), "Nodes must have 2 or more items");
-        Nodes(inner)
+/// Creates a `Replicas` if there are two or more services.
+///
+pub(crate) fn replicas<K, S>(inner: &OrderMap<K, S>) -> Result<Replicas<K, S>, TooFew> {
+    if inner.len() < 2 {
+        return Err(TooFew);
     }
 
+    Ok(Replicas(inner))
+}
+
+/// Indicates that there were not at least two services.
+#[derive(Copy, Clone, Debug)]
+pub struct TooFew;
+
+/// Holds two or more services.
+// TODO hide `K`
+pub struct Replicas<'a, K: 'a, S: 'a>(&'a OrderMap<K, S>);
+
+impl<'a, K: 'a, S: 'a> Replicas<'a, K, S> {
     pub fn len(&self) -> usize {
         self.0.len()
     }
 }
 
-impl<'a, K: 'a, N: 'a> ::std::ops::Index<usize> for Nodes<'a, K, N> {
-    type Output = N;
+impl<'a, K: 'a, S: 'a> ::std::ops::Index<usize> for Replicas<'a, K, S> {
+    type Output = S;
 
     fn index(&self, idx: usize) -> &Self::Output {
-        let (_, node) = self.0.get_index(idx).expect("out of bounds");
-        node
+        let (_, service) = self.0.get_index(idx).expect("out of bounds");
+        service
     }
 }
--- a/tower-balance/src/choose/mod.rs
+++ b/tower-balance/src/choose/mod.rs
@@ -14,10 +14,10 @@ pub trait Choose {
     type Key: Hash + Eq;
     type Loaded: Loaded;
 
-    /// Returns the key of a ready endpoint.
+    /// Returns the index of a ready endpoint.
     ///
     /// ## Panics
     ///
     /// If `ready` is empty.
-    fn call<'s>(&mut self, ready: &'s OrderMap<Self::Key, Self::Loaded>) -> &'s Self::Key;
+    fn call(&mut self, ready: &OrderMap<Self::Key, Self::Loaded>) -> usize;
 }
--- a/tower-balance/src/choose/mod.rs
+++ b/tower-balance/src/choose/mod.rs
@@ -8,12 +8,13 @@ mod round_robin;
 pub use self::p2c::PowerOfTwoChoices;
 pub use self::round_robin::RoundRobin;
 
-/// A strategy for selecting nodes.
+/// A strategy for choosing nodes.
 pub trait Choose {
+
     /// Returns the index of a ready endpoint.
     ///
     /// ## Panics
     ///
-    /// If `ready` is empty.
-    fn call<K, L: Loaded>(&mut self, ready: &OrderMap<K, L>) -> usize;
+    /// If `nodes` is empty.
+    fn call<K, L: Loaded>(&mut self, nodes: &OrderMap<K, L>) -> usize;
 }
--- a/tower-balance/src/choose/p2c.rs
+++ b/tower-balance/src/choose/p2c.rs
@@ -1,8 +1,7 @@
-use ordermap::OrderMap;
 use rand::Rng;
-use std::marker::PhantomData;
 
-use {Load,  Choose};
+use {Load, Choose};
+use choose::Nodes;
 
 /// Chooses nodes using the [Power of Two Choices][p2c].
 ///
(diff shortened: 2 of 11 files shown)


tower-balance/src/load/pending_requests.rs
```diff
@@ -0,0 +1,134 @@
...
+}
+
+impl<S: Service> PollLoad for PendingRequests<S> {
+    fn poll_load(&mut self) -> Poll<Load, S::Error> {
+        try_ready!(self.poll_ready());
+
+        // Calculate a load that approaches 1.0 as the number of pending requests
+        // increases.
+        let mut load = 0.0;
+        for _ in 0..self.pending.len() {
```
olix0r: instead of doing this we should use an atomic counter, since we don't really need to be notified of changes to load.

tower-balance/src/load/mod.rs
```diff
@@ -0,0 +1,52 @@
...
+
+pub use self::constant::{Constant, WithConstant};
+pub use self::pending_requests::{PendingRequests, WithPendingRequests};
+
+/// A `Service` that has a load metric when it is ready.
+pub trait PollLoad: Service {
+    /// Returns a `Load` metric for the underlying `Service`, if the service is ready.
+    ///
+    /// Must not return a value unless `Service::poll_ready` would return `Async::Ready`.
+    fn poll_load(&mut self) -> Poll<Load, Self::Error>;
```
olix0r: Alternatively, this could just be `fn load(&self) -> Load` and say that it's fatal to call it when `poll_ready` is not ready.

tower-balance/src/lib.rs
```diff
@@ -26,166 +53,158 @@ pub enum Error<T, U> {
...
+    type Error = Error<<L::PollLoad as Service>::Error, D::DiscoverError>;
+    type Future = ResponseFuture<<L::PollLoad as Service>::Future, D::DiscoverError>;
 
-        loop {
-            let change = match try!(self.discover.poll()) {
-                Async::Ready(change) => change,
-                Async::NotReady => return Ok(()),
-            };
+    fn poll_ready(&mut self) -> Poll<(), Self::Error> {
+        try_ready!(self.poll_discover());
```
olix0r: we need to clear `ready_endpoint` before polling discover (in case there's a ready endpoint that is removed).

tower-balance/src/select/mod.rs
```diff
@@ -0,0 +1,28 @@
...
+use PollLoad;
+
+mod p2c;
+mod round_robin;
+
+pub use self::p2c::PowerOfTwoChoices;
+pub use self::round_robin::RoundRobin;
+
+/// A strategy for selecting nodes.
+pub trait Select {
```
hawkw: This is _definitely_ bikeshedding but I'm not sure if `Select` is the best name or not --- at the first glance, when I saw the `Select` field in `Balance`, I thought it was a `stream::Select` or `future::Select` or something. I think maybe a name like `SelectPolicy` or `SelectStrategy` would better communicate what this trait is (or even just `select::Policy`/`select::Strategy`?)

tower-balance/src/load/pending_requests.rs
```diff
@@ -0,0 +1,134 @@
...
+
+// ===== impl ResponseFuture =====
+
+impl<S: Service> Future for ResponseFuture<S> {
+    type Item = S::Response;
+    type Error = S::Error;
+
+    fn poll(&mut self) -> Poll<Self::Item, Self::Error> {
+        let item = try_ready!(self.inner.poll());
+        if let Some(tx) = self.done.take() {
```
carllerche: This path is skipped if the response errors. I would think you would want to flag the request as done in the event that it errors.

tower-balance/src/select/mod.rs
carllerche: Honestly, we may want to rename `Balance` -> `LoadBalancer`. That would free up the trait to be named `Balance`.

/disclaimer This is just an idea and not a change request.

tower-balance/src/select/mod.rs
```diff
@@ -0,0 +1,28 @@
...
+    type Key: Hash + Eq;
+    type Service: PollLoad;
+
+    /// Returns the key of a ready endpoint.
+    ///
+    /// ## Panics
+    ///
+    /// If `endpoints` is empty.
+    fn poll_next_ready<'s>(
```
carllerche: Maybe just name this `poll_next`.

tower-balance/src/select/round_robin.rs
```diff
@@ -0,0 +1,64 @@
...
+impl<K, S> Select for RoundRobin<K, S>
+where
+    K: Hash + Eq,
+    S: PollLoad,
+{
+    type Key = K;
+    type Service = S;
+
+    fn poll_next_ready<'s>(&mut self, iter: IterMut<'s, K, S>) -> Poll<&'s Self::Key, S::Error> {
+        let mut nodes = iter.collect::<Vec<_>>();
```
carllerche: This is pretty unfortunate. I'd probably want to try to figure out a way to avoid the unnecessary overhead.

The only option that comes to mind right now is to pass in `endpoints: &[Entry<S>]` where `Entry` wraps the service and tracks if it is ready or not.

tower-balance/src/lib.rs
```diff
@@ -1,21 +1,48 @@
...
+    select: S,
+
+    /// Holds the key of an endpoint that has been selected to be used next.
+    ready_endpoint: Option<D::Key>,
+
+    /// Holds all possibly-available endpoints (i.e. from `discover`).
+    endpoints: OrderMap<D::Key, L::PollLoad>,
+
+    /// Newly-added endpoints that have not yet become ready.
+    new_endpoints: OrderMap<D::Key, L::PollLoad>,
```
carllerche: Could you explain why you opted to split out `new_endpoints`?

Also, if there are two maps, why not have it be `ready_endpoints` vs. `not_ready_endpoints`, i.e., remove endpoints from `endpoints` when `poll_ready` returns `NotReady`?

tower-balance/src/lib.rs
```diff
@@ -26,166 +53,158 @@ pub enum Error<T, U> {
...
-                    // given `key` in the set?
-                    self.endpoints.entry(key).or_insert(val);
-                }
-                Remove(key) => {
-                    // TODO: What to do if there is no entry?
-                    self.endpoints.remove(&key);
-                }
-            }
+        let nodes = self.endpoints.iter_mut();
+        let key =  try_ready!(self.select.poll_next_ready(nodes).map_err(Error::Inner));
```
carllerche: I wonder if this could be called from `call`.

For this to be true, `poll_next_ready` would actually not be able to be async. It would have to always return a service.

Would it be sane for the load balancer to say that it is ready if at least one endpoint is ready? In this case, `poll_ready` would only need to iterate the list of not-ready endpoints, polling them. If any are ready, then the load balancer is ready itself. This would also simplify the implementation of `Select`.

tower-balance/src/load/mod.rs
olix0r: Yeah, you're right that load can be called when not ready...

tower-balance/src/load/mod.rs
```diff
@@ -0,0 +1,52 @@
...
+use tower::Service;
+
+mod constant;
+mod pending_requests;
+
+pub use self::constant::{Constant, WithConstant};
+pub use self::pending_requests::{PendingRequests, WithPendingRequests};
+
+/// A `Service` that has a load metric when it is ready.
+pub trait PollLoad: Service {
```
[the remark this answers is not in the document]
olix0r: Mostly to signal intent (and to yoink the service's Error type).  I think it'd be fair to relax this.

tower-balance/src/lib.rs
olix0r: I opted to split out new_endpoints because, once we get an update from service discovery, we basically _know_ the endpoint isn't ready, so if we offer it up to the selector, it has a very low likelihood of being selected.

However, after this, the balancer is no longer (and should not be) responsible for polling all nodes for readiness (let's assume we'll end up with 100s of nodes in a LB)-- we want the selector to be responsible for this polling so it's polling the fewest nodes needed to make a balancing determination.

I think we can probably remove `new_endpoints`, as there isn't a strong motivation for it.

tower-balance/src/lib.rs
olix0r: For this to be sync, we'd have to know the readiness of all nodes. I don't think we should expect the balancer to manage this totally (with 10^2-10^3 nodes).  Rather, the node selection strategy manages this polling so it only polls the nodes for which it actually needs state.

tower-balance/src/load/mod.rs
```diff
@@ -0,0 +1,52 @@
...
+    /// Returns a `Load` metric for the underlying `Service`, if the service is ready.
+    ///
+    /// Must not return a value unless `Service::poll_ready` would return `Async::Ready`.
+    fn poll_load(&mut self) -> Poll<Load, Self::Error>;
+}
+
+/// Wraps services so that they implement `PollLoad`.
+pub trait WithLoad {
```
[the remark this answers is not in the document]
olix0r: I disagree that the load metric should be coupled with discovery -- this seems like something that is strictly bound to the balancer.

I could see removing this, but it seems like a generally good thing to keep discovery and load completely orthogonal.

Does Discover implement `map` etc out-of the-box?

tower-balance/src/select/round_robin.rs
olix0r: We need to figure out what is responsible for polling readiness ;)  This is only possible if the balancer is responsible for managing the readiness state of all nodes -- at that point the balancer can just maintain a total ordering of all nodes weights; and we wouldn't even need to do P2C since we have an ordered heap.

This isn't the craziest approach, but i worry it won't stand up as we support more nodes.

carllerche:
Ok, so `Service` guarantees:

I don't think a response future should ever depend on `Service::poll_ready` being called. However, the `Reconnect` middleware *does* require `poll_ready` to be called in order to process the reconnect.

So, if `Service::poll_ready` is "randomly" called, you could get into the situation where a client socket tries to reconnect because `poll_ready` is called, opens the TCP socket, but the `poll_ready` function is not called again in a timely fashion, so the TCP socket gets disconnected, which could lead to a waste of resources.

I have not formulated a strong opinion regarding the load balancer calling `Service::poll_ready` on all inner services eagerly. I see there being pros / cons either way. I'm leaning towards having `LoadBalancer::poll_ready` iterating all services and calling `poll_ready` on them, but I am not sure.

olix0r:
I'm feeling much better about this PR after addressing the feedback fully.

WithLoad has been removed and replaced by Discover implementations that add load metrics.

The `Load` struct has been removed because it seemed to indicate that various types of load should be comparable (i.e. how do we normalize # pending requests with latency? can't be done, really).  So now, the load type is not common across implementations.  The `Load` trait has a type parameter indicting what type it generates and `Choose` (formerly `Select`) also has such a type parameter.

tower-balance/src/load/pending_requests.rs
```diff
@@ -0,0 +1,136 @@
...
+    pending: Arc<AtomicUsize>,
+}
+
+/// Wraps a response future so that it is tracked by `PendingRequests`.
+///
+/// When the inner future completes, either with a value or an error, `pending` is
+/// decremented.
+pub struct ResponseFuture<S: Service> {
+    inner: S::Future,
+    pending: Option<Handle>,
```
adleong: Why use a Handle with Drop instead of just putting the `Arc<AtomicUsize>` directly on the ResponseFuture struct here?

tower-balance/src/lib.rs
```diff
@@ -26,166 +83,185 @@ pub enum Error<T, U> {
...
-    fn update_endpoints(&mut self) -> Result<(), T::DiscoverError> {
-        use tower_discover::Change::*;
+        let (key, svc) = self.ready.swap_remove_index(idx).expect("invalid ready index");
+        self.not_ready.insert(key, svc);
+        Some(Ok(Async::NotReady))
+    }
 
+    /// Chooses the next service to which a request will be dispatched.
+    ///
+    /// Ensures that .
```
olix0r: Incomplete

tower-balance/src/choose/round_robin.rs
```diff
@@ -0,0 +1,25 @@
...
+pub struct RoundRobin {
+    /// References the index of the next node to be polled.
+    pos: usize,
+}
+
+impl Choose for RoundRobin {
+    type Metric = ();
+
+    fn call<K, L: Load>(&mut self, ready: &OrderMap<K, L>) -> usize {
+        assert!(2 <= ready.len(), "must choose over 2 or more ready nodes");
```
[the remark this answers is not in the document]
olix0r: because that's dictated by the interface contract. there's no reason to call `choose` on fewer than 2 nodes. this is enforced in Balance.

tower-balance/src/choose/p2c.rs
```diff
@@ -0,0 +1,72 @@
...
+        if ready.len() == 2 {
+            // Avoid looping if there are only two nodes.
+            if self.rng.gen::<bool>() {
+                return (0, 1);
+            } else {
+                return (1, 0);
+            }
+        }
+
+        let idx0 = self.rng.gen::<usize>() % ready.len();
```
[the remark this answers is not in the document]
olix0r: That's much less readable to me.

Also it uses two rng calls, which shouldn't be necessary.

tower-balance/src/load/pending_requests.rs
olix0r: You could do that, but this enforces the desired behavior more correctly (leveraging the type system).

Once `PendingRequests` increments the counter, it produces a Handle.  When the Handle is dropped, no matter how, this ensures that we decrement the counter.  This handles the case where the future is polled as complete and we intentionally drop it.  However, it also handles the case when the Future is dropped before it completes.  We could achieve this by implementing `Drop for ResponseFuture`, but the current implementation is a bit more rigorous imo.

tower-balance/src/choose/p2c.rs
```diff
@@ -0,0 +1,69 @@
...
+}
+
+impl<M: PartialOrd, R: Rng> Choose for PowerOfTwoChoices<M, R> {
+    type Metric = M;
+
+    /// Chooses two distinct nodes at random and compares their load.
+    ///
+    /// Returns the index of the lesser-loaded node.
+    fn call<K, L: Load<Metric = M>>(&mut self, ready: &OrderMap<K, L>) -> usize {
+        assert!(2 <= ready.len(), "must choose over 2 or more ready nodes");
```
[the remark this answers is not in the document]
olix0r: it can't be triggered if it's invoked by Balance.  i'll make it a debug_assert

tower-balance/src/lib.rs
```diff
@@ -26,166 +83,185 @@ pub enum Error<T, U> {
...
+        self.promote_to_ready()?;
+
+        // Choose the next service to be used by `call`.
+        self.choose_and_poll_ready()
     }
 
-    fn inc_pos(&mut self) {
-        self.pos = (self.pos + 1) % self.endpoints.len();
+    fn call(&mut self, request: Self::Request) -> Self::Future {
+        let idx = self.chosen_ready_index.take().expect("not ready");
```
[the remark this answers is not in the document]
olix0r: It could, but isn't it a contract violation to invoke `call` without previously calling `poll_ready`?

tower-balance/src/lib.rs
```diff
@@ -26,166 +83,185 @@ pub enum Error<T, U> {
...
+    type Error = Error<<D::Service as Service>::Error, D::DiscoverError>;
+    type Future = ResponseFuture<<D::Service as Service>::Future, D::DiscoverError>;
+
+    /// Prepares the balancer to process a request.
+    ///
+    /// When `Async::Ready` is returned, `chosen_ready_index` is set with a valid index
+    /// into `ready` referring to a `Service` that is ready to disptach a request.
+    fn poll_ready(&mut self) -> Poll<(), Self::Error> {
+        // Clear before `ready` is altered.
+        self.chosen_ready_index = None;
```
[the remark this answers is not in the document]
olix0r: This is intended. If it's called again, we might update `ready` below, and this index would no longer be reliable (because the ordermap may change its order).

tower-balance/src/choose/p2c.rs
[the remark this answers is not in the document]
olix0r: not when invoked from `Balance`.  With your suggested change, we'll make it an assertion on the type that Entries has 2 or more items

tower-balance/src/choose/mod.rs
```diff
@@ -0,0 +1,23 @@
...
+/// A strategy for choosing nodes.
+pub trait Choose {
+    type Metric: PartialOrd;
+
+    /// Returns the index of a ready endpoint.
+    ///
+    /// ## Panics
+    ///
+    /// If `nodes` has fewer than 2 entries.
+    fn call<K, L>(&mut self, nodes: &OrderMap<K, L>) -> usize
```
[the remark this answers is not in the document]
olix0r: > Don't expose OrderMap directly, instead wrap it in a new type.

Yeah this is a good idea.

> Ideally, the K wouldn't be exposed. Unfortunately, that would be a bit annoying (though, possible) to hide.

Have an example handy?

> Having generics on the fn means that it can't be used as a trait object.

What does this mean, practically?

olix0r:
> Also, it seems that you still opt not to poll all not-ready services in Balance::poll_ready. Would you mind sharing the thought process?

The meat of poll_ready is

```
       // Update `not_ready` and `ready`.
       self.update_from_discover()?;
       self.promote_to_ready()?;

       // Choose the next service to be used by `call`.
       self.choose_and_poll_ready()
```

`promote_to_ready polls all of `not_ready` and promotes nodes to `ready`

tower-balance/src/lib.rs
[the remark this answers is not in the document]
carllerche: It isn’t a violation per se but there is no guarantee it will succeed.

carllerche:
@olix0r Right, but services are never moved from "ready" to "not ready" as far as I can tell?

olix0r:
@carllerche We poll ready nodes in two situations:

1. Before they are used.
2. After they are used.

In the first case, `poll_ready_index` ensures that the node is ready and moves it to not_ready if it is not.  We do this check [before setting chosen_ready_index](https://github.com/tower-rs/tower/blob/962bca83125375ffb5766e983f8fbc0dc24e16f5/tower-balance/src/lib.rs#L208).

In the second case, as soon as poll_ready is called, we [determine if the last-used index is still ready](https://github.com/tower-rs/tower/blob/962bca83125375ffb5766e983f8fbc0dc24e16f5/tower-balance/src/lib.rs#L238).  It is moved to not_ready if it is not still ready.

PR #39, opened 2018-01-25
Add an examples that demonstrates p2c rr behavior

olix0r:
The new _demo_ example sends a million simulated requests through each load balancer configuration and records the observed latency distributions.

Furthermore, this fixes a critical bug in `Balancer`, where we did not properly iterate through not-ready nodes.

```
:; cargo run --example p2c
   Compiling tower-balance v0.1.0 (file:///Users/ver/b/rs/tower/tower-balance)
    Finished dev [unoptimized + debuginfo] target(s) in 2.55 secs
     Running `/Users/ver/b/rs/tower/target/debug/examples/p2c`
p2c samples: 1000000
p2c p50:  454
p2c p90:  498
p2c p95:  522
p2c p99:  1984
p2c p999: 2034
rr samples: 1000000
rr p50:  457
rr p90:  1944
rr p95:  1971
rr p99:  2006
rr p999: 2038
```

PR #40, opened 2018-01-25
Fix potential integer overflow in Balance

hawkw:
An usize overflow can occur in `Balance::promote_to_ready` when `self.not_ready` has length 0.

See my comment here: https://github.com/tower-rs/tower/pull/39#discussion_r163967979

### DIFF
--- a/tower-balance/src/lib.rs
+++ b/tower-balance/src/lib.rs
@@ -151,7 +151,7 @@ where
         debug!("promoting to ready: {}", n);
         // Iterate through the not-ready endpoints from right to left to prevent removals
         // from reordering services in a way that could prevent a service from being polled.
-        for idx in (0..n-1).rev() {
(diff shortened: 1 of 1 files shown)


tower-balance/src/lib.rs
```diff
@@ -151,7 +151,7 @@ where
         debug!("promoting to ready: {}", n);
         // Iterate through the not-ready endpoints from right to left to prevent removals
         // from reordering services in a way that could prevent a service from being polled.
-        for idx in (0..n-1).rev() {
```
olix0r: i think this would be slightly clearer as:

```yaml
if n == 0 {
  return Ok(());
}
for idx in (0..n-1).rev() { 
  ...
}
```

it makes the intent a bit clearer...

hawkw: Sorry, I don't follow? This is the line that was removed.

PR #41, opened 2018-01-26
Fix off-by-one error in promote_to_ready

hawkw:
The `..` syntax creates a _half-open_ range (see https://doc.rust-lang.org/std/ops/struct.Range.html), so all that messing about with `n-1` in #39 and #40 was never actually necessary. This actually fixes the Conduit test I mentioned in https://github.com/tower-rs/tower/pull/39#discussion_r163967979; it no longer hangs.


hawkw:
I left in the trace logging that I used to fix this, since I think it'll continue to be useful to actually know what's happening inside the loop.

PR #42, opened 2018-01-26
Test `Balance::poll_ready`

olix0r:
Test `Balance::poll_ready` by creating a random number of services, each
of which must be polled a random number of times before becoming ready.
As the balancer is polled, the test ensures that it does not become
ready prematurely and that services are promoted from not_ready to
ready.

`Balance::num_ready()` and `Balance::is_ready()` have been added to
expose the number of ready services, as well as `Balance::num_not_ready()`
to expose the number of pending services.

### DIFF
--- a/tower-balance/src/lib.rs
+++ b/tower-balance/src/lib.rs
@@ -107,21 +107,29 @@ where
     }
 
     /// Returns true iff there are ready services.
-    pub fn is_ready(&self) -> bool {
+    ///
+    /// This is not authoritative and is only useful after `poll_ready` has been called.
+    fn is_ready(&self) -> bool {
         !self.ready.is_empty()
     }
 
     /// Returns true iff there are no ready services.
+    ///
+    /// This is not authoritative and is only useful after `poll_ready` has been called.
     pub fn is_not_ready(&self) -> bool {
         self.ready.is_empty()
     }
 
     /// Counts the number of services considered to be ready.
+    ///
+    /// This is not authoritative and is only useful after `poll_ready` has been called.
     pub fn num_ready(&self) -> usize {
         self.ready.len()
     }
 
     /// Counts the number of services not considered to be ready.
+    ///
+    /// This is not authoritative and is only useful after `poll_ready` has been called.
     pub fn num_not_ready(&self) -> usize {
         self.not_ready.len()
     }
--- a/tower-balance/src/lib.rs
+++ b/tower-balance/src/lib.rs
@@ -107,18 +107,22 @@ where
     }
 
     /// Returns true iff there are ready services.
-    pub fn is_empty(&self) -> bool {
+    pub fn is_ready(&self) -> bool {
+        !self.ready.is_empty()
+    }
+
+    /// Returns true iff there are no ready services.
+    pub fn is_not_ready(&self) -> bool {
         self.ready.is_empty()
     }
 
     /// Counts the number of services considered to be ready.
-    pub fn len(&self) -> usize {
+    pub fn num_ready(&self) -> usize {
         self.ready.len()
     }
 
-
     /// Counts the number of services not considered to be ready.
-    pub fn not_ready_len(&self) -> usize {
+    pub fn num_not_ready(&self) -> usize {
         self.not_ready.len()
     }
 
@@ -394,15 +398,18 @@ mod tests {
                     }
                 }
 
-                if balancer.len() != ready {
+                if balancer.num_ready() != ready {
                     return TestResult::failed();
                 }
 
-                if balancer.not_ready_len() != pending {
+                if balancer.num_not_ready() != pending {
                     return TestResult::failed();
                 }
 
-                if balancer.is_empty() != (ready == 0) {
+                if balancer.is_ready() != (ready > 0) {
+                    return TestResult::failed();
+                }
+                if balancer.is_not_ready() != (ready == 0) {
                     return TestResult::failed();
                 }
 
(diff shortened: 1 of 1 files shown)


tower-balance/src/lib.rs
```diff
@@ -106,6 +106,22 @@ where
         }
     }
 
+    /// Returns true iff there are ready services.
+    pub fn is_empty(&self) -> bool {
```
carllerche: Can this be named `is_ready(&self)`? Does this function returning `false` imply that `poll_ready` will return `Ready`?

tower-balance/src/lib.rs
```diff
@@ -106,6 +106,22 @@ where
...
+        self.ready.is_empty()
+    }
+
+    /// Counts the number of services considered to be ready.
+    pub fn len(&self) -> usize {
```
carllerche: This isn't _really_ a collection structure, which `len` implies.

Maybe `num_ready()`?

tower-balance/src/lib.rs
```diff
@@ -106,6 +106,22 @@ where
...
+        self.ready.len()
+    }
+
+
+    /// Counts the number of services not considered to be ready.
+    pub fn not_ready_len(&self) -> usize {
```
carllerche: If `len` is renamed, this should be updated to match that pattern.

tower-balance/src/lib.rs
hawkw: +1 to `num_ready()`

tower-balance/src/lib.rs
```diff
@@ -283,3 +299,137 @@ impl<F: Future, E> Future for ResponseFuture<F, E> {
...
+    quickcheck! {
+        /// Creates a random number of services, each of which must be polled a random
+        /// number of times before becoming ready. As the balancer is polled, ensure that
+        /// it does not become ready prematurely and that services are promoted from
+        /// not_ready to ready.
+        fn poll_ready(service_tries: Vec<usize>) -> TestResult {
+            // Stores the number of pending services after each poll_ready call.
+            let mut pending_at = Vec::new();
+
+            let disco = {
```
hawkw: This is really minor, so TIOLI, but I'm wondering if the construction of the `ReluctantDisco` could be moved out of the actual test function?

tower-balance/src/lib.rs
```diff
@@ -283,3 +299,137 @@ impl<F: Future, E> Future for ResponseFuture<F, E> {
...
+                    Err(_) => return TestResult::error("poll_ready failed"),
+                    Ok(p) => {
+                        if p.is_ready() != (ready > 0) {
+                            return TestResult::failed();
+                        }
+                    }
+                }
+
+                if balancer.len() != ready {
+                    return TestResult::failed();
```
hawkw: Here you make a bunch of boolean assertions and fail the test if any of them is true. If this test fails, I think it would be nice to know *which* assertion was the one that failed, to aid in debugging any regressions. 

I'd consider using `TestResult::error`, which AFAIK is the only way to add a message to a QuickCheck test failure. I know it's technically supposed to represent a test failure "due to a runtime error", but my understanding is that this doesn't actually change anything except for the message that QuickCheck will print on failure (let me know if I'm incorrect!).

hawkw: Alternatively, I suppose you could replace these with `assert!()`, though I'm not sure how that interacts with QuickCheck off the top of my head --- will a panic stop quickcheck from shrinking etc, or is it just treated as a failure?

tower-balance/src/lib.rs
```diff
@@ -283,3 +299,137 @@ impl<F: Future, E> Future for ResponseFuture<F, E> {
...
+                ReluctantDisco(changes)
+            };
+            pending_at.push(0);
+
+            let mut balancer = Balance::new(disco, choose::RoundRobin::default());
+
+            let services = service_tries.len();
+            let mut next_pos = 0;
+            for pending in pending_at.iter().map(|p| *p) {
+                assert!(pending <= services);
```
hawkw: Is there any particular reason for using an assertion here and `if ... { return TestResult::failed() }` elsewhere in the test?

tower-balance/src/lib.rs
hawkw: Another option is to consider breaking this test into multiple tests that each only check for a _specific_ property --- my understanding is that this is considered a more idiomatic pattern in property-based testing. Of course if the test already takes a long time to run, this may not be ideal either.

tower-balance/src/lib.rs
olix0r: because this is expressing an invariant on the test logic, not validating app behavior

tower-balance/src/lib.rs
olix0r: I agree that narrower tests are desirable. This is better than what we had.  We should add more tests.

olix0r: fwiw, `error` is meant to indicate runtime errors. `failed` is the "right" thing to use here.

tower-balance/src/lib.rs
```diff
@@ -106,6 +106,26 @@ where
         }
     }
 
+    /// Returns true iff there are ready services.
+    pub fn is_ready(&self) -> bool {
+        !self.ready.is_empty()
```
carllerche: Looking through the `impl` again, it seems like the check also needs to factor in `dispatched_ready_index`.

Specifically, if `dispatched_ready_index` is `Some`, it is possible for an entry in `ready` to not *actually* be empty.

I would also add a comment stating that even if `is_ready` returns `true`, you still have to call `poll_ready` (otherwise, panic).

olix0r: yeah... these calls only make sense after poll_ready is called. we don't _need_ to expose these publicly, even, so perhaps better to remove?  i'll clarify the comments at least, though.

PR #45, opened 2018-01-29
Support weighted balancing

olix0r:
In order to implement red-line testing, blue-green deployments, and
other operational use cases, many service discovery and routing schemes
support weighting.

In this iteration, we provide a decorator type, `Weighted`, that may be
used to wrap load-bearing services to alter their load according to a
weight.  The `Weighted` type may also be used to wrap `Discover`
implementations, in which case it will wrap all new services with
`Weighted`.

The demo has been updated as well:
```
p2c + pending requests
  p50	152
  p90	193
  p95	201
  p99	1947
  p999	2013
p2c + pending requests + weighted
  p50	151
  p90	193
  p95	202
  p99	230
  p999	1972
round robin
  p50	156
  p90	1929
  p95	1966
  p99	2000
  p999	2027
```

### DIFF
--- a/tower-balance/examples/demo.rs
+++ b/tower-balance/examples/demo.rs
@@ -23,11 +23,12 @@ use tower::Service;
 use tower_balance::*;
 use tower_discover::{Change, Discover};
 
+#[derive(Clone, Debug)]
 struct DelayService(Timer, Duration);
 
 struct Delay(Sleep, Instant);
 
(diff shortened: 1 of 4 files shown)


carllerche:
Ok, feel free to merge. If you want to pursue the question of whether or not `Load` should be be generic, we can create an issue to track that.

carllerche:
This PR has gone stale. If we want to continue it, we can open a new PR that has had master merged.

PR #49, opened 2018-02-19
InFlightLimit middleware

carllerche:
This PR adds an `InFlightLimit` middleware that enforces a max number if in-flight requests for the inner service as well as all clones.

This middleware works for both `Service` and `ReadyService`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,6 +23,7 @@ members = [
   "tower-buffer",
   "tower-discover",
   "tower-filter",
+  "tower-in-flight-limit",
   "tower-mock",
   "tower-rate-limit",
   "tower-ready-service",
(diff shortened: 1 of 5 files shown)


carllerche:
The implementation should be ready to go. There is one remaining question around trait implementations.

`InFlightLimit` applies both to `Service` and `ReadyService`. Currently, `InFlightLimit<T: ReadyService>` implements `ReadyService` itself. However, it could also implement `Service` directly because the limiting factor is itself the `poll_ready` implementation.

The problem is, coherence doesn't let us implement both `Service` AND `ReadyService` for `InFlightLimit<T: ReadyService>` if there also is an implementation for `T: Service`...

One solution would be to have separate middleware for taking a `T: ReadyService` and lifting it to a `Service` by bounding the max in-flight.

carllerche:
Great, I'm going to merge the PR as is. Handling the issue discussed in the above comment depends on the fate of `ReadyService` (#50).

PR #51, opened 2018-02-23
Implement std::error::Error for Tower error types

hawkw:
I've implemented `std::error::Error` for the error types in the `tower-balance`, `tower-buffer`, `tower-in-flight-limit`, and `tower-reconnect` middleware crates.

This is required upstream for runconduit/conduit#442, and also just generally seems like the right thing to do as a library.

### DIFF
--- a/tower-balance/src/lib.rs
+++ b/tower-balance/src/lib.rs
@@ -318,15 +318,15 @@ impl<F: Future, E> Future for ResponseFuture<F, E> {
 
 impl<T, U> fmt::Display for Error<T, U>
 where
-    T: fmt::Debug,
-    U: fmt::Debug,
+    T: fmt::Display,
+    U: fmt::Display,
 {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         match *self {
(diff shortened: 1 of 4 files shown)


tower-balance/src/lib.rs
```diff
@@ -312,6 +313,48 @@ impl<F: Future, E> Future for ResponseFuture<F, E> {
...
+
+impl<T, U> fmt::Display for Error<T, U>
+where
+    T: fmt::Debug,
+    U: fmt::Debug,
+{
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match *self {
+            Error::Inner(ref why) =>
+                write!(f, "inner service error: {:?}", why),
```
carllerche: Should this not use `Display` as well? The inner errors should impl `Display` as they are also errors (as perf the impl below).

hawkw: Yeah, you're right, that's probably more correct. When I started on this branch, I wasn't sure how far down the stack I was going to have to add `Error` impls, so I wasn't sure if all the inner errors would be `Display`. Will change it.

PR #57, opened 2018-02-28
If tower-buffer sees something, it now says something

hawkw:
When a `tower-buffer` `Buffer` would be closed, it now logs the inner service error that would cause it to be closed. Otherwise, the error is never surfaced, because the buffer returns the `Error::Closed` type (which doesn't wrap an underlying error) and the error that *caused* it to close is discarded.

I made this change to help debug a Conduit issue, and found it to be useful enough that I thought we probably want to make the change here.

### DIFF
--- a/tower-buffer/Cargo.toml
+++ b/tower-buffer/Cargo.toml
@@ -7,7 +7,6 @@ publish = false
 [dependencies]
 futures = "0.1"
 tower = { version = "0.1", path = "../" }
-log = "0.4.1"
 
 [dev-dependencies]
 tower-mock = { version = "0.1", path = "../tower-mock" }
--- a/tower-buffer/Cargo.toml
+++ b/tower-buffer/Cargo.toml
@@ -7,6 +7,7 @@ publish = false
 [dependencies]
 futures = "0.1"
 tower = { version = "0.1", path = "../" }
+log = "0.4.1"
 
 [dev-dependencies]
 tower-mock = { version = "0.1", path = "../tower-mock" }
--- a/tower-buffer/src/lib.rs
+++ b/tower-buffer/src/lib.rs
@@ -11,7 +11,6 @@
 
 extern crate futures;
 extern crate tower;
-#[macro_use] extern crate log;
 
 use futures::{Future, Stream, Poll, Async};
 use futures::future::Executor;
@@ -21,6 +20,8 @@ use tower::Service;
 
 use std::{error, fmt};
 use std::sync::{Arc, RwLock};
+use std::sync::atomic::AtomicBool;
+use std::sync::atomic::Ordering::SeqCst;
 use std::marker::PhantomData;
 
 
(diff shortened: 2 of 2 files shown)


carllerche:
I don't know how I feel about the `error::Error` bound. On one hand, logging the error seems good. On the other, this requires *all* service implementations to ensure that they (manually) impl `Error` for their service's error type.

Perhaps another option would be to somehow bubble up the error. The error could be stored in the `Arc<State>` and `Error::Closed` could somehow reference it?

wdyt

hawkw:
@carllerche that was the other option I was considering --- could go down that route instead, if you prefer? 

I've noticed `tower` tends to have a design philosophy of generally not requiring `Error` implementations, is that indeed the case?

hawkw:
@carllerche and @seanmonstar, I've changed  the PR so that`tower-buffer` will propagate the error rather than logging it in 9c3b39350dca23d4d455887aa91cfdbdf36b937c. 

This change adds a lot of additional complexity --- I'd love your advice on how to make this simpler.

carllerche:
@hawkw I'll do a more thorough review in a bit, but the first thing that stands out is you can get rid of all the poisoned stuff... it's not necessary.

I'd also recommend to keep the `State` struct wrapped with an `Arc` and just have an `error` field on the state struct that is `RwLock<Option<Error>>`.

hawkw:
@carllerche okay, I considered adding an Option<Error> to the state struct,
but I thought the enum seemed cleaner. I can definitely change it back,
though!

Should the buffer just panic if the lock is poisoned, then?

seanmonstar:
Keeping an `Arc<E>` to give to every waiter feels less idiomatic to me. As an example, in finagle's `WatermarkPool`, which is kind of doing similar, if the wrapped factory dies, it rejects all the waiters with simply `ServiceClosedException`.

hawkw:
@seanmonstar hmm, but --- correct me if I'm wrong --- if we don't log the error, and we just give the waiters an empty `Error::Closed`, the error is never surfaced and we have no idea _why_ the buffer was closed, which is the motivating factor behind this PR. Let me know if I'm overlooking some other way to make those errors visible?

seanmonstar:
The `Service` that `Buffer` is wrapping can inspect its own errors and report them, can't it? I assume in finagle, the inner exception is logged or counted just once, and all the waiting requests see that the buffer died. It does seem different to receive 1 `ConnAborted` log/count versus seeing it 57 times in the waiters and thinking that 57 connections suddenly blew up.

olix0r:
> I don't know how I feel about the error::Error bound. On one hand, logging the error seems good. On the other, this requires all service implementations to ensure that they (manually) impl Error for their service's error type.

this does sound pretty heavy-handed to put into buffer...

Could we make a general tower-log-err middleware to do this sort of thing independently of buffer?

i'm not sure it's worth pursuing until we have a broader context about how/where errors should be handled.

carllerche:
What was the conclusion of this? (Keep open or close)?

hawkw:
@carllerche I tool @olix0r's suggestion 
> Could we make a general tower-log-err middleware to do this sort of thing independently of buffer?

and wrote PR #61, which would make this unnecessary if it were merged. Do we want to move forwards with either this PR or #61? 

I think having this kind of functionality could be very useful for debugging, and make the logs more user-friendly...

carllerche:
Ok, I'm going to close this in favor of the other PR. Thanks.

PR #59, opened 2018-02-28
Change Travis test script to `cargo test --all`

hawkw:
This will ensure that all crates are built & tested, and hopefully prevent errors like the one fixed in #58 from slipping past CI.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ matrix:
     - rust: nightly
 
 script:
-- cargo test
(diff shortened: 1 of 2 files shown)


carllerche:
It seems like everything is broken! 😢

hawkw:
@carllerche oh, jeez --- I'll see about fixing those issues as well

hawkw:
@carllerche oh, jeez --- I'll see about fixing those issues as well

PR #61, opened 2018-03-05
Add LogErrors and LogResponses

hawkw:
In https://github.com/tower-rs/tower/pull/57#issuecomment-370177072, @olix0r suggested adding a general purpose log-err middleware. I had some free time this weekend, so I threw one together. 

The `LogErrors` middleware wraps any `Service` or `NewService` and logs any errors returned by the inner service. This can be useful in cases when these errors should be recorded but are not propagated with all of their information intact. For example, the `tower-buffer` middleware  will generate opaque `Closed` errors which hide the error returned by the underlying `Service`, so those errors would need to be logged below the buffer service.

If the logging middleware were to simply log errors using the `log!` macro, the resulting log records will have the module path and target of the `tower-log` crate. This can make it difficult to determine where the errors occurred, if multiple modules in the application using the middleware generate errors with the same message. Therefore, I've also provided a `log_errors!` macro which may be used to wrap services, similarly to `LogErrors::new`. Using the macro will cause the log records to have the location information of the macro expansion site, to make it easier to determine where the errors were seen. 

For example, here's a log message from Conduit using the `log_errors!` macro:
```
ERROR 2018-03-05T18:36:36Z: conduit_proxy::bind: Future::poll: Error attempting to establish underlying session layer: operation timed out after Duration { secs: 0, nanos: 0 }
```
If `LogErrors::new` was used instead, the message would look like
```
ERROR 2018-03-05T18:36:36Z: tower-log: Future::poll: Error attempting to establish underlying session layer: operation timed out after Duration { secs: 0, nanos: 0 }
```
no matter where the `LogErrors` middleware was created.

It was more or less trivial to also implement a `LogResponses` middleware which functions analogously to `LogErrors` but logs successful responses instead, so I've included that as well. The `LogResponses` middleware may be configured to only log `Ready` responses, or to log all successful responses to polls, which may be useful for debugging.

### DIFF
--- a/tower-log/src/lib.rs
+++ b/tower-log/src/lib.rs
@@ -8,10 +8,11 @@ extern crate futures;
 extern crate tower;
 extern crate log;
 
-use futures::{Future, Poll};
+use futures::{Async, Future, Poll};
 use tower::{Service, NewService};
 
 use std::error::Error;
+use std::fmt;
 
 /// Wrap a `Service` or `NewService` with `LogErrors` middleware.
 ///
@@ -69,9 +70,28 @@ macro_rules! log_errors {
             .at_location(file!(), line!())
             .at_level($level)
     };
-
 }
 
+/// Wrap a `Service` or `NewService` with `LogResponses` middleware.
+///
+/// Unlike using `LogResponses::new`, this macro will configure the returned
+/// middleware to log messages with the module path and file/line location of
+/// the _call site_, as using the `log!` macro in that file would.
+#[macro_export]
+macro_rules! log_responses {
+    ($inner:expr) => {
+        log_responses!(level: ::log::Level::Debug, $inner)
+    };
+    (target: $target:expr, $($rest:tt)+) => {
+        log_responses!($($rest),+).target($target)
+    };
+    (level: $level:expr, $inner:expr) => {
+        $crate::LogResponses::new($inner)
+            .in_module(module_path!())
+            .at_location(file!(), line!())
+            .at_level($level)
+    };
+}
 /// Logs error responses.
 #[derive(Clone, Debug)]
 pub struct LogErrors<T> {
@@ -82,6 +102,19 @@ pub struct LogErrors<T> {
     file: Option<&'static str>,
     line: Option<u32>,
 }
+
+/// Logs successful responses.
+#[derive(Clone, Debug)]
+pub struct LogResponses<T> {
+    inner: T,
+    level: log::Level,
+    not_ready: bool,
+    target: Option<&'static str>,
+    module_path: Option<&'static str>,
+    file: Option<&'static str>,
+    line: Option<u32>,
+}
+
 
 // ===== impl LogErrors =====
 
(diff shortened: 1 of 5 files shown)


hawkw:
@carllerche does https://github.com/tower-rs/tower/pull/57#issuecomment-387853011 imply that we intend to move forward with reviewing this PR?

tower-log/src/lib.rs
```diff
@@ -0,0 +1,452 @@
...
+///         LogErrors::new(timeout)
+///     }
+/// }
+/// ```
+/// will produce log messages like
+/// ```notrust,ignore
+/// ERROR 2018-03-05T18:36:36Z: tower_log_errors: Future::poll: operation timed out after Duration { secs: 1, nanos: 0 }
+/// ```
+#[macro_export]
+macro_rules! log_errors {
```
carllerche: I think we should avoid the macro for now. We can consider bringing back a macro if the pattern comes up a lot. I personally am not certain that including file / line numbers is going to be that common.

tower-log/src/lib.rs
```diff
@@ -0,0 +1,452 @@
...
+    };
+}
+
+/// Wrap a `Service` or `NewService` with `LogResponses` middleware.
+///
+/// Unlike using `LogResponses::new`, this macro will configure the returned
+/// middleware to log messages with the module path and file/line location of
+/// the _call site_, as using the `log!` macro in that file would.
+#[macro_export]
+macro_rules! log_responses {
```
carllerche: Same as above.

tower-log/src/lib.rs
```diff
@@ -0,0 +1,452 @@
...
+    (level: $level:expr, $inner:expr) => {
+        $crate::LogResponses::new($inner)
+            .in_module(module_path!())
+            .at_location(file!(), line!())
+            .at_level($level)
+    };
+}
+/// Logs error responses.
+#[derive(Clone, Debug)]
+pub struct LogErrors<T> {
```
carllerche: I'm not certain if `LogErrors` and `LogResponses` should be separate instead of making a single `Log` service. I'm thinking having fewer types would be better here (i.e, a single `Log` service).

tower-log/src/lib.rs
```diff
@@ -0,0 +1,452 @@
...
+
+    /// Set the file and line number of the produced log records.
+    pub fn at_location(mut self, file: &'static str, line: u32) -> Self {
+        self.file = Some(file);
+        self.line = Some(line);
+        self
+    }
+
+    fn child<U>(&self, inner: U) -> LogErrors<U> {
+        LogErrors {
```
carllerche: This struct looks pretty big. I'm wondering if this would work out better storing the config in an `Arc` and cloning. But, this could be figured out later too.

tower-log/src/lib.rs
```diff
@@ -0,0 +1,452 @@
...
+            module_path: None,
+            file: None,
+            line: None,
+        }
+    }
+
+    /// Set the log level of the produced log records.
+    ///
+    /// Log records will be logged at the `Error` level by default.
+    pub fn at_level(mut self, level: log::Level) -> Self {
```
carllerche: Could these be broken out into a `Builder`. Then we can match the fn names from [`RecordBuilder`](https://docs.rs/log/0.4.1/log/struct.RecordBuilder.html):

* `level`
* `target`
* `file`
* `line`

etc...

carllerche:
This PR has gone stale. I think a logging middleware is still useful, but the PR should probably be refreshed.

I'm going to close this for now, we can open a new PR later when we are ready.

PR #67, opened 2018-04-25
Extract `Service` trait and related into crate.

carllerche:
This makes the `tower` crate available to be a "batteries included"
facade.


hawkw:
🙌

PR #68, opened 2018-04-25
Remove `ReadyService`

carllerche:
The value added by having a separate trait is not obvious. Equivalent
behavior can be provided by a `Service` implementation that is always
"ready".


olix0r:
And, so, generally, we expect service implementations to be graceful when poll_ready has not been called (i.e. should not panic, at least). Should this be reflected in the trait docs?

Issue #71, opened 2018-05-10
Buffer's queue does not remove canceled requests


seanmonstar:
Due to `Buffer` using a `futures::sync::mpsc` channel, any `ResponseFutures` that have been dropped will continue to consume space in the queue until the underlying service has progressed through the requests in front of them. A `Buffer` could be wrapped in `Timeout`, which could cancel the requests if waiting took too long. The `oneshot::Sender` still being somewhere in the queue means the buffer's capacity could become full of canceled requests.

There's the additional issue that a wrapped `Reconnect` may wish to only retry a failed connect if there are still response futures waiting, but being in the queue makes it impossible to determine that.

This is kind of the "other" half of a pool. hyper does have a queue of waiters internally, and can check when they are canceled, since they are actually in a `VecDeque`. To allow new requests to enter this queue, it's wrapped in an `Arc<Mutex>`. While perhaps not the best thing in the world, it does work, and people still get excellent performance from hyper's client, so we could consider that as a first pass.

Related Conduit issue: https://github.com/runconduit/conduit/issues/899

carllerche:
I believe that this describes two separate issues.

* Only reconnecting when there is a request to send.
* The inability to purge canceled requests from the queue.

## `Reconnect`

Thinking more about it. I am leaning towards refining `Service::poll_ready` with the following:

* An error is terminal. The service will never work again.
* The fn should only be called when there is a request to send. This means that if the request "goes away", `poll_ready` should not be called anymore.

This means that, if there are no requests in the buffer, `poll_ready` would not be called and the connection would not be established.

The reason that I like this is because it models what would happen if you did a select over the request being canceled and the service's ready future. It also encapsulates the `poll_ready` concept and avoids leaking out of the abstraction.

The problem is that the connection logic might end up being half way there (socket connected and TLS handshake in progress), then `poll_ready` is no longer called and the handshake never finishes. Either this is OK (the remote will timeout) or there there could be a task spawned to drive the connection to completion regardless if `poll_ready` is called. I would also say that this problem exists today w/ the current `Service` API.

## `Buffer`

The second problem is the queue used by `Buffer` does not eagerly release resources for canceled requests. I'm not exactly sure what you are proposing.

[...]

PR #72, opened 2018-05-10
buffer: check for canceled requests before polling inner service

seanmonstar:
Polling the queue for a message and checking if it is canceled before we call `poll_ready` can save the service work, and allow us to clear the queue as requests get canceled. For instance, if a service has reached max concurrent streams and returns `NotReady`, queued requests could start to hit their own timeouts. This change will now allow the queue to clear those out, so the canceled requests don't count to some in-flight limit.

Additionally, if all requests have been canceled, the inner service doesn't need to try to do anything, if that were a `Reconnect` that needed to re-establish a connection.

Related to #71

### DIFF
--- a/tower-buffer/src/lib.rs
+++ b/tower-buffer/src/lib.rs
@@ -201,22 +201,18 @@ where T: Service
             }
         }
 
-        loop {
-            // Get the next request
-            match self.rx.poll()? {
-                Async::Ready(Some(mut msg)) => {
-                    if msg.tx.poll_cancel()?.is_not_ready() {
-                        return Ok(Async::Ready(Some(msg)));
-                    }
-                }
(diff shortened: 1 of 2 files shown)


tower-buffer/src/lib.rs
```diff
@@ -181,46 +187,70 @@ where T: Service
...
+            // Returning NotReady means it is still alive, so we should still
+            // use it.
+            if msg.tx.poll_cancel()?.is_not_ready() {
+                return Ok(Async::Ready(Some(msg)));
+            }
+        }
+
+        loop {
+            // Get the next request
+            match self.rx.poll()? {
```
olix0r: it looks like this can be wrapped in `try_ready!` to aviod the Async matching

PR #74, opened 2018-05-23
balance: Account for long-lived responses

olix0r:
tower-balance provides a PendingRequests load metric that counts the number of responses that have not yet been received. However, especially in the case of HTTP, responses have bodies that remain active far past the initial receipt of the response. We want load metrics to be able to take such streams into account.

This change introduces a new utility trait, `load::Measure`, which is used by implementors of `Load` (like `PendingRequests`) to handle the protocol-specific details of attaching an _instrument_ with a response message.  _Instruments_ are implemented as RAII-guarded types that ensure that load calculations are updated as a response completes.  An instrument is dropped when the load metric no longer needs information from a response.

This all being changed in service of a PeakEWMA balancer implementation, though this should benefit the existing load metric as well.


tower-balance/src/load/track.rs
```diff
@@ -0,0 +1,77 @@
...
+/// This utility allows load metrics to have a protocol-agnostic means to track streams
+/// past their initial response future. For example, if `V` represents an HTTP response
+/// type, an implementaton could add `T`-typed trackers to the HTTP response extensions.
+///
+/// Trackers are intended to be RAII guards that primarily implement `Drop` and update
+/// load metric state as they are dropped.
+///
+/// A base `impl<T, V> Track<T, V> for ()` is provided that drops the tracker
+/// immediately.
+pub trait Track<T, V> {
```
carllerche: Why is this generic over `T`? Is there a reason to not pin this to `Tracker` or whatever is defined above?

tower-balance/src/load/track.rs
```diff
@@ -0,0 +1,77 @@
...
+    type Tracked;
+
+    /// Attaches a `T`-typed trakcer to a `V`-typed value.
+    fn track(tracker: T, value: V) -> Self::Tracked;
+
+    /// Wraps an `F`-typred Future so that a `T`-typed tracker is attached to its result.
+    fn track_future<F>(tracker: T, future: F) -> TrackFuture<F, Self, T>
```
carllerche: Does this need to be a fn on the trait? Since this is a utility trait, the answer would be yes if `track_future` is ever overridden. However, I don't think it can be overridden in a meaningful way.

tower-h2-balance/src/lib.rs
```diff
@@ -0,0 +1,102 @@
...
+    type Data = B::Data;
+
+    fn is_end_stream(&self) -> bool {
+        self.body.is_end_stream()
+    }
+
+    /// Polls a stream of data.
+    fn poll_data(&mut self) -> Poll<Option<Self::Data>, h2::Error> {
+        let data = try_ready!(self.body.poll_data());
+        if data.is_none() {
```
carllerche: I don't think that this is a **guarantee** that there are no trailers.

tower-balance/src/load/track.rs
olix0r: The intent is that each `Load` type may provide a different type.  pending_requests::Tracker simply holds a handle so that, when the request is dropped, the reference count is updated.

peak_ewma::Tracker will have different logic. When a `peak_ewma::Tracker` is blocked, it will grab a mutex and update the cost with the appropriate decay.

Implementations `Track` may or may not care what type of Tracker is provided.  For instance, the tower_h2_balance Track implementations don't care what type of tracker is used. It's conceivable, though, that `Load` and `Track` implementations could rely on a richer tracker type (that has more features than implementing Drop).

tower-balance/src/load/track.rs
olix0r: You're correct. This is provided as a convenience to make construction slightly more ergonomic.  It can probably be changed to `TrackFuture::new`

tower-balance/src/load/track.rs
carllerche: Yeah, this seems good.

carllerche: Ok, this makes sense.

olix0r:
@carllerche 

> Unrelated to this PR and you should take no action based off of this thought: I still wonder if Load::Metric is needed vs. having a 0 -> 1 value. I tend to try to avoid traits / generic unless it has a proven reason to be needed.

Yes, I've also been thinking about this. I suspect this is going to make things much nicer, especially if we can be clear about the significance of values. I think we can say that 0 values indicate no load; a value of 1.0 indicates that the endpoint is _at its maximum estimated capacity_ -- this requires that either the theoretical capacity is known or determined dynamically. When the capacity cannot be known, load types should choose values that _approach_ 1.0 as load increases.  Load values greater than 1.0 indicate that the resource is oversubscribed.

In terms of actual implementation, I might be inclined to use a u64 and treat `1_000_000` as 1.0... wdyt?

carllerche:
> In terms of actual implementation, I might be inclined to use a u64 and treat 1_000_000 as 1.0... wdyt?

Hmm... that seems a bit arbitrary (I also don't have a clear sense of all the implications). That said, the actual public interface should probably be a type w/ operators defined on it, so we can change the internal representation as needed. If this can be done, the initial representation matters less.

olix0r:
> To clarify, Track is only intended to be used with PendingRequests and isn't part of the main load balancer functionality?

peak_ewma will use Track as well.

> I wonder if this track logic should be extracted from the balancer because it seems generally useful. For example, the InFlightLimit middleware could use this functionality as well.

Good point!

carllerche:
Regarding extracting, that also does not have to be done this PR.

My thought is that we can create tracking issues for these other ideas when this gets merged.

tower-balance/src/load/pending_requests.rs
```diff
@@ -1,137 +1,149 @@
...
-    inner: T,
-    pending: Arc<AtomicUsize>,
+#[derive(Debug)]
+pub struct PendingRequests<S, M = NoMeasure>
+where
+    S: Service,
+    M: Measure<Instrument, S::Response>,
+{
+    service: S,
+    track: Arc<()>,
```
hawkw: This is a clever use of `Arc`, but I think a comment explaining what this field is doing might be worth adding, since it's a kind of weird type.

olix0r:
@carllerche @hawkw `Measure` or `Instrument` or ...?

carllerche:
Looks like you already used `Instrument`, so go with that!

tower-balance/src/load/pending_requests.rs
```diff
@@ -1,137 +1,237 @@
...
+    D: Discover,
+{
+    pub fn new(discover: D) -> Self {
+        Self {
+            discover,
+            _p: PhantomData,
+        }
+    }
+
+    pub fn instrument<M>(self) -> WithPendingRequests<D, M>
```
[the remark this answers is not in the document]
olix0r: > Maybe for now, we have new take all the needed info?

This causes users to have to annotate types. For now, I'll make this part of new and we can add builders later if it's a pain.

tower-balance/src/load/pending_requests.rs
```diff
@@ -1,137 +1,237 @@
...
+        Arc::strong_count(&self.0)
     }
 }
 
-// ===== impl ResponseFuture =====
 
-impl<S: Service> Future for ResponseFuture<S> {
-    type Item = S::Response;
-    type Error = S::Error;
+#[cfg(test)]
```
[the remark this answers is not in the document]
olix0r: No, but I'd prefer to address test structure in a followup.

tower-balance/src/load/instrument.rs
```diff
@@ -0,0 +1,84 @@
...
+}
+
+// ===== impl InstrumentFuture =====
+
+impl<F, I, H> InstrumentFuture<F, I, H>
+where
+    F: Future,
+    I: Instrument<H, F::Item>,
+{
+    pub fn new(instrument: I, handle: H, future: F) -> Self {
```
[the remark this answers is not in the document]
olix0r: probably not

PR #77, opened 2018-06-02
balance: Add debug logging for p2c decisions

olix0r:
When debugging load balancer behavior, it's convenient to observe the
individual node selection decisions. To that end, this change requires
that `Load::Metric` implement `fmt::Debug` when used by
`PowerOfTwoChoices`.

### DIFF
--- a/tower-balance/src/choose/p2c.rs
+++ b/tower-balance/src/choose/p2c.rs
@@ -1,7 +1,7 @@
 use rand::Rng;
 
-use Load;
 use choose::{Choose, Replicas};
+use Load;
 
 /// Chooses nodes using the [Power of Two Choices][p2c].
 ///
@@ -51,7 +51,7 @@ impl<R: Rng> PowerOfTwoChoices<R> {
 impl<K, L, R> Choose<K, L> for PowerOfTwoChoices<R>
 where
     L: Load,
-    L::Metric: PartialOrd,
+    L::Metric: PartialOrd + ::std::fmt::Debug,
     R: Rng,
 {
     /// Chooses two distinct nodes at random and compares their load.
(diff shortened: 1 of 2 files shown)


tower-balance/src/choose/p2c.rs
```diff
@@ -51,15 +51,25 @@ impl<R: Rng> PowerOfTwoChoices<R> {
...
 {
     /// Chooses two distinct nodes at random and compares their load.
     ///
     /// Returns the index of the lesser-loaded node.
     fn choose(&mut self, replicas: Replicas<K, L>) -> usize {
         let (a, b) = self.random_pair(replicas.len());
-        if replicas[a].load() <= replicas[b].load() {
+
+        let aload = replicas[a].load();
+        let bload = replicas[b].load();
```
hawkw: NIT: would prefer `a_load` and `b_load`.

tower-balance/src/choose/p2c.rs
```diff
@@ -51,15 +51,25 @@ impl<R: Rng> PowerOfTwoChoices<R> {
...
+        debug!(
+            "choose node[{a}]={aload:?} node[{b}]={bload:?}",
```
hawkw: NIT: Maybe this should be "choos_ing from_ ...`" rather than just "choose"?

tower-balance/src/choose/p2c.rs
```diff
@@ -59,14 +59,24 @@ impl PowerOfTwoChoices {
...
     /// Chooses two distinct nodes at random and compares their load.
     ///
     /// Returns the index of the lesser-loaded node.
     fn choose(&mut self, replicas: Replicas<K, L>) -> usize {
         let (a, b) = self.random_pair(replicas.len());
-        if replicas[a].load() <= replicas[b].load() {
+
+        let a_load = replicas[a].load();
+        let b_load = replicas[b].load();
+        debug!(
```
seanmonstar: Is this info important enough to be in debug logs? My initial feeling is that it's noisy enough to be `trace!`, but you probably have better context on how often you need to see these.

PR #78, opened 2018-06-02
balance: Do not require RNG to build a P2C balancer

olix0r:
`PowerOfTwoChoices` requires a Random Number Generator. In order for
this randomization source to be configurable (i.e. for tests),
`PowerOfTwoChoices` was generic over its implementation of `rand::Rng`;
however, this leads to needless boilerplate when building P2C balancers.

In order to resolve this, the `p2c` module now uses `rand 0.5`'s `SmallRng`
type, which is smaller faster, though not cryptographically secure.
`PowerOfTwoChoices` may either be constructed with an instance of
`SmallRng`, or it may be constructed as a `Default`, using system entropy.

### DIFF
--- a/tower-balance/src/choose/p2c.rs
+++ b/tower-balance/src/choose/p2c.rs
@@ -1,7 +1,7 @@
-use rand::Rng;
+use rand::{self, Rng};
 
-use Load;
 use choose::{Choose, Replicas};
+use Load;
 
 /// Chooses nodes using the [Power of Two Choices][p2c].
 ///
@@ -18,13 +18,19 @@ use choose::{Choose, Replicas};
 ///
 /// [finagle]: https://twitter.github.io/finagle/guide/Clients.html#power-of-two-choices-p2c-least-loaded
 /// [p2c]: http://www.eecs.harvard.edu/~michaelm/postscripts/handbook2001.pdf
-#[derive(Debug)]
-pub struct PowerOfTwoChoices<R> {
+#[derive(Debug, Default)]
+pub struct PowerOfTwoChoices<R: Rng = ThreadRng> {
     rng: R,
 }
 
+/// A helper implementation of `Rng` that lazily uses the thread-local RNG.
+#[derive(Debug, Default)]
+pub struct ThreadRng(());
+
+// ==== impl PowerOfTwoChoices ====
+
 impl<R: Rng> PowerOfTwoChoices<R> {
-    pub fn new(rng: R) -> Self {
+    pub fn new_with_rng(rng: R) -> Self {
         Self { rng }
     }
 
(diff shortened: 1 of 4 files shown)


tower-balance/src/choose/p2c.rs
```diff
@@ -18,13 +18,19 @@ use choose::{Choose, Replicas};
 ///
 /// [finagle]: https://twitter.github.io/finagle/guide/Clients.html#power-of-two-choices-p2c-least-loaded
 /// [p2c]: http://www.eecs.harvard.edu/~michaelm/postscripts/handbook2001.pdf
-#[derive(Debug)]
-pub struct PowerOfTwoChoices<R> {
+#[derive(Debug, Default)]
+pub struct PowerOfTwoChoices<R: Rng = ThreadRng> {
```
briansmith: I am not sure why we need to parameterize the RNG. I think just using one default one would be OK.

If you're going to use the `rand` crate and constructing `PowerOfTwoChoices` objects is infrequently then I would suggest that `PowerOfTwoChoices` just holds a `SmallRng` that it seeds when it is first constructed and never reseeds.

I suggest documenting that the RNG doesn't need to be cryptographically secure but does need to be fast. (This is why I don't like `ThreadRng`; it tries to mostly be cryptographically secure and fast at the same time.)

tower-balance/src/choose/p2c.rs
```diff
@@ -80,7 +93,7 @@ mod tests {
                 return TestResult::discard();
             }
 
-            let (a, b) = PowerOfTwoChoices::new(rand::thread_rng()).random_pair(n);
+            let (a, b) = PowerOfTwoChoices::<ThreadRng>::default().random_pair(n);
```
briansmith: IS `<ThreadRng>` needed here?

tower-balance/src/choose/p2c.rs
briansmith: > I am not sure why we need to parameterize the RNG. I think just using one default one would be OK.

OIC, so it can be swapped out for tests, as in the initial description of the PR. AFAICT you don't need to actually parameterize on the choice of PRNG algorithm, but instead you just need to provide a way to override the random seed of the PRNG. For example, if you just used `SmallRng` then you'd just need a constructor that takes a `&[u8; 16]` parameter that you can pass to `SmallRng::from_seed()`.

tower-balance/src/choose/p2c.rs
olix0r: i was surprised that, for some reason, the compiler needed it. i think it may somehow be related to quickcheck...

briansmith:
LGTM.

carllerche:
I think this is good. It looks like rustfmt was run on the file, which makes it a bit tricky to figure out what exactly the changes are.

Is it mostly p2c.rs up to lines 30ish?

tower-balance/src/lib.rs
```diff
@@ -138,9 +136,9 @@ where
     /// Polls `discover` for updates, adding new items to `not_ready`.
     ///
     /// Removals may alter the order of either `ready` or `not_ready`.
-    fn update_from_discover(&mut self)
-        -> Result<(), Error<<D::Service as Service>::Error, D::DiscoverError>>
-    {
+    fn update_from_discover(
+        &mut self,
+    ) -> Result<(), Error<<D::Service as Service>::Error, D::DiscoverError>> {
```
olix0r: ugh, got rustfmtd. will undo

PR #80, opened 2018-06-02
balance: Make Balance::is_ready public

olix0r:
`Balance::is_ready` is erroneously private. This causes `dead_code`
warnings.

Make `is_ready` public and prohibit `dead_code`.

### DIFF
--- a/tower-balance/src/lib.rs
+++ b/tower-balance/src/lib.rs
@@ -1,3 +1,5 @@
+#![deny(dead_code)]
+
 #[macro_use]
 extern crate futures;
 #[macro_use]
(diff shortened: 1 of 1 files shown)


tower-balance/src/lib.rs
```diff
@@ -1,3 +1,5 @@
+#![deny(dead_code)]
```
seanmonstar: I tend to just put `#![deny(warnings]` in all libraries, it catches plenty of mistakes. Well, a slightly bigger list:

```rust
#![deny(warnings)]
#![deny(missing_docs)]
#![deny(missing_debug_implementations)]
```

PR #81, opened 2018-06-03
balance: Move constructors under `Balance`

olix0r:
Previously, `power_of_two_choices` and `round_robin` constructors were
exposed from the crate scope.

These have been replaced by `Balance::p2c`, `Balance::p2c_from_rng`, and
`Balance::round_robin`.

Depends on #78.

### DIFF
--- a/tower-balance/src/lib.rs
+++ b/tower-balance/src/lib.rs
@@ -11,6 +11,7 @@ extern crate tower_service;
 
 use futures::{Async, Future, Poll};
 use indexmap::IndexMap;
+use rand::{rngs::SmallRng, SeedableRng};
 use std::marker::PhantomData;
 use std::{error, fmt};
 use tower_discover::Discover;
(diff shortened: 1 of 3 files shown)


tower-balance/src/lib.rs
```diff
@@ -88,6 +59,52 @@ pub struct ResponseFuture<F: Future, E>(F, PhantomData<E>);
...
+    /// > The maximum load variance between any two servers is bound by `ln(ln(n))` where
+    /// > `n` is the number of servers in the cluster.
+    ///
+    /// [finagle]: https://twitter.github.io/finagle/guide/Clients.html#power-of-two-choices-p2c-least-loaded
+    /// [p2c]: http://www.eecs.harvard.edu/~michaelm/postscripts/handbook2001.pdf
+    pub fn p2c(discover: D) -> Self {
+        Self::new(discover, choose::PowerOfTwoChoices::default())
+    }
+
+    /// Initializes a P2C load balancer from the provided randomization source.
```
hawkw: NIT: Maybe the docs should note that the provided RNG is used to initialize a `SmallRng`?

tower-balance/src/lib.rs
```diff
@@ -88,6 +59,52 @@ pub struct ResponseFuture<F: Future, E>(F, PhantomData<E>);
...
+    ///
+    /// This may be preferable when an application instantiates many balancers.
+    pub fn p2c_from_rng<R: rand::Rng>(
```
carllerche: Nit: I've been using `with` in this case for naming: `p2c_with_rng`. I think it would be worth to stay consistent. I understand that it is *slightly* different in that `rng` is not being moved to a field, but I think it fits anyway.

olix0r: This naming was inspired by `rand::SmallRng::from_rng` -- but happy to change to be more consistent within tower.

PR #85, opened 2018-06-08
in-flight-limit: Add missing task notification.

### LINKED ISSUE #70: tower-in-flight missing notification
The `Service` task is never notified when the service becomes ready (number of in-flight requests drop below the max).

carllerche:
Previously, there was no notification when capacity is made available by
requests completing. This patch fixes the bug.

This also switches the tests to use [`MockTask`](https://github.com/carllerche/tokio-test/blob/master/src/lib.rs) from tokio-test.

Fixes #70

### DIFF
--- a/tower-in-flight-limit/src/lib.rs
+++ b/tower-in-flight-limit/src/lib.rs
@@ -220,7 +220,7 @@ impl Shared {
     pub fn release(&self) {
         let prev = self.curr.fetch_sub(1, SeqCst);
 
-        // Cannot go abovee the max number of in-flight
+        // Cannot go above the max number of in-flight
         debug_assert!(prev <= self.max);
 
         if prev == self.max {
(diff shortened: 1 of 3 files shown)


tower-in-flight-limit/src/lib.rs
```diff
@@ -218,7 +218,14 @@ impl Shared {
     /// request has completed OR the service that made the reservation has
     /// dropped.
     pub fn release(&self) {
-        self.curr.fetch_sub(1, SeqCst);
+        let prev = self.curr.fetch_sub(1, SeqCst);
+
+        // Cannot go abovee the max number of in-flight
```
olix0r: typo

carllerche: G.D. macbook pro keyboard... been having trouble w/ double `e`s all week.

PR #86, opened 2018-06-08
balance: Use a default RTT estimate of 1s

olix0r:
After speaking with @roanta and @adleong, I realized that our
default RTT estimate is too optimstic: it gives new endpoints
an _ideal_ RTT. Instead, our intent is to assign a slightly
pessimistic cost to new endpoints so they don't take on more load than
they are due before the EWMA can establish a baseline.

To that end, this changes the default RTT estimate to be 1 second.

---
In a followup review, we may want to adopt builders that make this configurable.

### DIFF
--- a/tower-balance/src/load/peak_ewma.rs
+++ b/tower-balance/src/load/peak_ewma.rs
@@ -24,8 +24,8 @@ use Load;
 /// the underlying service to the response future being satisfied (or dropped).
 ///
 /// When no latency information has been measured for an endpoint, an arbitrary default
-/// RTT of 1 second is used to prevent new endpoints from being inundated before a
-/// meaningful baseline can be esatablished..
+/// RTT of 1 second is used to prevent the endpoint from being overloaded before a
+/// meaningful baseline can be established..
 ///
 /// ## Note
 ///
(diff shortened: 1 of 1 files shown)


briansmith:
> To that end, this changes the default RTT estimate to be 1 second.

What if the average RTT of existing nodes is larger than 1 second already? It seems like ideally it would be a function of the current average, if that's possible.

olix0r:
@briansmith 

> It seems like ideally it would be a function of the current average, if that's possible.

It's a design choice that the each node's decisions can be made without having to maintain state across all replicas in the balancer. We never look at all of the loads in the replica set at once: with p2c we only ever look at 2 random nodes to make a decision. This makes node selection cheap, at the expense of making less correct decisions (though, p2c says the the randomized choice is _effectively_ equivalent to maintaining a full cluster review, statistically).  All that said, I'm not against other approaches if we find they're necessary.

As a point of reference, Finagle uses a totally arbitrary penalty value of 140737488355327ns (~39 hours), because they are especially conservative with regard to stop costs in JVM-rich environments. Per discussion with @adleong, I don't think we should be quite so pessimistic, though we should almost certainly should be less optimistic than than the ideal 500,000ns that I used initially.

briansmith:
OK, sounds reasonable. I just wanted to make sure that situation wasn't being overlooked.

olix0r:
@seanmonstar 

> Personally, I'd rather use u64s. We likely don't actually need a floating point precision. But this PR seems fine.

Since this ends up in a `Cost(f64)` (which in non-deafult cases is computed as a function of floating point math), I thought it simpler to avoid the runtime conversion.

PR #87, opened 2018-06-22
Introduce tower-watch

olix0r:
`tower_watch::WatchService` provides a dynamically-bound `Service` that
updates in response to a `Watch`. A `WatchService` is constructed with a
`Watch<T>` and a `Bind<T>` -- `Bind` is a newly introduced trait that
supports instantiating new service instances with a borrowed value, i.e.
from a watch.

This can be used to reconfigure Services from a shared or otherwise
externally-controlled configuration source (for instance, a file
system).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,6 +13,7 @@ members = [
   "tower-service",
   "tower-timeout",
   "tower-util",
+  "tower-watch",
 ]
 
 [dev-dependencies]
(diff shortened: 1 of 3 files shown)


olix0r:
Strike that, reverse it. `StreamNewService` is more difficult to implement than originally hoped. I propose that we use this approach until we find a better one.

seanmonstar:
This is essentially the discovery stuff we use in Conduit, generalized, right?

olix0r:
@seanmonstar Sort of. We use this within the bound stacks produced by discovery so that the discover implementation can update an endpoint's stack without removing it from the load balancer.

olix0r:
This PR could use a review ;)

PR #90, opened 2018-07-13
tower-watch: Add Error impls for tower_watch::Error

hawkw:
This branch adds `std::fmt::Display` and `std::error::Error` impls
to `tower-watch`'s `Error` type.

This is necessary to adopt `tower-watch` in Linkerd 2's proxy
(rather than our internal implementation of a similar type), as 
the errors will be part of an error chain which requires these
traits.

### DIFF
--- a/tower-watch/src/lib.rs
+++ b/tower-watch/src/lib.rs
@@ -4,7 +4,6 @@ extern crate futures_watch;
 extern crate tower_service;
 
 use std::{fmt, error};
-
 use futures::{Async, Future, Poll, Stream};
 use futures_watch::{Watch, WatchError};
 use tower_service::Service;
@@ -95,7 +94,7 @@ where
     fn cause(&self) -> Option<&error::Error> {
         match self {
             Error::WatchError(_) => None,
-            Error::Inner(e) => Some(e),
+            Error::Inner(e) => e.cause(),
         }
     }
 }
--- a/tower-watch/src/lib.rs
+++ b/tower-watch/src/lib.rs
@@ -3,6 +3,8 @@ extern crate futures;
 extern crate futures_watch;
 extern crate tower_service;
 
+use std::{fmt, error};
+
 use futures::{Async, Future, Poll, Stream};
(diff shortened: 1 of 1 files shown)


tower-watch/src/lib.rs
```diff
@@ -72,6 +74,32 @@ impl<T, B: Bind<T>> Service for WatchService<T, B> {
...
+}
+
+impl<E> error::Error for Error<E>
+where
+    E: error::Error,
+{
+    fn cause(&self) -> Option<&error::Error> {
+        match self {
+            Error::WatchError(_) => None,
+            Error::Inner(e) => Some(e),
```
carllerche: Hmm is the inner error the cause or should this fn proxy to the inner error’s cause?

Thoughts @seanmonstar?

seanmonstar: It should proxy. I've been bitten by this before. Consider things that print a cause chain:

```rust
println!("error: {}", err);

loop {
    err = err.cause();
    if let Some(err) = err {
        println!("\tCaused by: {}", err);
        err = err.cause();
    } else {
        break;
    }
}
```

Since the inner error would be printed as the root, it's confusing to then see it also as the cause.

tower-watch/src/lib.rs
```diff
@@ -3,6 +3,8 @@ extern crate futures;
 extern crate futures_watch;
 extern crate tower_service;
 
+use std::{fmt, error};
+
```
olix0r: tioli: i personally don't think there's much value in splitting std imports out into another block.

tower-watch/src/lib.rs
hawkw: Changed it to proxy in 142f0fcfe1c85002999925ba8e99f46637314bd0.

tower-watch/src/lib.rs
hawkw: Removed the whitespace in 142f0fcfe1c85002999925ba8e99f46637314bd0.

Issue #14, opened 2017-10-08
Retry middleware


danburkert:
It would be nice to have a service that can handle retrying requests.  There are a lot of variables that should be considered for a fully general retry mechanism:

* Distinguishing retriable errors from non-retriable errors
* Configurable backoff between retries
* Failing over between multiple downstream services

Some prior work in the area: https://crates.io/crates/tokio-retry

carllerche:
This is on master.

PR #96, opened 2018-08-14
Tower Retry

seanmonstar:
Initial sketch of a `Retry` middleware, ref #14.

### DIFF
--- a/tower-retry/src/lib.rs
+++ b/tower-retry/src/lib.rs
@@ -1,0 +1,97 @@
+extern crate futures;
+extern crate tower_service;
+
+use futures::{Future, Poll};
+use tower_service::Service;
+
+#[derive(Clone, Debug)]
+pub struct Retry<P, S> {
+    policy: P,
+    service: S,
+}
+
+#[derive(Debug)]
+pub struct ResponseFuture<P, S: Service> {
+    attempt: usize,
+    future: S::Future,
+    request: Option<S::Request>,
+    retry: Retry<P, S>,
+}
+
+pub trait Policy<R, E> {
+    fn retry(&self, err: E, attempt: usize) -> Result<(), E>;
+    fn clone_request(&self, req: &R) -> Option<R>;
+}
+
+
+// ===== impl Retry =====
+
+impl<P, S> Retry<P, S>
+where
+    P: Policy<S::Request, S::Error> + Clone,
+    S: Service + Clone,
+{
+    pub fn new(policy: P, service: S) -> Self {
+        Retry {
+            policy,
+            service,
+        }
+    }
+}
+
+impl<P, S> Service for Retry<P, S>
+where
+    P: Policy<S::Request, S::Error> + Clone,
+    S: Service + Clone,
+{
+    type Request = S::Request;
+    type Response = S::Response;
+    type Error = S::Error;
+    type Future = ResponseFuture<P, S>;
+
+    fn poll_ready(&mut self) -> Poll<(), Self::Error> {
+        self.service.poll_ready()
+    }
+
+    fn call(&mut self, request: Self::Request) -> Self::Future {
+        let cloned = self.policy.clone_request(&request);
+        let future = self.service.call(request);
+        ResponseFuture {
+            attempt: 1,
+            future,
+            request: cloned,
+            retry: self.clone(),
+        }
+    }
+}
+
+// ===== impl ResponseFuture =====
+
+impl<P, S> Future for ResponseFuture<P, S>
+where
+    P: Policy<S::Request, S::Error> + Clone,
+    S: Service + Clone,
+{
+    type Item = S::Response;
+    type Error = S::Error;
+
+    fn poll(&mut self) -> Poll<Self::Item, Self::Error> {
+        loop {
+            match self.future.poll() {
+                Ok(v) => return Ok(v),
+                Err(err) => {
+                    if let Some(req) = self.request.take() {
+                        self.retry.policy.retry(err, self.attempt)?;
+                        self.attempt += 1;
+                        self.request = self.retry.policy.clone_request(&req);
+                        self.future = self.retry.service.call(req);
+                    } else {
+                        // request wasn't cloned, so no way to retry it
+                        return Err(err);
+                    }
+                }
+            }
+        }
+    }
+}
+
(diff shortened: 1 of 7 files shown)


seanmonstar:
Some thoughts I had while investigating:

- Having the policy return a duration or `Future` seemed to conflate the concerns, if one wants backoff support, that can be done by wrapping `Retry` around a `Backoff`.
- The policy only inspects `Error`s along with an attempt count. If certain kinds of responses are supposed to be retried, an inner service can map them to an `Error` kind.
- Not all requests can be cloned, or retried (such as HTTP requests with a non-buffered body), so it can't be `S::Request: Clone`. Therefore, the `Policy` determines whether and how to clone the request before it is passed to the inner service.

tower-retry/src/lib.rs
```diff
@@ -0,0 +1,97 @@
...
+    fn poll(&mut self) -> Poll<Self::Item, Self::Error> {
+        loop {
+            match self.future.poll() {
+                Ok(v) => return Ok(v),
+                Err(err) => {
+                    if let Some(req) = self.request.take() {
+                        self.retry.policy.retry(err, self.attempt)?;
+                        self.attempt += 1;
+                        self.request = self.retry.policy.clone_request(&req);
+                        self.future = self.retry.service.call(req);
```
seanmonstar: I *think* this may want to call `service.poll_ready()` again, because in the meantime, some other request may have taken this one's spot, or we may otherwise just need to check for room again. Especially since the wrapped service could be a `Balance`, and thus the retry could go to a different endpoint.

seanmonstar:
Changes since trying to use this:

- Inspection now receives the `Request` and `Result<Response, Error>`, since at times you need to look at the successful `Response` and the `Request` at the same time to determine if it was actually a failure and can be retried.
- If retry is desired, now returns an `Option<impl Future>` to allow adding some sort of delay. The value from that future is a new `Policy` to allow adjusting state for subsequent retries.

tower-retry/src/lib.rs
```diff
@@ -0,0 +1,138 @@
...
+    /// Polling the future from `Policy::retry`
+    Checking(P, Option<Result<R, E>>),
+    /// Polling `Service::poll_ready` after `Checking` was OK.
+    Retrying,
+}
+
+pub trait Policy<Req, Res, E>: Sized {
+    type Future: Future<Item=Self, Error=()>;
+    fn retry(&self, req: &Req, res: Result<&Res, &E>) -> Option<Self::Future>;
+    fn clone_request(&self, req: &Req) -> Option<Req>;
```
carllerche: I would recommend a different name than `clone_request`. The `clone` name implies that the return value is identical to the source. Instead, the fn might want to modify the request somehow. For example, it might add an HTTP header w/ the attempt count.

tower-retry/src/lib.rs
```diff
@@ -0,0 +1,138 @@
...
+enum State<F, P, R, E> {
+    /// Polling the future from `Service::call`
+    Called(F),
+    /// Polling the future from `Policy::retry`
+    Checking(P, Option<Result<R, E>>),
+    /// Polling `Service::poll_ready` after `Checking` was OK.
+    Retrying,
+}
+
+pub trait Policy<Req, Res, E>: Sized {
```
carllerche: Why are these not associated types? I have some idea, but it would be helpful to explain it out.

If ATs don't work, `Policy` could be generic over `S: Service` to keep it a bit more succinct?

Also, why the `Sized` bound?

carllerche:
I am currently reviewing, but this would also be of interest to @danburkert.

tower-retry/src/lib.rs
```diff
@@ -0,0 +1,138 @@
...
+    type Future: Future<Item=Self, Error=()>;
+    fn retry(&self, req: &Req, res: Result<&Res, &E>) -> Option<Self::Future>;
```
carllerche: My initial reaction to this was that I found it odd. Specifically, `retry` returning a future of `Self`.

Could you elaborate a bit more on the thought process that went into this design? I assume it has something to do w/ backoff.

Also, given that it takes `&self` and returns a future of `Self`, that implies cloning `Self` at some point. This could be tricky if `Self` contains data that cannot easily be cloned. Do you think this is a problem in practice?

seanmonstar:
> How would you implement a policy that had a max # of retries?

https://github.com/tower-rs/tower/pull/96/files#diff-80fb02560ebb57c0ba7c3ca0187d812fR126

tower-retry/src/lib.rs
seanmonstar: I started with generic parameters since they're "input" types, not "output" types. That's as far as my thinking on the matter has gone.

seanmonstar: The `Sized` bound was needed to allow `type Future = Future<Item = Self>`.

tower-retry/src/lib.rs
seanmonstar: Yea, it was around wanting to provide backoff functionality. I realized that we could classify "successes" as needing a retry (like an HTTP 503), but an underlying `Backoff` wouldn't have known it was an error and that it should backoff.

> Also, given that it takes &self and returns a future of Self, that implies cloning Self at some point. 

It *could* mean cloning, or it could mean constructing a new version of `Self`. If there is expensive values to clone, they can be put in an `Arc`.

What I found useful here though is that it allows a retry policy to adapt depending on multiple factors: request, response or error, and it's current state. The `Limit` test case shows how it can therefore adapt, and build more "scoped" policies as retries go on.

Both of these combined are from noticing that finagle's `RetryPolicy` returns `Option<(Duration, RetryPolicy)>`. The duration allows delaying the retry, and the new policy can contain knowledge of the previous attempts. By returning a `Future` directly, it seemed more flexible. It *can* just be a delay of some kind, or it can be something else. And it means `tower_retry` doesn't need a dependency on `tokio_timer`.

tower-retry/src/lib.rs
seanmonstar: Any name recommendations?

carllerche:
I think this is a great start and we can merge it in and iterate on it as we discover cases that are not well handled.

My only thought is to change `trait Policy` to be generic over `S: Service` instead of explicitly taking `Req, Res, E`. This would clean up  the trait...

There are pros / cons either way, so what do others think?

PR #97, opened 2018-08-15
Make error types forward-compatible

hawkw:
This branch obsoletes and closes #90.

Currently, a number of Tower crates expose public error types which are
`enum`. This can lead to forward-compatibility issues, as enum matching
is exhaustive, and if additional variants are added, this becomes a
breaking change for any consumer of the library who is matching on the
enum's variants. 

This branch reworks the error types to consist of public `struct`s with
a single field, which is a private `ErrorKind` enum. The structs provide
methods to check whether the error is a given kind (so that, for
example, if I have a `pub struct Error` and 
`enum ErrorKind { Connect,... }`, `Error` has a method `is_connect`
that returns `true` if its' kind is `ErrorKind::Connect`). Furthermore,
if one of the enum variants wraps another error type, it also has
methods which consume it and return the inner type optionally if the
error is that kind, and which borrow the inner type optionally. 

Since this code ended up being very repetitive, I made this change using
a macro to generate the various error types.

This is a breaking change for any downstream consumers who are matching
against Tower's old enum error types, so we may not want to merge it
right away.

### DIFF
--- a/tower-balance/src/lib.rs
+++ b/tower-balance/src/lib.rs
@@ -26,6 +26,11 @@ pub mod load;
 pub use choose::Choose;
 pub use load::Load;
 
+#[macro_use]
+mod macros {
+    include! { concat!(env!("CARGO_MANIFEST_DIR"), "/../gen_errors.rs") }
+}
+
 /// Balances requests across a set of inner services.
 #[derive(Debug)]
 pub struct Balance<D: Discover, C> {
@@ -50,20 +55,17 @@ pub struct Balance<D: Discover, C> {
     not_ready: IndexMap<D::Key, D::Service>,
 }
 
-/// Error produced by `Balance`
-#[derive(Debug)]
-pub struct Error<T, U> {
-    kind: ErrorKind<T, U>,
-}
-
-#[derive(Debug)]
-enum ErrorKind<T, U> {
-    Inner(T),
-    Balance(U),
-    NotReady,
-}
-
 pub struct ResponseFuture<F: Future, E>(F, PhantomData<E>);
+
+kind_error! {
+    /// Errors produced by `Balance`.
+    #[derive(Debug)]
+    pub struct Error from enum ErrorKind {
+        Inner(T) => is: is_inner, into: into_inner, borrow: borrow_inner,
+        Balance(U) => is: is_balance, into: into_balance, borrow: borrow_balance,
+        NotReady => fmt: "not ready", is: is_not_ready, into: UNUSED, borrow: UNUSED
+    }
+}
 
 // ===== impl Balance =====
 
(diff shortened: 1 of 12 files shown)


tower-balance/src/lib.rs
```diff
@@ -26,6 +26,11 @@ pub mod load;
 pub use choose::Choose;
 pub use load::Load;
 
+#[macro_use]
+mod macros {
+    include! { concat!(env!("CARGO_MANIFEST_DIR"), "/../gen_errors.rs") }
```
hawkw: In case anyone is curious about why the macro for generating error types is defined in a bare Rust file in the crate root and then `include!`ed into every crate that uses it, here's my rationale:

I wanted to be able to reuse the macro across multiple crates without having to copy it into all of them. However, I did _not_ want it to be part of a public API surface that we'd have to maintain. 

If the macro was in its own crate that the other crates depended on, and we didn't want to publish the macro to crates.io (making it public), it would have to be depended on as a path dependency. Crates with path dependencies cannot be published to crates.io, which would be a problem when we eventually release Tower to crates.io. Additionally, people would still be able to pull in the macro as a Git dependency. 

While it would be theoretically possible to add the bare .rs file as a dependency (such as by including the Tower repo as a submodule and `include!`ing the macro file), there's enough friction that I feel like it's not really as big a worry.  Thus, while this approach is admittedly somewhat ugly, it's the only way I could think of to reuse the macro without inadvertently making it public API.

As to why the macro is `include!`ed into its own module: there are two helper macros (`ignore_t!` and `arm!`) that are used inside the main macro for generating error types. I didn't want to expose those inside the crates that consume the macro, as they don't really make any sense outside of the context of the main macro. By including the macro in a module with `#[macro_use]` and only marking the main macro as `#[macro_use]`, crates using the macro can't access the helper macros.

hawkw:
As of c723eefc7215375e15d61a493b5fc8f58cdb92ab, all the generated methods on the public error types also have RustDoc comments:
screen shot 2018-08-16 at 11 12 23 am

carllerche:
What are your thoughts @seanmonstar?

seanmonstar:
Sorry, missed the notification. I think the macro makes it harder to know what exactly is exported and thus part of the "public" API.

There's also #131 which touches on this with a different proposed solution.

carllerche:
@hawkw thanks so much for doing the work. I believe that we are all leaning towards #131 as the strategy for errors. Maybe you can try to take some of that on when you have a second? I would recommend doing one crate per PR and one PR at a time.

I'm going to close this, but feel free to reopen if I am incorrect.

hawkw:
That sounds good to me 👍

Issue #99, opened 2018-08-29
Should `Service` be generic over `Request`


carllerche:
# Problem

Currently, `Service` represents the request as an associated type. This was originally done to mirror `Sink` in the `futures` crate (relevant [issue](https://github.com/rust-lang-nursery/futures-rs/issues/623)). However, requests are an "input" to `Service` which can make representing some things a bit tricky.

For example, HTTP services accept requests in the form of `http::Request<T: BufStream>`. In order to implement a service that handles these requests, the service type must have a phantom generic:

```rust
struct MyService<RequestBody> {
    // ...
    _phantom: PhantomData<RequestBody>,
}

impl<RequestBody> Service for MyService<RequestBody> {
    type Request = http::Request<RequestBody>;
}
```

This, in turn, can cause type inference issues as the request body must be picked when `MyService` is initialized.

# Proposal

The `Service` trait would become:

```rust
pub trait Service<Request> {
    type Response;
    type Error;
    type Future: Future<Item = Self::Response, Error = Self::Error>;

    fn poll_ready(&mut self) -> Poll<(), Self::Error>;

    fn call(&mut self, req: Request) -> Self::Future;
}
```

This moves the request type to a generic.

# Impact

The full impact of this change is unknown. My best guess is it will be mostly ergonomic. For example:

**Before:**

```rust
fn foo<S: Service>(service: S) { }
```

**After:**

```rust
fn foo<R, S: Service<R>>(service: S) { }
```

And it linearly gets worse on the number of generic services:

**Before:**

```rust
fn foo<S1, S2>(s1: S1, s2: S2)
where
    S1: Service,
    S2: Service<Response = S1::Response>,
{ }
```

**After:**

```rust
fn foo<S1, S2, R1, R2>(s1: S1, s2: S2)
where
    S1: Service<R1>,
    S2: Service<R2, Response = S1::Response>,
{ }
```

*today* there already are plenty of cases in which the one has to add additional generic parameters. In order to bound an HTTP service *today*, the following is required:

```rust
fn foo<S, B>(service: S)
where
    S: Service<Request = http::Request<B>>,
    B: BufStream,
{ }
```

In practice, this turns out to be quite painful and it gets unmanageable quickly. To combat this, ["alias" traits](https://github.com/carllerche/tower-web/blob/master/src/util/http/service.rs#L12-L38) are created. It is unclear how this change would impact the ability to use this trick.

carllerche:
[...]

PR #100, opened 2018-08-30
Update tower-rate-limit to work with new tokio_timer

hawkw:
While working on a prototype implementation of a generic `Service` trait
(#99), I noticed that `tower-rate-limit` no longer compiles. It depends
on `tokio-timer` v0.1, which was pulling in the _new_ version of
`tokio-timer`, but it was using the old API from before the tokio
reform.

This branch updates that crate to work with the current `tokio` and
`tokio-timer`.

### DIFF
--- a/tower-rate-limit/Cargo.toml
+++ b/tower-rate-limit/Cargo.toml
@@ -7,7 +7,7 @@ publish = false
 [dependencies]
 futures = "0.1"
 tower-service = { version = "0.1", path = "../tower-service" }
-tokio-timer = "0.2"
+tokio-timer = "0.2.6"
 
 [dev-dependencies]
 tower-mock = { version = "0.1", path = "../tower-mock" }
--- a/tower-rate-limit/Cargo.toml
+++ b/tower-rate-limit/Cargo.toml
@@ -7,7 +7,8 @@ publish = false
 [dependencies]
 futures = "0.1"
 tower-service = { version = "0.1", path = "../tower-service" }
-tokio-timer = "0.1"
+tokio-timer = "0.2"
 
 [dev-dependencies]
 tower-mock = { version = "0.1", path = "../tower-mock" }
+tokio = "0.1"
--- a/tower-rate-limit/src/lib.rs
+++ b/tower-rate-limit/src/lib.rs
@@ -10,7 +10,7 @@ extern crate tokio_timer;
 
 use futures::{Future, Poll};
 use tower_service::Service;
-use tokio_timer::{timer, Delay};
+use tokio_timer::Delay;
 
 use std::{error, fmt};
 use std::time::{Duration, Instant};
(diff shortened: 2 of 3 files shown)


hawkw:
N.B. that the tests are _somewhat_ uglier now --- I think this is fine, but if anyone has suggestions on how to make them simpler, they'd be appreciated!

tower-rate-limit/Cargo.toml
```diff
@@ -7,7 +7,8 @@ publish = false
 [dependencies]
 futures = "0.1"
 tower-service = { version = "0.1", path = "../tower-service" }
-tokio-timer = "0.1"
+tokio-timer = "0.2"
```
carllerche: I've been switching to including the latest patch. We should probably be respecting "min versions" in our deps.

I wouldn't eagerly update any other dependency versions, just on a forward basis.

So this would be `tokio-timer = "0.2.6"`

tower-rate-limit/src/lib.rs
```diff
@@ -143,7 +141,7 @@ where S: Service
                     self.state = State::Ready { until, rem };
                 } else {
                     // The service is disabled until further notice
-                    let sleep = self.timer.sleep(until - now);
+                    let sleep = timer::Handle::current().delay(until);
```
carllerche: This should just be `Delay::new`.

tower-rate-limit/tests/rate_limit.rs
```diff
@@ -3,73 +3,70 @@ extern crate tower_mock;
...
 
-use std::time::Duration;
-use std::thread;
+use std::time::{Duration, Instant};
 
 #[test]
 fn reaching_capacity() {
-    let (mut service, mut handle) =
-        new_service(Rate::new(1, from_millis(100)));
+    let mut rt = tokio::runtime::Runtime::new().unwrap();
```
carllerche: For tests, I would use the current thread runtime.

You also shouldn't need `lazy` anymore with the change I suggested above.

Instead of chaining `and_then`, I would make multiple sequential calls to `block_on`.

```rust

let one = rt.block_on(...);
let two = rt.block_on(...);
```

hawkw: Ah, good point!

hawkw: @carllerche
> Instead of chaining and_then, I would make multiple sequential calls to `block_on`.
 
I'm not sure if that works, since `tower_mock::Handle::poll_request` has to be called from within a task context. 

I could block on `future::lazy(move || { handle.poll_request() })`but then `Handle` would have to be inside an `Arc<Mutex<...>>` and that doesn't seem any better...

carllerche: current_thread’s block_on shouldn’t require static. You can use poll_fn even I think.

hawkw: Ah, you're right, it that does work -- 89666da31948493d9d83b7947154d40704a78b10 should make all the changes you requested.

seanmonstar:
Anything blocking this? It looks good at a glance

hawkw:
@seanmonstar this should be ready to merge whenever it's ✅ 'd.

PR #104, opened 2018-09-01
Add combinators for working with Services

fafhrd91:
* ServiceExt trait
* and_then, then, map, map_err, from_err, apply combinators

### DIFF
--- a/tower-util/src/ext/and_then.rs
+++ b/tower-util/src/ext/and_then.rs
@@ -1,4 +1,4 @@
-use futures::{Async, Future, Poll};
+use futures::{Future, Poll};
 use tower_service::Service;
 
 /// Service for the `and_then` combinator, chaining a computation onto the end of
@@ -13,8 +13,7 @@ pub struct AndThen<A, B> {
 impl<A, B> AndThen<A, B>
 where
     A: Service,
-    A::Error: Into<B::Error>,
-    B: Service<Request = A::Response> + Clone,
+    B: Service<Request = A::Response, Error = A::Error> + Clone,
 {
     /// Create new `AndThen` combinator
     pub fn new(a: A, b: B) -> AndThen<A, B> {
@@ -25,8 +24,7 @@ where
 impl<A, B> Service for AndThen<A, B>
 where
     A: Service,
-    A::Error: Into<B::Error>,
-    B: Service<Request = A::Response> + Clone,
+    B: Service<Request = A::Response, Error = A::Error> + Clone,
 {
     type Request = A::Request;
     type Response = B::Response;
@@ -34,11 +32,8 @@ where
     type Future = AndThenFuture<A, B>;
 
     fn poll_ready(&mut self) -> Poll<(), Self::Error> {
-        match self.a.poll_ready() {
-            Ok(Async::Ready(_)) => self.b.poll_ready(),
-            Ok(Async::NotReady) => Ok(Async::NotReady),
-            Err(err) => Err(err.into()),
-        }
+        let _ = try_ready!(self.a.poll_ready());
+        self.b.poll_ready()
     }
 
     fn call(&mut self, req: Self::Request) -> Self::Future {
@@ -49,8 +44,7 @@ where
 pub struct AndThenFuture<A, B>
 where
     A: Service,
-    A::Error: Into<B::Error>,
-    B: Service<Request = A::Response>,
+    B: Service<Request = A::Response, Error = A::Error>,
 {
     b: B,
     fut_b: Option<B::Future>,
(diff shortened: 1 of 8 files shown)


fafhrd91:
@carllerche

tower-util/src/ext/then.rs
```diff
@@ -0,0 +1,98 @@
...
+    A::Error: Into<B::Error>,
+    B: Service<Request = Result<A::Response, A::Error>> + Clone,
+{
+    type Request = A::Request;
+    type Response = B::Response;
+    type Error = B::Error;
+    type Future = ThenFuture<A, B>;
+
+    fn poll_ready(&mut self) -> Poll<(), Self::Error> {
+        match self.a.poll_ready() {
```
hawkw: Take it or leave it: I think this could just be 
```rust
let _ =  try_ready!(self.a.poll_ready());
self.b.poll_ready()
```

tower-util/src/ext/map.rs
```diff
@@ -0,0 +1,82 @@
...
+impl<T, F, R> Future for MapFuture<T, F, R>
+where
+    T: Service,
+    F: Fn(T::Response) -> R,
+{
+    type Item = R;
+    type Error = T::Error;
+
+    fn poll(&mut self) -> Poll<Self::Item, Self::Error> {
+        match self.fut.poll()? {
```
hawkw: This could just be
```rust
let resp = try_ready!(self.fut.poll());
Ok(Async::Ready((self.f)(resp)))
```

tower-util/src/ext/apply.rs
```diff
@@ -0,0 +1,50 @@
+use std::marker::PhantomData;
+
+use futures::{Future, IntoFuture, Poll};
+use tower_service::Service;
+
+/// `Apply` service combinator
+pub struct Apply<T, F, R, Req> {
+    service: T,
+    f: F,
+    r: PhantomData<Fn(Req) -> R>,
```
hawkw: Nit: since this field isn't used, it should probably have a leading `_`.

tower-util/src/ext/and_then.rs
```diff
@@ -0,0 +1,98 @@
...
+{
+    type Item = B::Response;
+    type Error = B::Error;
+
+    fn poll(&mut self) -> Poll<Self::Item, Self::Error> {
+        if let Some(ref mut fut) = self.fut_b {
+            return fut.poll();
+        }
+
+        match self.fut_a.poll() {
```
hawkw: This can just be
```rust
let resp = try_ready!(self.fut_a.poll());
self.fut_b = Some(self.b.call(resp));
self.poll()
```

tower-util/src/ext/and_then.rs
```diff
@@ -0,0 +1,98 @@
...
+    A::Error: Into<B::Error>,
+    B: Service<Request = A::Response> + Clone,
+{
+    type Request = A::Request;
+    type Response = B::Response;
+    type Error = B::Error;
+    type Future = AndThenFuture<A, B>;
+
+    fn poll_ready(&mut self) -> Poll<(), Self::Error> {
+        match self.a.poll_ready() {
```
hawkw: This could be:
```rust
let _ = try_ready!(self.a.poll_ready());
self.b.poll_ready()
```

tower-util/src/ext/and_then.rs
```diff
@@ -0,0 +1,98 @@
...
+/// This is created by the `ServiceExt::and_then` method.
+pub struct AndThen<A, B> {
+    a: A,
+    b: B,
+}
+
+impl<A, B> AndThen<A, B>
+where
+    A: Service,
+    A::Error: Into<B::Error>,
```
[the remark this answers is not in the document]
fafhrd91: just figured out inference problems.

tower-util/src/ext/mod.rs
```diff
@@ -0,0 +1,98 @@
...
+    /// the second service isn't called until call to the fist service have finished.
+    /// Result of the call to the first service is used as an input parameter
+    /// for the second service's call.
+    ///
+    /// Note that this function consumes the receiving service and returns a
+    /// wrapped version of it.
+    fn and_then<B, U>(self, service: U) -> AndThen<Self, B>
+    where
+        Self: Sized,
+        U: Into<B>,
```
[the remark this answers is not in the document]
fafhrd91: this would be useful for #69, but i found that `From/Into` does not work well with type inference.
we need to use custom `IntoService` trait. i can remove conversion or implement IntoService, it is up to you?

tower-util/src/ext/apply.rs
fafhrd91: i am not sure, we need `F` which is `Fn(Req, T) -> IntoFuture`. do i miss anything?

tower-util/src/ext/mod.rs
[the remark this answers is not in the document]
carllerche: Yeah, I think we should limit `Into` / `From` bounds due to type inference issues.

Let's remove the conversion for now.

Separately, I think `IntoService` will helpful at some point, but lets punt for this PR so that we can get it merged.

fafhrd91:
Yes, it is ready

carllerche:
Thanks 👍

Issue #108, opened 2018-10-11
Consider adding an argument to `NewService`


carllerche:
This has been requested in the past by @olix0r.

Doing so would allow providing additional context. Given that `NewService` is most often called to process a new connection, the argument could be used to pass connection level data for the service to consume.

Ideas:

* Remote peer's socket address.
* TLS context (https://github.com/tower-rs/tower-grpc/issues/67).

carllerche:
If this change is done, the best way to go about it will probably be to represent `NewService` as a "service of services". In other words, to implement a `NewService`, one would implement `Service` such that `Response: Service`.

To ease usage of this trait, the main "tower" crate can have a `NewService` "trait alias".

Also, what do people think about renaming `NewService` -> `MakeService` and use the `Make` prefix for "factories" instead of `New`.

The main problem I have is, what does this return:

```
fn build_new_service() -> ....
```
Does it return `T: Service` or `T: NewService`?

`fn build_make_service()` is a bit clearer.

Thoughts @seanmonstar @olix0r?

seanmonstar:
The name `NewServivce` does cause a little confusion. I'm fine with `MakeService`, I think.

seanmonstar:
There's also `Service::poll_ready`, which suggests there needing to be a `MakeService::poll_ready`. I actually think that makes sense, and allows for a server to say it cannot handle any more connections for now...

carllerche:
@seanmonstar yes, I think `poll_ready` makes sense, the question is about `&mut self`. However, `&mut self` makes sense with `poll_ready`.

PR #114, opened 2018-10-19
change NewService to MakeService<Target, Request>

seanmonstar:
Builds off the `generic-request` branch.

cc #108

### DIFF
--- a/tower-reconnect/src/lib.rs
+++ b/tower-reconnect/src/lib.rs
@@ -4,16 +4,17 @@ extern crate log;
 extern crate tower_service;
 
 use futures::{Future, Async, Poll};
-use tower_service::{Service, NewService};
+use tower_service::{Service, MakeService};
 
 use std::{error, fmt};
 
-pub struct Reconnect<T, Request>
+pub struct Reconnect<T, Target, Request>
 where
-    T: NewService<Request>,
+    T: MakeService<Target, Request>,
 {
-    new_service: T,
-    state: State<T, Request>,
+    context: Target,
+    mk_service: T,
+    state: State<T::Future, T::Service>,
 }
 
 #[derive(Debug)]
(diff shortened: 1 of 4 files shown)


seanmonstar:
This also changed `NewServiceFn` to just `ServiceFn`, since you can get a `MakeService` by having a `ServiceFn` that returns a `Service`.

One place where this was weird was with `Reconnect`, where it now has `ResponseFuture<T, Ctx, Request>`, even though it doesn't really need those generics otherwise...

tower-reconnect/src/lib.rs
```diff
@@ -23,47 +24,47 @@ pub enum Error<T, U> {
...
 }
 
 // ===== impl Reconnect =====
 
-impl<T, Request> Reconnect<T, Request>
+impl<T, Ctx, Request> Reconnect<T, Ctx, Request>
 where
-    T: NewService<Request>,
+    T: MakeService<Ctx, Request>,
+    Ctx: Clone,
```
seanmonstar: I'm thinking that instead of requiring `Clone` here, it'd be more flexible if the bounds were `T: MakeService<&Ctx, Request>`. Then, if the new `Service` needs a clone, it can make one, but it's otherwise not wasted.

jonhoo:
I'd like to see something in the docs about what `Ctx` is, what it's intended for, etc.

seanmonstar:
I explored if something like this could be used in hyper (allowing user's to grab the remote address) (and while not being a breaking change yet), and it works: https://github.com/hyperium/hyper/commit/94cc8065490497dcaf6bbb9862c0b1f28bf10126

Though I had to some nasty crap to get `S: for<'a> MakeService<&'a Io>` to work...

tower-reconnect/src/lib.rs
```diff
@@ -4,16 +4,17 @@ extern crate log;
...
 
-pub struct Reconnect<T, Request>
+pub struct Reconnect<T, Ctx, Request>
 where
-    T: NewService<Request>,
+    T: MakeService<Ctx, Request>,
 {
-    new_service: T,
-    state: State<T, Request>,
+    context: Ctx,
```
carllerche: If we are bike shedding, I prefer "target" over "context". I feel that it makes the intent a bit more clear.

tower-reconnect/src/lib.rs
```diff
@@ -23,47 +24,47 @@ pub enum Error<T, U> {
     NotReady,
 }
 
-pub struct ResponseFuture<T, Request>
+pub struct ResponseFuture<T, Ctx, Request>
```
carllerche: I know it is an unrelated change... but can this just be `ResponseFuture<T>` where `T` is the inner future?

tower-reconnect/src/lib.rs
carllerche: How did that pan out?

carllerche:
Can we 🚢 🇮🇹 ?

jonhoo:
Note to self: this affects #119.

tower-reconnect/src/lib.rs
seanmonstar: Not easily, since the error is mapped by `Error::Inner` which means the bounds are needed. Perhaps with some better refactoring...

jonhoo:
I'd still like to see some docs for what the `Target` param is intended for. Is it reasonable to make it `()`? Why/why not? For #119, I have a thing that makes new services on demand; how can it generate appropriate `Target` values for anything but `()`?

PR #117, opened 2018-11-05
Add Stream adapter for Discover

jonhoo:
This provides a convenient mechanism for taking a `Stream` over `Service` changes (i.e., `Change`) and using it as a `Discover`.

### DIFF
--- a/tower-discover/src/lib.rs
+++ b/tower-discover/src/lib.rs
@@ -5,14 +5,16 @@
 //! via the network; however, it is possible to discover services available in
 //! other processes or even in process.
 
+#[macro_use]
 extern crate futures;
 extern crate tower_service;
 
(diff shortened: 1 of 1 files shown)


tower-discover/src/lib.rs
```diff
@@ -82,6 +84,58 @@ where
     }
 }
 
+/// Dynamic service discovery based on a stream of service changes.
+pub struct Services<S, K, Svc> {
+    inner: S,
+    _marker_k: PhantomData<K>,
+    _marker_v: PhantomData<Svc>,
+    finished: bool,
```
carllerche: The `finished` logic is what `Fuse` does: https://docs.rs/futures/0.1.25/futures/future/struct.Fuse.html

Issue #110, opened 2018-10-16
Better way to drive Service than poll_ready


jonhoo:
I'm looking to implement `tokio-tower::pipeline::Client`, and as part of that I need to be able to "drive" the `Service` implementation forward to send pending requests, publish responses etc. Currently, the only way to do that is to do all the work in `poll_ready`. While that works just fine, it feels like a bit of an abuse of the API. @carllerche mentioned [here](https://gitter.im/tokio-rs/dev?at=5bc65da6f0a85e40a750fcd1) that we might want a `Service::drive` method where we do this kind of heavy-lifting.

carllerche:
In theory, I really want something like this as it would reduce the required overhead of using the Tower Service abstraction.

That said, I thought more about it, and it is a bit more complicated...

The "drive" function would have to do two things:

1) Drive the internal state of the `Service`
2) Notify the caller when it is safe to drop the `Service` (i.e., the socket is closed and all response futures have completed).

The second point also implies that there needs to be some way to tell the service to start shutting down.

So, at the very least, there would need to be two additional functions:

* Some "drive" function.
* Some function to signal the service should start shutting down.

(Note that `Sink` provides APIs to achieve this).

This adds non trivial complexity to using a `Service`. It would be unfortunate if *all* users of `Service` are subject to this additional API.

This implies that there would need to be a separate trait. How would that work?

jonhoo:
I don't really have an answer, just more observations about why the current situation isn't ideal: if you do work to "drive" the returned future(s) in `poll_ready`, you now end up in the weird situation where you need to basically alternate between polling the future and calling `poll_ready` until the former resolves. That's a pretty awkward pattern (maybe you have a nice way to deal with it?). It could be that all `tower` needs to provide is some nice way to say "drive the `Service` until this future resolves`?

jonhoo:
The other advantage to have `Service: Future` (at least in some cases) is that I could stick a bunch of them in `FuturesUnordered` or something like that. In my case, I have multiple `tokio_tower::pipeline::Client`, one to each "shard" of a database. I send a request on some subset of them, and now I need to drive them *all* until I the reply futures resolve...

carllerche:
If you stash them in FuturesUnordered how do you call them?

jonhoo:
`FuturesUnordered::iter_mut`, but it's obviously not ideal. `FuturesOrdered` may be needed to give predictable access to the services.

jonhoo:
[...]

PR #118, opened 2018-11-05
Add the DirectService trait

jonhoo:
This patch adds the `DirectService` trait, and related implementations over it in `tower_balance` and `tower_buffer`, as discussed in #110. `DirectService` is similar to a `Service`, but must be "driven" through calls to `poll_service` for the futures returned by `call` to make progress.

The motivation behind adding this trait is that many current `Service` implementations spawn long-running futures when the service is created, which then drive the work necessary to turn requests into responses. A simple example of this is a service that writes requests over a `TcpStream` and reads responses over that same `TcpStream`. The underlying stream must be read from to discover new responses, but there is no single entity to drive that task. The returned futures would share access to the stream (and worse yet, may get responses out of order), and then service itself is not guaranteed to see any more calls to it as the client is waiting for its requests to finish.

`DirectService` solves this by introducing a new method, `poll_service`, which must be called to make progress on in-progress futures. Furthermore, like `Future::poll`, `poll_service` must be called whenever the associated task is notified so that the service can also respect time-based operations like heartbeats.

The PR includes changes to both `tower_balance::Balance` and `tower_buffer::Buffer` to add support for wrapping `DirectService`s. For `Balance` this is straightforward: if the inner service is a `Service`, the `Balance` also implements `Service`; if the inner service is a `DirectService`, the `Balance` is itself also a `DirectService`. For `Buffer`, this is more involved, as a `Buffer` turns any `DirectService` *into* a `Service`. The `Buffer`'s `Worker` is spawned, and will therefore drive the wrapped `DirectService`.

One complication arises in that `Buffer<T>` requires that `T: Service`, but you can safely construct a `Buffer` over a `DirectService` per the above. `Buffer` works around this by exposing

```rust
impl Service for HandleTo<S> where S: DirectService {}
```

And giving out `Buffer<HandleTo<S>>` when the `new_directed(s: S)` constructor is invoked. Since `Buffer` never calls any methods on the service it wraps, `HandleTo`'s implementation just consists of calls to `unreachable!()`.

Note that `tower_buffer` now also includes a `DirectedService` type, which is a wrapper around a `Service` that implements `DirectService`. In theory, we could do away with this by adding a blanket impl:

```rust
impl<T> DirectedService for T where T: Service {}
```

but until we have specialization, this would prevent downstream users from implementing `DirectService` themselves.

Finally, this also makes `Buffer` use a bounded mpsc channel, which introduces a new capacity argument to `Buffer::new`.


jonhoo:
For context, the ultimate goal of this is to enable [`tokio-tower`](https://github.com/jonhoo/tokio-tower-wip/tree/direct_service), which has convenience bindings for writing servers and clients using `tower` + `tokio`. It's similar to the old `tokio-proto`, but written in terms of `Service` (and now `DirectService`).

tower-balance/src/lib.rs
```diff
@@ -300,31 +345,103 @@ where
...
+        self.call(D::Service::call, request)
+    }
 
-        let rsp = svc.call(request);
-        ResponseFuture(rsp, PhantomData)
+    fn poll_service(&mut self) -> Result<Async<()>, Self::Error> {
+        let ready = &mut self.ready;
+        let not_ready = &mut self.not_ready;
+        let mut err = None;
+        self.pending.retain(|k| {
```
carllerche: I don't see where any keys are inserted into this set.

jonhoo: Ah, yes, you're completely right. How silly of me. Sadly, I think we'll have to require `D::Key: Clone` to be able to maintain `pending` in a reasonable fashion. Maybe we should just get rid of the optimization?

tower-buffer/src/lib.rs
```diff
@@ -42,18 +40,85 @@ pub struct ResponseFuture<T> {
...
+        Ok(Async::Ready(()))
+    }
+
+    fn call(&mut self, req: Request) -> Self::Future {
+        self.inner.call(req)
+    }
+}
+
+/// A handle to a `DirectService` that has been spawned elsewhere.
+pub struct HandleTo<T, Request>
```
carllerche: This can be removed by removing `T: Service` fro m`Buffer`. `Buffer` only needs to be generic over the request and the future.

jonhoo: It does mean that I end up with a bunch of code that looks like this:
```rust
foo: Buffer<FooRequest, <FooSvc<A> as Service<FooRequest>>::Future>,
```

jonhoo: It gets particularly noisy when you have multiple layers of `Service` combinators, especially with #109 because now the `Request` type has to be named multiple type.

tower-buffer/src/lib.rs
```diff
@@ -42,18 +48,87 @@ pub struct ResponseFuture<T> {
...
+
+    fn call(&mut self, _: Request) -> Self::Future {
+        unreachable!("tried to call a DirectService through a marker reference")
+    }
+}
+
+/// A wrapper that exposes a `Service` (which does not need to be driven) as a `DirectService` so
+/// that a construct that is *able* to take a `DirectService` can also take instances of
+/// `Service`.
+pub struct DirectedService<T, Request>
```
seanmonstar: This could just be `DirectedService<T>`, if the bound is removed, yea?

tower-buffer/src/lib.rs
```diff
@@ -42,18 +48,87 @@ pub struct ResponseFuture<T> {
...
+
+impl<T, Request> DirectService<Request> for DirectedService<T, Request>
+where
+    T: Service<Request>,
+{
+    type Response = T::Response;
+    type Error = T::Error;
+    type Future = T::Future;
+
+    fn poll_ready(&mut self) -> Result<Async<()>, Self::Error> {
```
seanmonstar: Just a small thing, but noticed this was `Result<..>` instead of `Poll<..>`, any particular reason?

PR #119, opened 2018-11-06
Add a dynamically-sized load-balanced pool

jonhoo:
This patch adds a new type, `Pool`, which wraps a `tower_balance::Balance` and a `tower_service::NewService` together so that new `Service` instances are added when load is high, and removed again if load is low.

The pool uses an exponentially weighted moving average of successful calls to `poll_ready` on the underlying `Balance` to estimate whether there are enough services available. If `poll_ready` frequently returns `NotReady`, then a new service is produced, whereas if `poll_ready` pretty much never returns `NotReady`, the most recently added service is removed from the pool (down to a minimum of 1).

### DIFF
--- a/tower-balance/src/pool.rs
+++ b/tower-balance/src/pool.rs
@@ -1,0 +1,295 @@
+//! This module defines a load-balanced pool of services that adds new services when load is high.
+//!
+//! The pool uses `poll_ready` as a signal indicating whether additional services should be spawned
+//! to handle the current level of load. Specifically, every time `poll_ready` on the inner service
+//! returns `Ready`, [`Pool`] consider that a 0, and every time it returns `NotReady`, [`Pool`]
+//! considers it a 1. [`Pool`] then maintains an [exponential moving
+//! average](https://en.wikipedia.org/wiki/Moving_average#Exponential_moving_average) over those
+//! samples, which gives an estimate of how often the underlying service has been ready when it was
+//! needed "recently" (see [`Builder::urgency`]). If the service is loaded (see
+//! [`Builder::loaded_above`]), a new service is created and added to the underlying [`Balance`].
+//! If the service is underutilized (see [`Builder::underutilized_below`]) and there are two or
+//! more services, then the latest added service is removed. In either case, the load estimate is
+//! reset to its initial value (see [`Builder::initial`] to prevent services from being rapidly
+//! added or removed.
+#![deny(missing_docs)]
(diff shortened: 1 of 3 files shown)


carllerche:
What is the status of this PR? In progress, needs review, or should it be closed? (it still says untested, so it sounds like a work in progress at the moment).

jonhoo:
@carllerche renamed it to WIP. I anticipate it to be mostly stable, but haven't done enough testing to warrant merging yet.

tower-balance/src/pool.rs
```diff
@@ -0,0 +1,280 @@
...
+{
+    type Key = usize;
+    type Service = MS::Service;
+    type Error = MS::MakeError;
+
+    fn poll(&mut self) -> Poll<Change<Self::Key, Self::Service>, Self::Error> {
+        if let Load::High = self.load {
+            if self.making.is_none() {
+                try_ready!(self.maker.poll_ready());
+                // TODO: it'd be great if we could avoid the clone here and use, say, &Target
```
LucioFranco: yeah I think for now this is a pattern we will have to live with 😞

tower-balance/src/pool.rs
```diff
@@ -0,0 +1,280 @@
...
+                self.load = Load::Normal;
+                let rm = self.services;
+                self.services -= 1;
+                Ok(Async::Ready(Change::Remove(rm)))
+            }
+        }
+    }
+}
+
+/// A
```
LucioFranco: unexpected new line here?

tower-balance/src/pool.rs
```diff
@@ -0,0 +1,280 @@
...
+        mut make_service: MS,
+        target: Target,
+        choose: C,
+    ) -> Pool<C, MS, Target, Request>
+    where
+        MS: MakeService<Target, Request>,
+        Target: Clone,
+        C: Choose<usize, MS::Service>,
+    {
+        let s = make_service.make_service(target.clone());
```
LucioFranco: not totally sure but is it correct to make the service in build instead of on the first call?

jonhoo:
@LucioFranco Great, thanks! I think I've addressed all your comments. Feel free to merge once CI passes if you agree :tada:

tower-balance/src/pool.rs
carllerche: In theory, `Target` should be "light"

PR #120, opened 2018-11-06
Make `Buffer` use a bounded channel

jonhoo:
This change moves `Buffer` from `mpsc::unbounded` to `mpsc::channel`. The primary motivation for this change is that bounded channels provide back-pressure to callers, so that `Balance<Buffer>` for example works as expected. Currently, `Buffer` will accept as many requests as you can make for it without ever stopping down, slowly eating up all your memory.

### DIFF
--- a/tower-balance/examples/demo.rs
+++ b/tower-balance/examples/demo.rs
@@ -222,7 +222,7 @@ impl Discover for Disco {
     fn poll(&mut self) -> Poll<Change<Self::Key, Self::Service>, Self::Error> {
         match self.changes.pop_front() {
             Some(Change::Insert(k, svc)) => {
-                let svc = Buffer::new(svc, &self.executor).unwrap();
+                let svc = Buffer::new(svc, 0, &self.executor).unwrap();
                 let svc = InFlightLimit::new(svc, ENDPOINT_CAPACITY);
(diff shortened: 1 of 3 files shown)


jonhoo:
I'm still looking at the failing test, so don't merge until that's resolved.

jonhoo:
Interesting.. The test that hangs on Travis fails for me even on `master` (at least 90%+ of the time). @carllerche can you reproduce?

carllerche:
Hmm... I can try to investigate

jonhoo:
Doing a little more digging: on `master`, `clears_canceled_requests` fails for me in ~50% of runs with
```
thread 'clears_canceled_requests' panicked at 'called `Option::unwrap()` on a `None` value'
```
This always happens at the `unwrap` at
```rust
// res2 was dropped, so it should have been canceled in the buffer
handle.allow(1);

let req3 = handle.next_request().unwrap();
```
I don't know the mocking code well enough to understand what's going on.

jonhoo:
@seanmonstar I think it's trying to figure out the failing test, but since it also happens on `master`, it's probably then not a blocker

jonhoo:
#132 seems to have fixed the failing test, so I think we're all good to merge!

PR #122, opened 2018-11-08
Make Buffer::new use DefaultExecutor

### LINKED ISSUE #121: Have `Buffer` use default executor
`Buffer` requires an executor to spawn tasks. Currently, constructing a buffer requires an explicit executor to be passed in. Instead, `Buffer::new` should default to [`DefaultExecutor`](https://docs.rs/tokio/0.1.12/tokio/executor/struct.DefaultExecutor.html).

Though, currently `Buffer::new` spawns the task eagerly in `new`, but often times the `Buffer` instance is created from outside of the runtime **before**, in which case `DefaultExecutor` won't work.

There would be a new constructor fn `Buffer::new_with_executor` (though, the name might be improved) that takes an executor.

`Buffer` should also be updated to avoid exposing the concrete task type that needs to be spawned. We can opt for the strategy described [here](https://github.com/tokio-rs/tokio/issues/625#issuecomment-429563091).

jonhoo:
Fixes #121.

### DIFF
--- a/tower-balance/examples/demo.rs
+++ b/tower-balance/examples/demo.rs
@@ -56,34 +56,30 @@ fn main() {
     println!("]");
 
     let mut rt = runtime::Runtime::new().unwrap();
-    let executor = rt.executor();
 
-    let exec = executor.clone();
     let fut = future::lazy(move || {
         let decay = Duration::from_secs(10);
-        let d = gen_disco(exec.clone());
+        let d = gen_disco();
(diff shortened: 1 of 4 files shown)


jonhoo:
Here are the same changes on top of #118: https://github.com/jonhoo/tower/commit/02fb6bbfcd2ed88bf9cccbdd894b6a47c2ffa362

PR #125, opened 2018-11-15
Seal the MakeService trait

seanmonstar:
Otherwise, people may try to implement it directly (like I did), and then later places that simply depend on `Service<Target>` will reject those implementations.


jonhoo:
Oh, interesting.. Why should users not implement `MakeService`?

hawkw:
@jonhoo Users shouldn't try to implement it because there's a blanket implementation for all `T: Service` where `T::Response: Service`, the easiest way to implement `MakeService` is just to implement a `Service` that returns `Service`s. 

Perhaps this should be documented as well?

PR #130, opened 2018-11-28
Adjust retry Budget to overflow less

seanmonstar:


tower-retry/src/budget.rs
```diff
@@ -40,29 +40,38 @@ struct Generation {
...
+            // If there is no percent, then you gain nothing from deposits.
+            // Withdrawals can only be made against the reserve, over time.
+            (0, 1)
+        } else if retry_percent <= 1.0 {
+            (1, (1.0 / retry_percent) as isize)
+        } else {
+            // Support for when retry_percent is between 1.0 and 1000.0,
+            // meaning for every deposit D, D*retry_percent withdrawals
+            // can be made.
+            (1000, (1000.0 / retry_percent) as isize)
```
adleong: Is this right that the withdraw amount depends on the retry percent while deposit amount remains fixed?  Imagine that the retry_percent is very low.  In that case the withdraw amount would be very high and the reserve could be depleted in just one request.

Should, instead, the withdraw amount stay constant and the deposit amount depend on the retry percent?

adleong: Oh wait, the reserve also depends on the retry percent?  this is very confusing...

adleong: Is it done this way so that can use integer counters *and* work with percentages above and below 1.0?

Would it simplify things to keep floating point values in the buckets instead?

seanmonstar: Yes, it's done this way to be able to use integers instead of floats (and specifically, there is an `AtomicIsize`, but no atomic float types). This strategy is also used in finagle's `RetryBucket`/`WindowedAdder`.

tower-retry/src/budget.rs
```diff
@@ -40,29 +40,38 @@ struct Generation {
...
+        assert!(retry_percent <= 1000.0);
+        assert!(min_per_sec < ::std::i32::MAX as u32);
+
+
+        let (deposit_amount, withdraw_amount) = if retry_percent == 0.0 {
+            // If there is no percent, then you gain nothing from deposits.
+            // Withdrawals can only be made against the reserve, over time.
+            (0, 1)
+        } else if retry_percent <= 1.0 {
+            (1, (1.0 / retry_percent) as isize)
```
adleong: what happens if this cast overflows? ie if the retry_percent is extremely small

Issue #131, opened 2018-11-29
Consider changing middleware errors to Box<dyn Error>


seanmonstar:
## The Problem

Most of the various middlewares in tower define an `enum Error<E> { Inner(E), .. }` to return from `Service`. A problem starts to arise once you compose multiple middlewares together, but still want to be somewhat generic over handling the errors. Look at this gem:

```
error: Inner(Inner(Inner(B(Service(Inner(A(B(Error { kind: Connect, details: "interesting" }))))))))
```

In this case, we've wrapped a `hyper::Client` in a bunch of middleware. In order to inspect and react to the error (in my case, decide if the error qualifies to retry the request), I could:

- Set my expect error type exactly, and thus use `match e { UnNestAllTheLayers => () }`. This is very fragile, since adding new middleware, or changing the order, will cause compilation to fail. It's also hugely tedious being sure we've match every correct combination.
- Make some `trait CanRetry` and implement it for every single intermediary type. Adding a new middleware means we need to implement `CanRetry` for the potentially new nested error enum.

## Proposal

A different solution *might* be better: remove the `enum Error<E>`, and instead having all middleware just return `Box<dyn std::error::Error + Send + Sync>`.

The code needed to inspect errors if they were dynamic trait objects changes to something like this:

```rust
if let Some(e) = error.downcast_ref::<hyper::Error>() {
    if e.is_connect() { // or whatever
        return CanRetry::Yes;
    }
}
```

## Example

Consider [`tower-in-flight-limit`](https://github.com/tower-rs/tower/blob/master/tower-in-flight-limit/src/lib.rs):

```rust
impl<S, Req> Service<Req> for InFlightLimit<S>
where
    S: Service<Req>,
    S::Error: Into<Box<dyn StdError + Send + Sync>>,
{
    type Response = S::Response;
    type Error = Box<dyn StdError + Send + Sync>;

    // ...
}

impl<T> Future for ResponseFuture<T>
where
    T: Future,
    T::Error: Into<Box<dyn StdError + Send + Sync>>,
{
    type Item = T::Item;
    type Error = Box<dyn StdError + Send + Sync>;

    fn poll(&mut self) -> Poll<Self::Item, Self::Error> {
        match self.inner {
            Some(ref mut f) => {
                match f.poll() {
                    // ...
                    Err(e) => {
                        self.shared.release();
                        Err(e.into())
                    }
                }
            }
            None => Err(NoCapacity.into()),
        }
    }
}
```

[...]

PR #141, opened 2019-01-09
Use `tokio::timer` instead of `tokio_timer`; use Kind pattern

davidbarsky:
As @carllerche requested:

> Carl Lerche
> could you submit a PR that changes error to be a struct? https://github.com/tower-rs/tower/blob/master/tower-timeout/src/lib.rs#L26
>
> David Barsky
> sure with inner ErrorKind?
>
> Carl Lerche
> see here: https://github.com/tokio-rs/tokio/blob/master/tokio-timer/src/timeout.rs
> basically
> yeah, ErrorKind, but that is private https://github.com/tokio-rs/tokio/blob/master/tokio-timer/src/timeout.rs#L228-L287

### DIFF
--- a/tower-timeout/Cargo.toml
+++ b/tower-timeout/Cargo.toml
@@ -7,4 +7,8 @@ publish = false
 [dependencies]
 futures = "0.1"
 tower-service = { version = "0.2", path = "../tower-service" }
-tokio-timer = "0.2.6"
+
+[dependencies.tokio]
+version = "0.1.14"
+default-features = false
+features = ["timer"]
--- a/tower-timeout/src/lib.rs
+++ b/tower-timeout/src/lib.rs
@@ -3,16 +3,20 @@
 //! If the response does not complete within the specified timeout, the response
 //! will be aborted.
 
+#![doc(html_root_url = "https://docs.rs/tower-timeout/0.1.0")]
+#![deny(missing_debug_implementations, missing_docs)]
+#![cfg_attr(test, deny(warnings))]
+
 extern crate futures;
 extern crate tower_service;
-extern crate tokio_timer;
+extern crate tokio;
 
 use futures::{Future, Poll, Async};
 use tower_service::Service;
-use tokio_timer::{clock, Delay};
+use tokio::timer::{Delay, Error as TimerError};
 
 use std::{error, fmt};
-use std::time::Duration;
+use std::time::{Duration, Instant};
 
 /// Applies a timeout to requests.
 #[derive(Debug)]
@@ -23,13 +27,21 @@ pub struct Timeout<T> {
 
 /// Errors produced by `Timeout`.
 #[derive(Debug)]
-pub enum Error<T> {
-    /// The inner service produced an error
+pub struct Error<T>(Kind<T>);
+
+/// Timeout error variants
+#[derive(Debug)]
+enum Kind<T> {
+    /// Inner value returned an error
     Inner(T),
 
-    /// The request did not complete within the specified timeout.
-    Timeout,
+    /// The timeout elapsed.
+    Elapsed,
+
+    /// Timer returned an error.
+    Timer(TimerError),
 }
+
 
 /// `Timeout` response future
 #[derive(Debug)]
@@ -41,6 +53,7 @@ pub struct ResponseFuture<T> {
 // ===== impl Timeout =====
 
 impl<T> Timeout<T> {
+    /// Creates a new Timeout
     pub fn new(inner: T, timeout: Duration) -> Self {
         Timeout {
             inner,
@@ -59,13 +72,15 @@ where
 
     fn poll_ready(&mut self) -> Poll<(), Self::Error> {
         self.inner.poll_ready()
-            .map_err(Error::Inner)
+            .map_err(|e| {
+                Error(Kind::Inner(e))
+            })
     }
 
     fn call(&mut self, request: Request) -> Self::Future {
         ResponseFuture {
             response: self.inner.call(request),
-            sleep: Delay::new(clock::now() + self.timeout),
+            sleep: Delay::new(Instant::now() + self.timeout),
         }
     }
 }
@@ -83,14 +98,14 @@ where T: Future,
         match self.response.poll() {
             Ok(Async::Ready(v)) => return Ok(Async::Ready(v)),
             Ok(Async::NotReady) => {}
-            Err(e) => return Err(Error::Inner(e)),
+            Err(e) => return Err(Error(Kind::Inner(e))),
         }
 
         // Now check the sleep
         match self.sleep.poll() {
             Ok(Async::NotReady) => Ok(Async::NotReady),
-            Ok(Async::Ready(_)) => Err(Error::Timeout),
-            Err(_) => Err(Error::Timeout),
+            Ok(Async::Ready(_)) => Err(Error(Kind::Elapsed)),
+            Err(_) => Err(Error(Kind::Elapsed)),
         }
     }
 }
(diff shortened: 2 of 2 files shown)


tower-timeout/src/lib.rs
```diff
@@ -59,13 +72,15 @@ where
...
+            .map_err(|e| {
+                Error(Kind::Inner(e))
+            })
     }
 
     fn call(&mut self, request: Request) -> Self::Future {
         ResponseFuture {
             response: self.inner.call(request),
-            sleep: Delay::new(clock::now() + self.timeout),
+            sleep: Delay::new(Instant::now() + self.timeout),
```
carllerche: Why is this changed?

tower-timeout/src/lib.rs
```diff
@@ -23,14 +27,22 @@ pub struct Timeout<T> {
...
-    /// The request did not complete within the specified timeout.
-    Timeout,
+    /// The timeout elapsed.
+    Elapsed,
+
+    /// Timer returned an error.
+    Timer(TimerError),
 }
 
+
```
carllerche: Extra line?

seanmonstar:
We could leave the dependency on `tokio-timer`, right?

davidbarsky:
@seanmonstar Yeah, but tokio-timer (https://github.com/tokio-rs/tokio-timer) is deprecated, right?

seanmonstar:
It's just moved directly into the tokio workspace: https://github.com/tokio-rs/tokio/tree/master/tokio-timer

tower-timeout/src/lib.rs
davidbarsky: Fixed locally.

tower-timeout/src/lib.rs
davidbarsky: Fixed.

davidbarsky:
With regard to @carllerche's comment on:

> How about having: 
```rust
enum Kind<T> {
    Inner(T),
    Timeout(timer::TimeoutError),
}
```

I think it might make sense to re-export [`tokio::timer::timeout::Error`](https://docs.rs/tokio/0.1.14/tokio/timer/timeout/struct.Error.html) as tower-timeout's error. Internally, `timeout::Error`'s Kind is defined as:

```
// Timeout error variants
#[derive(Debug)]
enum Kind<T> {
    /// Inner value returned an error
    Inner(T),

    /// The timeout elapsed.
    Elapsed,

    /// Timer returned an error.
    Timer(::Error),
}
```

...where Timer(::Error) wrap `tokio::timer::Error`, which handles “something's gone terribly wrong”-style of error with timers. Thoughts?

tower-timeout/src/lib.rs
```diff
@@ -83,50 +98,114 @@ where T: Future,
...
 impl<T> error::Error for Error<T>
 where
-    T: error::Error,
+    T: error::Error + 'static,
 {
-    fn cause(&self) -> Option<&error::Error> {
-        if let Error::Inner(ref why) = *self {
+    fn description(&self) -> &str {
+        match self.0 {
+            Kind::Inner(_) => "inner service error",
```
hawkw: Since `T` is known to implement `Error`, this should probably be
```suggestion
            Kind::Inner(e) => e.description(),
```

tower-timeout/src/lib.rs
```diff
@@ -83,50 +98,114 @@ where T: Future,
...
+            Kind::Elapsed => "request timed out",
+            Kind::Timer(_) => "internal timer error"
```
hawkw: Again, this should probably be
```suggestion
            Kind::Timer(e) => e.description(),
```

tower-timeout/src/lib.rs
```diff
@@ -83,50 +98,114 @@ where T: Future,
...
 
-    fn description(&self) -> &str {
-        match *self {
-            Error::Inner(_) => "inner service error",
-            Error::Timeout => "request timed out",
+// ===== impl Error =====
+
+impl<T> Error<T> {
+    /// Create a new `Error` representing the inner value completing with `Err`.
+    pub fn inner(err: T) -> Error<T> {
```
hawkw: TIOLI: I'm not sure if it's necessary for these constructors to be `pub`?

tower-timeout/Cargo.toml
```diff
@@ -7,4 +7,8 @@ publish = false
 [dependencies]
 futures = "0.1"
 tower-service = { version = "0.2", path = "../tower-service" }
-tokio-timer = "0.2.6"
```
hawkw: Unless there's a specific reason for this change, my understanding is that it's best for libraries to depend on the component crates, rather than `tokio`, especially since the 0.2.x branch of `tokio-timer` is _not_ deprecated.

davidbarsky: Ah, I confused https://github.com/tokio-rs/tokio/blob/master/tokio-timer/ with https://github.com/tokio-rs/tokio-timer. Sorry about that. I'll revert the changes.

tower-timeout/src/lib.rs
hawkw: I do see a potential use for being able to construct these in test code, though... 🤷‍♀️

tower-timeout/src/lib.rs
davidbarsky: fixed, ty

tower-timeout/src/lib.rs
carllerche: It's fine for it to pub for testing.

hawkw: Okay, let's make it pub, then!

olix0r: Should this be guarded by `#[cfg(test)]` then?

seanmonstar: When *depending* on a crate, even in tests, the dependency will not have been compiled with `--cfg test`. So setting `#[cfg(test)]` would only make it usable in unit tests within `tower-timeout`.

carllerche: @olix0r that would only be enabled when running lib tests in the crate proper.

If it is useful for downstream dependencies' tests, then it is part of the public API.

davidbarsky:
> @davidbarsky are you suggesting to not have tower-timeout provide an error type, instead reuse tokio-timer's timeout::Error type?
> Usually, I avoid this as it is a forwards compat hazard. However, in this case, issuing breaking changes isn't too hard. I don't have much of an opinion either way.

Yep, I was suggesting that.

> In #131, @seanmonstar suggests to avoid having errors generic over the inner error and use Box<dyn Error> instead. If this happens, then the error types would deviate. Either way, switching to Box<dyn Error> would require a breaking change.

I'm increasingly somewhat sympathetic to Sean's suggestion there in #131. A generic wrapper the one we're working with make it kind of tricky (or at the very least, _tedious_) to compose Tower middleware.

tower-timeout/src/lib.rs
```diff
@@ -83,50 +97,114 @@ where T: Future,
...
+            Err(e) => return Err(Error(Kind::Inner(e))),
         }
 
         // Now check the sleep
         match self.sleep.poll() {
             Ok(Async::NotReady) => Ok(Async::NotReady),
-            Ok(Async::Ready(_)) => Err(Error::Timeout),
-            Err(_) => Err(Error::Timeout),
+            Ok(Async::Ready(_)) => Err(Error(Kind::Elapsed)),
+            Err(_) => Err(Error(Kind::Elapsed)),
```
seanmonstar: I realize this code was already "wrong", but I believe this should be `Kind::Timer(err)`.

davidbarsky: Nice catch, fixed.

seanmonstar:
@davidbarsky go ahead and squash/merge when you're ready :)

PR #142, opened 2019-01-10
balance: Update `rand` dependency

hawkw:
This branch updates `tower-balance` to depend on the latest 
released `rand` version.


hawkw:
Required by linkerd/linkerd2-proxy#170.

PR #143, opened 2019-01-11
Expose Service errors through tower-buffer

jonhoo:
In the past, any errors thrown by a `Service` wrapped in a `tower_buffer::Buffer` were silently swallowed, and the handles were simply informed that the connection to the `Service` was closed.

This patch captures errors from a wrapped `Service`, and communicates that error to all pending and future requests. It does so by wrapping up the error in an `Arc`, which is sent to all pending `oneshot` request channels, and is stored in a shared location so that future requests will see the error when their send to the `Worker` fail.

Note that this patch also removes the `open` field from `State`, as it is no longer necessary following #120, since bounded channels have a `try_ready` method we can rely on instead.

Note that this change is not entirely backwards compatible -- the error type for a `Service` that is wrapped in `Buffer` must now be `Send + Sync` so that it can safely be communicated back to callers. Furthermore, `tower_buffer::Error::Closed` now contains the error that the failed `Service` produced, which may trip up old code.

### DIFF
--- a/tower-buffer/src/lib.rs
+++ b/tower-buffer/src/lib.rs
@@ -45,10 +45,20 @@ pub struct ResponseFuture<T, E> {
 /// An error produced by a `Service` wrapped by a `Buffer`
 #[derive(Debug)]
 pub struct ServiceError<E> {
+    method: &'static str,
+    inner: E,
+}
+
+impl<E> ServiceError<E> {
     /// The method that was called on `Service` when it failed.
-    pub method: &'static str,
-    /// The error produced by the `Service`.
-    pub inner: E,
+    pub fn method(&self) -> &'static str {
+        self.method
+    }
+
+    /// The error produced by the `Service` when `method` was called.
+    pub fn error(&self) -> &E {
+        &self.inner
+    }
 }
 
 /// Errors produced by `Buffer`.
--- a/tower-buffer/src/lib.rs
+++ b/tower-buffer/src/lib.rs
@@ -222,6 +222,14 @@ where
             Err(DirectedService(service)) => Err(SpawnError { inner: service }),
         }
     }
+
+    fn get_error_on_closed(&self) -> Arc<ServiceError<T::Error>> {
+        self.state
+            .err
+            .borrow()
+            .cloned()
+            .expect("Worker exited, but did not set error.")
+    }
 }
 
 impl<T, Request> Buffer<DirectServiceRef<T>, Request>
@@ -263,15 +271,9 @@ where
 
     fn poll_ready(&mut self) -> Poll<(), Self::Error> {
         // If the inner service has errored, then we error here.
-        self.tx.poll_ready().map_err(move |_| {
-            Error::Closed(
-                self.state
-                    .err
-                    .borrow()
-                    .cloned()
-                    .expect("Worker exited, but did not set error."),
-            )
-        })
+        self.tx
+            .poll_ready()
+            .map_err(move |_| Error::Closed(self.get_error_on_closed()))
     }
 
     fn call(&mut self, request: Request) -> Self::Future {
@@ -284,14 +286,8 @@ where
         match self.tx.try_send(Message { request, tx }) {
             Err(e) => {
                 if e.is_disconnected() {
-                    let e = self
-                        .state
-                        .err
-                        .borrow()
-                        .cloned()
-                        .expect("Worker exited, but did not set error.");
                     ResponseFuture {
-                        state: ResponseState::Failed(e),
+                        state: ResponseState::Failed(self.get_error_on_closed()),
                     }
                 } else {
                     ResponseFuture {
--- a/tower-buffer/src/lib.rs
+++ b/tower-buffer/src/lib.rs
@@ -58,6 +58,8 @@ pub enum Error<E> {
     Inner(E),
     /// The underlying `Service` failed. All subsequent requests will fail.
     Closed(Arc<ServiceError<E>>),
+    /// The underlying `Service` is currently at capacity; wait for `poll_ready`.
+    Full,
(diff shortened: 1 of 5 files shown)


tower-buffer/src/lib.rs
```diff
@@ -264,15 +277,19 @@ where
...
     fn poll_ready(&mut self) -> Poll<(), Self::Error> {
         // If the inner service has errored, then we error here.
-        if !self.state.open.load(Ordering::Acquire) {
-            return Err(Error::Closed);
-        } else {
-            self.tx.poll_ready().map_err(|_| Error::Closed)
-        }
+        self.tx.poll_ready().map_err(move |_| {
+            Error::Closed(
+                self.state
```
hawkw: Nit: this code for cloning the error seems repeated between `poll_ready` and `call`, maybe it should be factored out into a separate function?

tower-buffer/src/lib.rs
```diff
@@ -437,16 +509,23 @@ where
                             }
                             // We may want to also make progress on current requests
                         }
-                        Err(_) => {
-                            self.state.open.store(false, Ordering::Release);
-                            return Ok(().into());
```
hawkw: I think no longer finishing the worker future in this case seems correct, but it does look like a possible change in behaviour here --- the worker will remain alive until the service is dropped...

tower-buffer/src/lib.rs
```diff
@@ -31,25 +30,36 @@ pub struct Buffer<T, Request>
...
+    /// The method that was called on `Service` when it failed.
+    pub method: &'static str,
+    /// The error produced by the `Service`.
+    pub inner: E,
 }
 
 /// Errors produced by `Buffer`.
 #[derive(Debug)]
-pub enum Error<T> {
+pub enum Error<E> {
```
hawkw: TIOLI: While we're here, should we maybe change this into a struct for forwards-compatibility reasons?

tower-buffer/src/lib.rs
jonhoo: Ah, no, the worker will still exit. It just won't exit until it has sent errors out to all outstanding requests. Specifically, `self.failed` will close `self.rx`, which means that `poll_next_msg` will eventually return `Ready(None)`, which will set `self.finish = true`, which will cause the `Worker` to exit. We need to not exit immediately (like the old code did), because then we'd just be dropping the `oneshot::Sender` for any outstanding requests, which would then not get notified of the error!

tower-buffer/src/lib.rs
jonhoo: I'm not sure why we'd want this to be a struct? If you're thinking of #131, then we could still expose an enum as a `dyn Error` in the future?

tower-buffer/src/lib.rs
hawkw: Thanks for clearing that up, the flow got a little more complicated here. I thought that the worker would continue to error any _new_ requests, but I think I misread it.

tower-buffer/src/lib.rs
```diff
@@ -222,6 +222,14 @@ where
             Err(DirectedService(service)) => Err(SpawnError { inner: service }),
         }
     }
+
+    fn get_error_on_closed(&self) -> Arc<ServiceError<T::Error>> {
```
hawkw: 👍

tower-buffer/src/lib.rs
jonhoo: Yeah, new requests will see the error by observing that the request `mpsc` has been closed, and then reading out the `Arc<Error>` that the `Worker` sets before closing the receive end of the channel

tower-buffer/src/lib.rs
carllerche: @jonhoo definitely can be part of a separate PR.

jonhoo:
So, the use-case for me is actually to have insight into service failures during _development_. It's really hard to debug a broken service that errors out for one reason or another when then error is just swallowed. We *could* just `panic` on service errors, but that seems bad too. So the most readily apparent alternative to me was to expose the error to the `Buffer`s. I sympathize with the concern that this may repeat the error many times downstream though.

Here's an alternative proposal: make `Buffer` require that `Service::Error = ()` for the wrapped `Service`. This communicates to the user that _they_ have to deal with service errors. It's a little less user-friendly, but mitigates the concern you had for error replication.

seanmonstar:
@jonhoo hm, that may be a good idea indeed. It's sort of like requiring `Future<Error = ()>` on `Executor`.

jonhoo:
Actually, on second thought, that doesn't quite work. We still want to communicate back errors that occur on `call`. It's only errors in the `poll_*` methods that render the service unusable... Sadly they all share a single error type, so we can't require that the service error is `()` without also requiring that `call`'s returned future has a `()` error type :/

tower-buffer/src/lib.rs
```diff
@@ -31,25 +30,36 @@ pub struct Buffer<T, Request>
...
-    state: ResponseState<T>,
+pub struct ResponseFuture<T, E> {
+    state: ResponseState<T, E>,
+}
+
+/// An error produced by a `Service` wrapped by a `Buffer`
+#[derive(Debug)]
+pub struct ServiceError<E> {
+    /// The method that was called on `Service` when it failed.
+    pub method: &'static str,
```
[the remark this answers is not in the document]
jonhoo: The idea was that the user may want to look at them, but I guess we could make them accessors instead.

carllerche:
I'm fine w/ this PR, i would wait for @seanmonstar's +1 though.

jonhoo:
Great! I'll merge once CI passes following the fix for the bug that @hawkw pointed out :tada:

PR #144, opened 2019-01-12
Move MakeService to DirectService for the inner service

LucioFranco:
This is one of many first steps in an attempt to move tower to `DirectService`, for right now only the service produced from the MakeService is a `DirectService`. This is needed to support `MakeService for https://github.com/LucioFranco/tower-hyper and future leaf service implementations.

Questions:

Should I also move the maker service to `DirectService`?

### DIFF
--- a/tower-util/Cargo.toml
+++ b/tower-util/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "tower-util"
-version = "0.1.0"
+version = "0.2.0"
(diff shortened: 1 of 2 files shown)


LucioFranco:
I don't know how we can slowly move to `DirectBuffer` without one large PR since bumping the version to indicate a breaking change will cause other crates to fail. I guess I could remove the bump but that might affect crates downstream that depend on the git version but I guess they should really probably be using crates.

PR #145, opened 2019-01-12
buffer: Add with_executor_direct and Error into_inner

LucioFranco:
I've added a couple useful functions to buffer to allow easy construction of `DirectService` backed buffers. I've also included an `into_inner` helper for `Error<T>` to easily fetch the inner service error, I don't think this is the right way to go about it but it made some things easy and I don't think it bloats the api.

Examples of these changes can be found here https://github.com/LucioFranco/tower-hyper/blob/master/examples/client.rs


tower-buffer/src/lib.rs
```diff
@@ -494,6 +507,15 @@ where
 
 // ===== impl Error =====
 
+impl<T> Error<T> {
```
LucioFranco: I'm fine with removing this in favor of the changes in #143

PR #147, opened 2019-01-14
Document tower-retry

seanmonstar:


davidbarsky:
I like these changes! @seanmonstar: Would you like me to open a PR to introduce an example where I implement retry on `http::Request<T> where T: Into<Body> + Clone` and where I have a client that mostly uses `http::Request<Bytes>` as the primary request type?

davidbarsky:
(It'll consist of copying code over from https://github.com/awslabs/aws-lambda-rust-runtime/blob/strawman-api/lambda-runtime-client-simple/src/hyper_tower.rs#L51)

seanmonstar:
@davidbarsky More examples is likely useful, but on the other hand, adding hyper as a dev-dependency would slow down compiling/testing... Maybe just a link?

PR #149, opened 2019-01-18
Add initial base tower crate

LucioFranco:
This adds the initial base tower crate, as of right now it contains
nothing and is only needed to ensure that cargo workspaces can
properly compile with rust 1.32.

Issue #136, opened 2018-12-13
Consider promoting DirectService to Service


carllerche:
The [`DirectService`](https://github.com/tower-rs/tower/blob/ffa6f0361874076a197cdaf4d48b22532ad94c1d/tower-direct-service/src/lib.rs#L26) trait (introduced in #118) is an alternate version of `Service` that is being experimented with. It provides the necessary hooks to "drive" a service's internal state machine forward. This allows avoiding to have to spawn new tasks to power service implementations.

We should consider whether or not this trait should be promoted to be the primary `Service` trait.

#### Driven vs. buffered

When introducing `DirectService`, we extensively discussed whether or not the functionality should just be provided by `Service`. We concluded that it should not. The primary reason is that middleware is not always able to drive the inner service. For example, a router cannot effectively call `poll_service` as needed. Because of this, there are service implementations that driven (require `poll_service` to be called) and there are service implementations that are buffered (requests get dispatched on a channel to another task). These services do not require `poll_service` to be called. Initially, we could not figure out a good way to implement middleware that could indicate which kind of service it wanted, so we opted for two completely separate traits.

It occurred to me that we could combine `DirectService` with `Service` and use `Clone` as an indicator of whether or not the service is driven vs. buffered. The router must already accept `T: Service + Clone`. The strategy taken is, when a request comes in, the request is routed to an inner `T: Service` and that `T` is cloned into the router's response future. `T::poll_ready` is called by the response future. Services that are unable to call `poll_ready` will all most likely take a `T: Service + Clone`.

So, the proposal is to make `Service` what `DirectService` is today. `Service` implementations like `Buffer` would have no-op `poll_service` and `poll_close` implementations.

#### Complexity

The other con is that implementing `Service` becomes more complicated. This is a real drawback, but I believe it can be mitigated by providing utilities for common patterns:

* `service_fn` for easily defining leaf services.
* Combinators and other utilities for common middleware patterns.

For example, basic logging could be implemented as:

```rust
service_fn(|request| process(request))
    .before(|request| println!("request = {:?}", request))
    .after(|response| println!("response = {:?}", response))
```

And more complicated:

```rust
service_fn(|request| ...)
    .around(|request, next| {
        let request = request.clone();

        next.call(request)
            .inspect(move |response| {
                println!("request = {:?}; response={:?}", request, response);
            })
    })
```

### Roadmap

The promotion could be done in a few steps.

[...]

PR #150, opened 2019-01-18
Add Default Implementations for DirectService

davidbarsky:
This might make the task described in #136 easier. Namely, this addresses some of the “complexity” for services that _don't_ need to spawn long-running futures, such as `Buffer`, certain kinds of middleware, and leaf services.

Note: this will need to be rebased atop of #149 to build successfully.

### DIFF
--- a/tower-direct-service/src/lib.rs
+++ b/tower-direct-service/src/lib.rs
@@ -15,7 +15,7 @@
 
 extern crate futures;
 
-use futures::{Future, Poll};
+use futures::{Async, Future, Poll};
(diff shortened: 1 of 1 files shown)


davidbarsky:
Closing because: 
- There's no mechanism of taking _back_ a default in a library without a breaking change.
- Helper functions provide the same or greater degree of utility as default implementations while avoiding a large number of [rakes](https://tvtropes.org/pmwiki/pmwiki.php/Main/RakeTake).

PR #151, opened 2019-01-19
Add MakeConnection trait alias

LucioFranco:
This adds a bare-bones connection creation service trait alias. The goal behind this is to allow libraries like `tower-h2`, `tower-hyper` and `tokio-tower`. To establish the underlying transport in a composable way. This also replaces https://github.com/carllerche/tokio-connect.

Example usage of this type of service alias:

https://github.com/tower-rs/tower-hyper/blob/master/src/client/connect.rs#L88

and 

https://github.com/tower-rs/tower-h2/pull/48

cc @carllerche @jonhoo

### DIFF
--- a/tower-util/src/make_connection.rs
+++ b/tower-util/src/make_connection.rs
@@ -1,0 +1,42 @@
+use futures::Future;
+use tokio_io::{AsyncRead, AsyncWrite};
+use tower_service::Service;
+
+/// The MakeConnection trait is used to create transports
+///
+/// The goal of this service is to allow composable methods for creating
+/// `AsyncRead + AsyncWrite` transports. This could mean creating a TLS
+/// based connection or using some other method to authenticate the connection.
+pub trait MakeConnection<Request> {
+    /// The transport provided by this service
+    type Response: AsyncRead + AsyncWrite;
+
+    /// Errors produced by the connecting service
+    type Error;
+
+    /// The future that eventually produces the transport
+    type Future: Future<Item = Self::Response, Error = Self::Error>;
+
+    /// Connect and return a transport asynchronously
+    fn make_connection(&mut self, target: Request) -> Self::Future;
+}
+
+impl<S, Request> self::sealed::Sealed<Request> for S where S: Service<Request> {}
+
+impl<C, Request> MakeConnection<Request> for C
+where
+    C: Service<Request>,
+    C::Response: AsyncRead + AsyncWrite,
+{
+    type Response = C::Response;
+    type Error = C::Error;
+    type Future = C::Future;
+
+    fn make_connection(&mut self, target: Request) -> Self::Future {
+        Service::call(self, target)
+    }
+}
+
+mod sealed {
+    pub trait Sealed<A> {}
+}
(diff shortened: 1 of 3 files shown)


LucioFranco:
@jonhoo I think there should be another trait alias that is specific to `tokio-tower` to convert from a `AsyncRead + AsyncWrite` to a `Stream + Sink`. One example would be a trait that takes a `AsyncRead + AsyncWrite` and uses `async-bincode` to generate the stream. I feel as this is something quite specific to `tokio-tower` since `tower-hyper` and `tower-h2` both accept `AsyncRead + AsyncWrite` not `Stream + Sink`.

That said, `MakeTransport` could be modified to take a `ConnectService`.

```rust
pub trait MakeTransport<Target, Request>: self::sealed::Sealed<Target, Request> 
where
     Target: AsyncRead + AsyncWrite
{
    /// Items produced by the transport
    type Item;

    /// Errors produced when receiving from the transport
    type Error;

    /// Errors produced when sending to the transport
    type SinkError;

    /// The `Sink + Stream` implementation created by this factory
    type Transport: Stream<Item = Self::Item, Error = Self::Error>
        + Sink<SinkItem = Request, SinkError = Self::SinkError>;

    /// Errors produced while building a transport.
    type MakeError;

    /// The future of the `Service` instance.
    type Future: Future<Item = Self::Transport, Error = Self::MakeError>;

    /// Returns `Ready` when the factory is able to create more transports.
    ///
    /// If the service is at capacity, then `NotReady` is returned and the task
    /// is notified when the service becomes ready again. This function is
    /// expected to be called while on a task.
    ///
    /// This is a **best effort** implementation. False positives are permitted.
    /// It is permitted for the service to return `Ready` from a `poll_ready`
    /// call and the next invocation of `make_transport` results in an error.
    fn poll_ready(&mut self) -> Poll<(), Self::MakeError>;

    /// Create and return a new transport asynchronously.
    fn make_transport(&mut self, target: Target) -> Self::Future;
```

The benefit here is that we can create a core set of `ConnectService`s that can do things like resolve DNS, negotiate TLS, etc and then have `MakeTransport`s that know how to implement the protocol framing etc. This would allow some nice composability. So its kinda like composable middleware for creating services.

I have start on [`tower-resolve`](https://github.com/LucioFranco/tower-resolve/blob/master/src/lib.rs#L38) crate that composes something that resovles with something that connects and implements `ConnectService`. So we could do the same for `MakeTransport`.

LucioFranco:
@jonhoo ah yes, I don't quite know we should couple them, The thing that connects them there is that `Target: AsyncRead + AsyncWrite`. We could ideally just create a kinda `Connector` struct that takes a `MakeConnection` and a `MakeTransport` and composes them together. Still kinda working that part out, I used `tower-resolve` as an example way to compose these types of traits. Suggestions are welcome though. 😄

tower-util/src/make_connection.rs
```diff
@@ -0,0 +1,42 @@
...
+    type Response: AsyncRead + AsyncWrite;
+
+    /// Errors produced by the connecting service
+    type Error;
+
+    /// The future that eventually produces the transport
+    type Future: Future<Item = Self::Response, Error = Self::Error>;
+
+    /// Connect and return a transport asynchronously
+    fn connect(&mut self, target: Request) -> Self::Future;
```
carllerche: Could this be renamed `make_connection`?

LucioFranco: done!

PR #153, opened 2019-01-28
Remove tower-router

carllerche:
The current implementation is not ideal and there is no plan to work on
it in the near term.


carllerche:
@LucioFranco No immediate plan. There are many different ways to implement routing. As of now, I'd say libs that need routing should do it themselves. We can then see if there is a way to extract common code.

PR #155, opened 2019-02-02
Update links in the README

Ophirr33:
I'm potentially interested in helping out on tower, figured a good place to start would be updating some links in the README.

Added tower-direct-service and tower-retry, removed tower-router. Looks like `DirectService` is going to get merged into `Service` eventually, but it'll be easy enough to delete the links when that happens.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,9 +27,6 @@ crates.
   handle the next request, `tower-buffer` stores the request in an internal
   queue ([docs][tbuf-docs]).
 
-* [`tower-direct-service`]: Similar to the `Service` trait, but allows for the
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -27,6 +27,9 @@ crates.
   handle the next request, `tower-buffer` stores the request in an internal
   queue ([docs][tbuf-docs]).
 
+* [`tower-direct-service`]: Similar to the `Service` trait, but allows for the
```
LucioFranco: Probably don't want to add this just now since we might not be going through with this idea.

carllerche:
I removed direct-service

PR #156, opened 2019-02-06
[WIP] Convert Tower's middleware to return Boxed Errors

davidbarsky:
This PR _begins_ implementation of #131, but does not finish it. So far, this PR converted:

- `tower-inflight-limit`
- `tower-rate-limit`
- `tower-timeout`

I ran into issues on the following: 

- Some middleware defines errors that have _two_ generic parameters or more than two variants. In those cases, I still think the error should be collapsed into a _single_ boxed error, but I think thee's value in preserving some form of an error type that can be downcasted. In those cases, I think the correct approach is to defined multiple _public_ error structs that an end-library user can downcast to.
- Tower Mock makes interesting uses of oneshot channels. I'm not sure what the correct approach is there (note—the WIP tower-mock commit _does not_ compile).

### DIFF
--- a/tower-buffer/src/lib.rs
+++ b/tower-buffer/src/lib.rs
@@ -14,13 +14,13 @@ extern crate tower_direct_service;
 extern crate tower_service;
 
 use futures::future::Executor;
-use tokio_sync::mpsc;
-use tokio_sync::oneshot;
 use futures::{Async, Future, Poll, Stream};
 use std::marker::PhantomData;
 use std::sync::Arc;
-use std::{error, fmt};
+use std::{error::Error as StdError, fmt};
 use tokio_executor::DefaultExecutor;
+use tokio_sync::mpsc;
+use tokio_sync::oneshot;
 use tower_direct_service::DirectService;
 use tower_service::Service;
 
@@ -43,6 +43,8 @@ pub struct ResponseFuture<T, E> {
     state: ResponseState<T, E>,
 }
 
+type Error = Box<StdError + Send + Sync>;
+
 /// An error produced by a `Service` wrapped by a `Buffer`
 #[derive(Debug)]
 pub struct ServiceError<E> {
@@ -57,15 +59,51 @@ impl<E> ServiceError<E> {
     }
 }
 
-/// Errors produced by `Buffer`.
+/// The underlying `Service` is currently at capacity; wait for `poll_ready`.
 #[derive(Debug)]
-pub enum Error<E> {
-    /// The `Service` call errored.
-    Inner(E),
-    /// The underlying `Service` failed. All subsequent requests will fail.
-    Closed(Arc<ServiceError<E>>),
-    /// The underlying `Service` is currently at capacity; wait for `poll_ready`.
-    Full,
+pub struct FullError;
+
+impl StdError for FullError {
+    fn source(&self) -> Option<&(dyn StdError + 'static)> {
+        Some(self)
+    }
+}
+
+impl fmt::Display for FullError {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "Service at capacity")
+    }
+}
+
+/// The underlying `Service` failed. All subsequent requests will fail.
+#[derive(Debug)]
+pub struct ClosedError<E: Into<Error>>(Arc<ServiceError<E>>);
+
+impl<E> ClosedError<E>
+where
+    E: Into<Error> + fmt::Display + fmt::Debug + 'static,
+{
+    pub fn error(&self) -> &Arc<ServiceError<E>> {
+        &self.0
+    }
+}
+
+impl<E> StdError for ClosedError<E>
+where
+    E: Into<Error> + fmt::Display + fmt::Debug + 'static,
+{
+    fn source(&self) -> Option<&(dyn StdError + 'static)> {
+        Some(self)
+    }
+}
+
+impl<E> fmt::Display for ClosedError<E>
+where
+    E: Into<Error> + fmt::Display + fmt::Debug + 'static,
+{
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "Service::{} failed: {}", self.0.method, self.0.inner)
+    }
 }
 
 /// An adapter that exposes the associated types of a `DirectService` through `Service`.
@@ -270,16 +308,18 @@ where
 impl<T, Request> Service<Request> for Buffer<T, Request>
 where
     T: Service<Request>,
+    <T as Service<Request>>::Error: Into<Error> + Send + Sync + 'static,
+    <T as Service<Request>>::Error: fmt::Display + fmt::Debug,
 {
     type Response = T::Response;
-    type Error = Error<T::Error>;
+    type Error = Error;
     type Future = ResponseFuture<T::Future, T::Error>;
 
     fn poll_ready(&mut self) -> Poll<(), Self::Error> {
         // If the inner service has errored, then we error here.
         self.tx
             .poll_ready()
-            .map_err(move |_| Error::Closed(self.get_error_on_closed()))
+            .map_err(move |_| ClosedError(self.get_error_on_closed()).into())
     }
 
     fn call(&mut self, request: Request) -> Self::Future {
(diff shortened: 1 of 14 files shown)


seanmonstar:
(Just so you don't think I'm ignoring, I'm offline until next week. I just stole a sneak at my inbox today... Bad Sean...)

davidbarsky:
Sean—no worries. Enjoy your time off!

tower-buffer/src/lib.rs
```diff
@@ -270,16 +308,18 @@ where
...
-    type Error = Error<T::Error>;
+    type Error = Error;
     type Future = ResponseFuture<T::Future, T::Error>;
 
     fn poll_ready(&mut self) -> Poll<(), Self::Error> {
         // If the inner service has errored, then we error here.
         self.tx
             .poll_ready()
-            .map_err(move |_| Error::Closed(self.get_error_on_closed()))
+            .map_err(move |_| ClosedError(self.get_error_on_closed()).into())
```
hawkw: TIOLI: It might be slightly tidier if you just modified `get_error_on_closed` to return a `ClosedError<T::Error>`, rather than having to wrap it here?

tower-buffer/src/lib.rs
```diff
@@ -337,21 +379,21 @@ where
 
             match self.state {
                 Full => {
-                    return Err(Error::Full);
+                    return Err(FullError.into());
                 }
                 Failed(ref e) => {
-                    return Err(Error::Closed(e.clone()));
+                    return Err(ClosedError(e.clone()).into());
```
hawkw: TIOLI: you could avoid having to explicitly call `into` here (and elsewhere) if you used the `?` operator instead:
```rust
Failed(ref e) => Err(ClosedError(e.clone()))?,
```

tower-buffer/tests/buffer.rs
```diff
@@ -95,9 +95,9 @@ fn when_inner_fails() {
     ::std::thread::sleep(::std::time::Duration::from_millis(100));
     with_task(|| {
         let e = res1.poll().unwrap_err();
-        if let Error::Closed(e) = e {
+        if let Some(e) = e.downcast_ref::<ClosedError<tower_mock::Error>>() {
             assert!(format!("{:?}", e).contains("poll_ready"));
-            assert_eq!(e.error(), &tower_mock::Error::Other("foobar"));
+            // assert_eq!(e.error(), &tower_mock::Error.downcast_ref::);
```
hawkw: This works but is somewhat ugly:
```suggestion
            assert_eq!(format!("{}", e.error().error()), "foobar".to_owned());
```

Unfortunately we can't downcast `e.error()` to `str`, as `downcast_ref` and `is` only work for `Sized` types, but the impl of `Error` is for `str` and not `&'a str`...

tower-buffer/src/lib.rs
```diff
@@ -57,15 +59,51 @@ impl<E> ServiceError<E> {
...
+/// The underlying `Service` is currently at capacity; wait for `poll_ready`.
 #[derive(Debug)]
-pub enum Error<E> {
-    /// The `Service` call errored.
-    Inner(E),
-    /// The underlying `Service` failed. All subsequent requests will fail.
-    Closed(Arc<ServiceError<E>>),
-    /// The underlying `Service` is currently at capacity; wait for `poll_ready`.
-    Full,
+pub struct FullError;
```
hawkw: TIOLI, but I would probably call this `Full` (and call `ClosedError` just `Closed`); IMO the impl of `Error` makes it pretty clear that this is an error...

tower-buffer/src/lib.rs
```diff
@@ -43,6 +43,8 @@ pub struct ResponseFuture<T, E> {
     state: ResponseState<T, E>,
 }
 
+type Error = Box<StdError + Send + Sync>;
```
hawkw: This is a bit of a nit, but I'd prefer we just refer to this type explicitly, and avoid the renaming import of `std::error::Error`...

tower-mock/src/lib.rs
```diff
@@ -307,22 +327,19 @@ impl<T, E> Respond<T, E> {
...
-                    Err(_) => Err(Error::Closed),
-                }
-            }
-            Error::NoCapacity => Err(Error::NoCapacity),
-            Error::Closed => Err(Error::Closed),
+        match self.rx.poll() {
+            Ok(Async::Ready(Ok(v))) => Ok(v.into()),
+            Ok(Async::Ready(Err(e))) => Err(e.into()),
+            Ok(Async::NotReady) => Ok(Async::NotReady),
+            Err(e) => Err(e.into()),
```
hawkw: Again, the question mark operator is probably our friend here, I think this match could be something like
```rust
match self.rx.poll()? {
    Async::Ready(ready) => Ok(ready?),
    Async::NotReady => Ok(Async::NotReady),
}
```

tower-buffer/src/lib.rs
carllerche: I personally prefer this (though I may have just referenced `std` directly instead of having a `use` statement, but that matters less to me. Repeatedly referencing `Box<Error + Send + Sync>` feels like code dup to me.

tower-buffer/src/lib.rs
carllerche: I am indifferent either way. 

That said, wouldn't the type be private after this change?

tower-buffer/src/lib.rs
davidbarsky: Oh, that'd be a good improvement. Most of the stuff I was doing here are _mostly_ mechanical.

tower-buffer/tests/buffer.rs
davidbarsky: That's way better, thanks!

tower-buffer/src/lib.rs
davidbarsky: I did `Box<Error + Send + Sync>` first, and then I switched to this type alias. I think the type alias definition being:

```rust
type Error = Box<std::error::Error + Send + Sync>
```
...would probably be best. Optionally, it could be placed in the root Tower crate, but I don't feel too strongly about it.

tower-mock/src/lib.rs
davidbarsky: True, yes—I'll switch it over to `?`.

tower-filter/src/lib.rs
```diff
@@ -99,11 +120,15 @@ impl<T, U> Filter<T, U> {
 }
 
 impl<T, U, Request> Service<Request> for Filter<T, U>
-where T: Service<Request> + Clone,
-      U: Predicate<Request>,
+where
+    T: Service<Request> + Clone,
+    T::Error: Into<Error>,
+    U: Predicate<Request>,
+    <U as Predicate<Request>>::Error: Into<Error>,
```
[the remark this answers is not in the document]
davidbarsky: Good point—it'll clean up the type signatures too.

tower-filter/src/lib.rs
```diff
@@ -99,11 +120,15 @@ impl<T, U> Filter<T, U> {
...
+    <T as Service<Request>>::Error: Into<Error>,
```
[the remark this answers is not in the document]
davidbarsky: I _don't_ recall—I'll double-check.

tower-buffer/src/lib.rs
```diff
@@ -57,15 +59,51 @@ impl<E> ServiceError<E> {
...
-    /// The `Service` call errored.
-    Inner(E),
-    /// The underlying `Service` failed. All subsequent requests will fail.
-    Closed(Arc<ServiceError<E>>),
-    /// The underlying `Service` is currently at capacity; wait for `poll_ready`.
-    Full,
+pub struct FullError;
+
+impl StdError for FullError {
+    fn source(&self) -> Option<&(dyn StdError + 'static)> {
```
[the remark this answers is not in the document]
davidbarsky: Good point, fixed.

tower-buffer/src/lib.rs
hawkw: @carllerche no, I don't think so, if the user wishes to downcast or use `error.source().is::<tower_buffer::Full>(), etc..

tower-buffer/src/lib.rs
hawkw: It's the renaming import that I'm really not fond of, but it's not particularly important to me...

tower-buffer/src/lib.rs
carllerche: I would probably put it in an `error` module... like https://docs.rs/tokio-sync/0.1.1/tokio_sync/mpsc/error/index.html

not sure if it should be `error` or `errors`.

carllerche:
Thanks for doing the initial work. This change has been applied to master over a number of PRs.

PR #161, opened 2019-02-21
Require `poll_ready` to be called before `call`

carllerche:
This updates the `Service` contract requiring `poll_ready` to be called
before `call`. This allows `Service::call` to panic in the event the
user of the service omits `poll_ready` or does not wait until `Ready` is
observed.

### DIFF
--- a/tower-buffer/src/lib.rs
+++ b/tower-buffer/src/lib.rs
@@ -58,8 +58,6 @@ pub enum Error<E> {
     Inner(E),
     /// The underlying `Service` failed. All subsequent requests will fail.
     Closed(Arc<ServiceError<E>>),
-    /// The underlying `Service` is currently at capacity; wait for `poll_ready`.
-    Full,
 }
 
 mod sealed {
@@ -113,7 +111,6 @@ struct State<E> {
 }
 
 enum ResponseState<T, E> {
-    Full,
     Failed(Arc<ServiceError<E>>),
     Rx(oneshot::Receiver<Result<T, Arc<ServiceError<E>>>>),
     Poll(T),
@@ -202,9 +199,7 @@ where
                         state: ResponseState::Failed(self.get_error_on_closed()),
                     }
                 } else {
-                    ResponseFuture {
-                        state: ResponseState::Full,
-                    }
+                    panic!("buffer full; poll_ready must be called first");
                 }
             }
             Ok(_) => ResponseFuture {
@@ -242,9 +237,6 @@ where
             let fut;
 
             match self.state {
-                Full => {
-                    return Err(Error::Full);
-                }
                 Failed(ref e) => {
                     return Err(Error::Closed(e.clone()));
                 }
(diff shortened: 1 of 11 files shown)


tower-buffer/src/lib.rs
```diff
@@ -202,9 +199,7 @@ where
                         state: ResponseState::Failed(self.get_error_on_closed()),
                     }
                 } else {
-                    ResponseFuture {
-                        state: ResponseState::Full,
-                    }
+                    panic!("buffer full; poll_ready must be called first");
```
seanmonstar: I know in Linkerd2, we make use of buffers in places where calling `poll_ready` isn't feasible or possible. This change would crash the proxy instead of denying the request.

carllerche: Ah yes, linkerd used this strategy before we came up w/ a good alternative. The strategy for linkerd is to clone the request into the response future (I believe the router is the primary spot for this).

seanmonstar: Clone the request? Or clone the `Service`?

jonhoo:
I think https://github.com/tower-rs/tower/blob/25eb52153d895377c9606b22cc4bff5b7f4193df/tower-buffer/src/lib.rs#L192-L200 can then also be fixed up a little?

tower-buffer/src/lib.rs
seanmonstar: Another possible pitfall: there isn't a guarantee that `poll_ready` means that `call` could succeed. If the service is backed by a channel, and a user calls `poll_ready` and in-between being notified it is ready, the receiver could be dropped. There'd be no way to guard against the panic in that case...

carllerche: poll_ready returning `Ready` does not mean that the call will succeed. It is fine for `poll_ready` to return `Ready` and `call` to fail the request.

carllerche:
@seanmonstar 

The motivation is to have more defined contract for the `Service` trait. Prior to this change, `Service` was defined to say "you sometimes need to call poll_ready before `call`... but sometimes you don't.. check the implementation's documentation for the specifics of how to use it". This is not great.

With this change, the contract is clearer: you must always get `Ready` from `poll_ready` before calling `call`.

panic vs. `Err` is more philisophical. IMO, bugs in code that result in violation of API contracts **should** result in a panic. The code path in question will never succeed and could result in corruption of the value. In which case, the service would need to maintain internal state saying "yo, i'm busted".

Also, I much prefer panics when there is a bug because you get a backtrace! If `call` returns `Err` (future that errors), that future will bubble way back up the call stack and best case you will get a log entry in your server saying "this request failed". A lot of the context will be lost.

Ideally, the panic is isolated to the task that is handling the request and does not impact unrelated tasks in the process.

I prefer to use reserve `Result` for error scenarios that happen despite correct code due to a runtime situation.

Also, while the `call` function still needs a branch for the panic (though it can *probably* be implemented as an `assert!` which would be clearer), the `ResponseFuture` struct is smaller and its state machine is simpler.

tower-buffer/src/lib.rs
seanmonstar: I was thinking of the behavior of `futures::sync`, where just because a sender saw ready from `poll_ready` doesn't mean a different sender couldn't use the slot and fill the buffer first. tokio-sync's channel will reserve the slot, so this can't happen.

What do you think of adding a comment here explaining this, so we know why this panic is justified?

carllerche: I added a comment.

PR #163, opened 2019-02-21
layer: Add `tower-layer` and the `Layer` trait

LucioFranco:
This change introduces the new `tower-layer` crate and the foundational `Layer` trait to go along with it. This trait allows one to easily compose a set of `Service`s that take an inner service. These services only modify the request/response. This also provides the `Layer` implementation for many of the tower crates.


tower-middleware/Cargo.toml
```diff
@@ -0,0 +1,25 @@
...
+# - Update html_root_url.
+# - Update documentation URL
+# - Create "v0.x.y" git tag.
+version = "0.1.0"
+authors = ["Carl Lerche <me@carllerche.com>"]
+license = "MIT"
+readme = "README.md"
+repository = "https://github.com/tower-rs/tower"
+homepage = "https://github.com/tower-rs/tower"
+documentation = "https://docs.rs/tokio-middleware/0.2.0"
```
carllerche: This should be tower-middleware/0.1.0.

carllerche:
I added @olix0r and @seanmonstar as reviewers to get their feedback too.

davidbarsky:
Do you think providing an example/documentation of how to compose middleware is out of scope for this PR?

carllerche:
@LucioFranco Any chance you could add some impls of `Middleware` for the various crates? That would help show what it looks like.

carllerche:
@davidbarsky I'd be happy to take examples / docs.

Anyone may also create PRs against *this* PR too w/ suggested changes / additions.

tower-middleware/src/lib.rs
```diff
@@ -0,0 +1,118 @@
...
+///
+/// Take request logging as an example:
+///
+/// ```rust
+/// # extern crate futures;
+/// # extern crate tower_service;
+/// # extern crate tower_web;
+/// # #[macro_use] extern crate log;
+/// use tower_service::Service;
+/// use tower_web::middleware::Middleware;
```
davidbarsky: I'm not sure that Tower-web should be a dependency in tower itself. This is a typo, correct?

tower-middleware/Cargo.toml
LucioFranco: good catch! Thanks!

tower-middleware/src/lib.rs
LucioFranco: Yup, I fixed the example, let me know if its a bit clearer.

LucioFranco:
I'd like to provide one crate migrated to this middleware before we merge then we can migrate the rest after.

carllerche:
@LucioFranco sounds like a good plan.

carllerche:
We still need to explore how this fits together with `MakeService`.

tower-middleware/src/lib.rs
```diff
@@ -0,0 +1,118 @@
...
+///         info!(target: self.target, "request = {:?}", request);
+///         self.service.call(request)
+///     }
+/// }
+/// ```
+///
+/// The above log implementation is decoupled from the underlying protocol and
+/// is also decoupled from client or server concerns. In other words, the same
+/// log middleware could be used in either a client or a server.
+pub trait Middleware<S> {
```
olix0r: Should `S` be constrained? i.e.`Middleware<R, S: Service<R>>` ?  My hunch is that lacking this constraint will cause some discomfort for implementors. But I'd have to think a bit more to articulate that

tower-middleware/Cargo.toml
```diff
@@ -0,0 +1,25 @@
+[package]
+name = "tower-middleware"
```
olix0r: are we convinced that "middleware" is the term we want to use? I'm still partial to "stack", but not enough to argue about it if there are folks who think middleware is better.

tower-middleware/LICENSE
```diff
@@ -0,0 +1,25 @@
+Copyright (c) 2018 Carl Lerche
```
olix0r: Should this be The Tower Authors?

carllerche: Probably

tower-middleware/src/lib.rs
carllerche: I don't think we can constrain it due to coherence. Or, if we can, it is just a huge PITA to do so w/o *much* win.

tower-middleware/Cargo.toml
carllerche: 🚲 🏠 😄 

I only reach for "middleware" because that is how its been referred to in prior ecosystems I've been involved with.

Pro: "middleware" has precedence. 
Con: It is long, possibly overloaded?

Stack feels like it refers to the stack of multiple middleware + service. Implementations of this trait are a "layer" of the stack. So, if anything `Layer` would be better... but that feels vague.

tower-middleware/LICENSE
carllerche:
```suggestion
Copyright (c) 2019 Tower Contributors
```

Here you go

tower-middleware/Cargo.toml
hawkw: I'm moderately pro-`Layer` personally. I think it's even reasonable to call the crate `tower-middleware` but the _trait_ `Layer` --- consider "a middleware layer which...."

LucioFranco: I guess for me personally I don't have strong feelings either way. I am just worried about our use of the term "middleware" as we use that in the readme. One thing to consider is what the API might look like for newer users, I think we should be careful how we use that terminology. 

### Current
```rust
ServiceBuilder::new(hyper)
        .middleware(InFlightLimitMiddleware::new(5))
        .middleware(RateLimitMiddleware::new(5, Duration::from_secs(1)))
        .build(dst);
```

### Layer
```rust
ServiceBuilder::new(hyper)
        .push(InFlightLimitLayer::new(5))
        .push(RateLimitLayer::new(5, Duration::from_secs(1)))
        .build(dst);
```

I do kind of like the name layer as its a bit more clear and differentiates between the high-level goal that tower is trying to provide and the direct goal you are trying to achieve by _layering_ the middleware.

And in that sense too, I would rather the crate be called `tower-layer` to make it a bit more clear. I would imagine someone reading the readme and seeing half the crate descriptions start with "Middleware that..." then seeing a `tower-middleware` might get a bit confused.

Plus, `Layer<S, Request>` is a lot less typing 😅 .

tower-middleware/src/lib.rs
LucioFranco: I'm not sure but I had to constrain it here but afaik I would need to continue to use constraints on all the impls even if it is on the core trait or not.

https://github.com/LucioFranco/tower/blob/service-builder/tower-in-flight-limit/src/lib.rs#L148

tower-middleware/Cargo.toml
olix0r: I'm +1 on Layer. Coincidentally, this is what I chose for a similar trait in linkerd2-proxy.

carllerche: I'm fine w/ `Layer` as well. It fits well.

carllerche: The crate *probably* should be renamed `tower-stack` at that point? So, a "stack" is a set of layers w/ a service.

davidbarsky: A benefit of a stack/layer naming convention is that, historically, “middleware” was primarily used in server contexts. Since Tower's services can be used in a client _or_ server context, a new naming convention for a new concept is fitting.

jonhoo: On my way out the door, but my 2c: I think the name should communicate what the property of this "thing" is that differentiates it from just any ol' `Service`. And I _believe_ that property is that it's a 1-1 mapping. I personally believe "layer" doesn't carry that implication, but will defer to consensus. I like words like "wrap", "chain", and "link" for this.

tower-middleware/src/lib.rs
olix0r: Yeah, it's probably  not worth pursuing...

jonhoo:
Let's see if I can make a more coherent naming argument. For the entire discussion below, substitute "zebra" for "layer", "middleware", "link", "wrapper", or whatever else floats your boat. I'll also use "giraffe" in place of "stack", "chain", "onion", and other terms that signify a sequence of Zebras.

To me, there are two different consumers of Tower: the _users_, and the _implementors_. Specifically, for this discussion, the former are people who want to add zebras to their application, and the latter are people who are creating or maintaining zebras.

For users, you want docs and names that are relatively high-level, and speaks to how to "think about" your application. I think using words like "layer" or "middleware" fits well there. As an end user though, I think I'd be confused to ever see a trait beyond `Service` mentioned in this regard, because "aren't they all `Service`s?". As a user, I wouldn't want to have to think about much beyond what `Service` things I want to combine as zebras of my application.

As far as I understand it, the reason we need another type is because a zebra can't _just_ be `Service`; we need to call _something_ to add a zebra to a giraffe. That is, we need something like a `fn wrap(Service) -> Service`. The user probably doesn't care about that, or about that type; to them, it's all just a bunch of `Service`s that happen one after the other. That's the abstraction they care about. Each `Service` is a zebra in their giraffe.

We _do_ need that trait though (unless we have a `Service<Service<Request>>::Response = Self` type thing), and it _does_ need to have a name. Since top-level users won't really deal with it (again, they'll think in terms of a giraffe that has many zebras of `Service`), I think we should rather think about implementors when picking its name. To an implementor, the trait name `Zebra` is unhelpful: they will have to understand the distinction between `Service` and `Zebra`, and `Zebra` doesn't work well as a verb in this context (like traits usually should), nor does it describe what it does or is. So, what is it then?

Well, how does a thing act as a zebra? Well, it can act as a zebra if it is able to wrap a `Service`. Traits represent behaviors, and are usually named after their primary method. And so, to that end, I present:

```rust
trait WrapService<S, Request>
  where S: Service<Request> {
    type Service: Service<Request>;
    fn wrap(self, S) -> Self::Service;
}
```

The exact definition we can bicker over later, but the name is what we're discussing. I would argue that `WrapService` is _immediately_ clear and unambiguous. If you were an implementor and saw a trait called `WrapService`, I think you'd pretty much immediately understand what it did. And that's a major benefit.

So what about users? They'll have a DSL of some kind that lets them add zebras to their giraffes all day long, but ultimately they're going to be greeted with a method signature along the lines of:
```rust
impl Giraffe {
    fn inject_zebra<X>(&mut self, zebra: X)
}
```
What should `X` be here? Well, in theory it is `X: WrapService` following the argument above, but that might confuse users. After all, they've never heard of `WrapService`, nor do I think they _should_. That's an implementation detail that they shouldn't need to be aware of. Instead, I propose we fix this with a type alias:
```rust
type Zebra = WrapService;
```
Which leads to a user-facing type of:
```rust
impl Giraffe {
    fn inject_zebra<Z: Zebra>(&mut self, zebra: Z) {}
}
```
Or, if you want something more concrete:
```rust
impl Stack {
    fn push<M: Middleware>(&mut self, middleware: M) {}
}
```
Whereas implementors would write things like:
```rust
impl WrapService for MyMiddleware { /* ... */ }
```

@LucioFranco raised a question on Gitter about what implementors should call their concrete types, given that each middleware is likely to both have an `impl Service` and an `impl WrapService`. I would probably recommend that for something like `RateLimit`, those types would be called `RateLimitService` and `RateLimit` respectively.

LucioFranco:
@carllerche and whoever this should be good to go. I've implemented all the `Layer` impls I can.

tower-buffer/src/lib.rs
```diff
@@ -40,6 +42,36 @@ where
...
+
+impl BufferLayer {
+    pub fn new(bound: usize) -> Self {
+        BufferLayer { bound }
+    }
+}
+
+impl<S, Request> Layer<S, Request> for BufferLayer
+where
+    Request: Send + 'static,
```
carllerche: Why are there `Send` and `Sync` bounds being added here?

tower-buffer/src/lib.rs
```diff
@@ -40,6 +42,36 @@ where
     worker: worker::Handle<T::Error>,
 }
 
+/// Buffer requests with a bounded buffer
+pub struct BufferLayer {
```
carllerche: I'd recommend leaving `Buffer` out from this PR:

* We can discuss the executor question in a dedicated buffer layer PR.
* There are are changes coming to buffer (see #168).

tower-layer/src/lib.rs
```diff
@@ -0,0 +1,106 @@
...
+#![doc(html_root_url = "https://docs.rs/tower-layer/0.1.0")]
+
+extern crate futures;
+extern crate tower_service;
+
+#[cfg(feature = "util")]
+mod util;
+
+#[cfg(feature = "util")]
+pub use util::{Chain, LayerExt};
```
carllerche: Since we are bikeshedding, I'd probably avoid re-exporting `Chain`. I'd also make `mod util` public (and leave `Chain` in there).

docs.rs/tokio-buf

seanmonstar:
I kinda don't want to add `{ServiceName}Layer` as public types to each crate... It's more a personal nitpick I suppose, but here's my desired alternative:

```rust
// in tower-timeout, for example
pub fn layer<S, Req>(timeout: Duration) -> impl Layer<S, Req>
where
    S: Service<Req>,
{
    tower_layer::layer_fn(move |svc| {
        Timeout::new(svc, timeout)
    })
}
```

carllerche:
@seanmonstar Is there a reason you don't want to add it? Doing that option would mean that the layer cannot be referenced by name (yet).

tower-buffer/src/lib.rs
```diff
@@ -40,6 +43,54 @@ where
     worker: worker::Handle<T::Error>,
 }
 
+/// Buffer requests with a bounded buffer
+pub struct BufferLayer<E, S, Request> {
```
carllerche: Could `E` be the last generic with a default value of `DefaultExecutor`? `struct BufferLayer<Request, S, E = DefaultExecutor`.

tower-buffer/src/lib.rs
```diff
@@ -40,6 +43,54 @@ where
...
+    E: WorkerExecutor<S, Request>,
+{
+    type Response = S::Response;
+    type Error = Error<S::Error>;
+    type Service = Buffer<S, Request>;
+
+    fn layer(&self, service: S) -> Self::Service {
+        match Buffer::with_executor(service, self.bound, &self.executor) {
+            Ok(b) => b,
+            Err(_) => panic!("Unable to spawn task on the provided executor"),
```
carllerche: This... is not good. Layering will happen at runtime.

This means that either `layer` has to return result, or it has to return a DOA service. The first is probably best.

olix0r: +1 on layer returning a result

carllerche: The alternative is to update `Buffer::new` to be infallible but if the spawn fails, then the first call to `poll_ready` returns Err, so a DOA Buffer instance.

carllerche: The problem with a DOA buffer instance is if you *then* wrap that with `Reconnect`, it would result in an infinite loop of `Reconnect` attempting to reconnect...

LucioFranco:
So one thing that has come up with adding a `Result` return type to `layer` has been the extra associated types. We now need to also return a specific error for creating a layer. This error would be `Layer::LayerError`. In most cases there is no error produced from a layer so we can use a `Never` error type for this. In the case of buffer, it would return a `SpawnError<S>`.

Now this also raises a few other issues, we now will have 4 associated types with Layer which is a lot. I think its fine to keep them as they mostly just map 1-1 with the service generic but we need them for this:

```rust
type Service: Service<Request, Response = Self::Response, Error = Self::Error>;
```

I don't think there is any other way around this :/

Here is the final proposed `Layer` trait:

```rust
pub trait Layer<S, Request> {
    /// The wrapped service response type
    type Response;

    /// The wrapped service's error type
    type Error;

    /// The error type produced when calling `layer`
    type LayerError;

    /// The wrapped service
    type Service: Service<Request, Response = Self::Response, Error = Self::Error>;

    /// Wrap the given service with the middleware, returning a new service
    /// that has been decorated with the middleware.
    fn layer(&self, inner: S) -> Result<Self::Service, Self::LayerError>;
}
```

cc @carllerche @seanmonstar @olix0r

carllerche:
The current question is whether or not `Layer` should forward the `Response` and `Error` associated types. The `Service` associated type is the authoritative source for these types. However, traversing associated types in bounds adds visual noise:

**Noisy**:

```rust
where <L::Service as Service<Request>>::Response: BufStream
```

**Better**:

```rust
where L::Response: BufStream
```

There are some downsides to forwarding, primarily the `Response` type is now specified in two places (`Service` and `Layer`). This involves duplication when implementing `Layer`:

```rust
impl<Service, Request> Layer<Service, Request> for MyLayer {
    type Response = <Self::Service as Service<Request>>::Response;
    type Error = <Self::Service as Service<Request>>::Error;
    type Service = ...;
}
```

One way to mitigate this is a macro.

---

As far as I can tell, there is no consensus as to whether `Layer` should forward the associated types. 

Currently 👍 on forwarding the associated types:

* @carllerche 
* @olix0r 

Currently 👎 on forwarding the associated types:

* @jonhoo 
* @LucioFranco 
* @seanmonstar

davidbarsky:
I'm a 👍🏻 on forwarding associated types if the primary forwarding mechanism is a procedural macro. I think that has the best path for forwards compatibility with associated type defaults, as the macro can simply be removed once associated type defaults land in stable.

carllerche:
An example of how an attr macro can remove some boilerplate:

```rust
#[layer_macro_power]
impl<Request, Service> Layer<Request, Service> for MyLayer {
    type LayerError = ();
    type Service = MyService<Request>;

    fn layer(&self, inner: Service) -> Result<Self::Service, Self::LayerError> {
        Ok(MyLayer::new(inner))
    }
}
```

tower-buffer/src/lib.rs
```diff
@@ -40,6 +43,52 @@ where
     worker: worker::Handle,
 }
 
+/// Buffer requests with a bounded buffer
+pub struct BufferLayer<S, Request, E = DefaultExecutor> {
```
carllerche: Why is the struct generic over `S` and `Request`.

tower-buffer/src/lib.rs
```diff
@@ -40,6 +43,52 @@ where
...
+        }
+    }
+}
+
+impl<E, S, Request> BufferLayer<S, Request, E>
+where
+    S: Service<Request>,
+    E: WorkerExecutor<S, Request>,
+{
+    pub fn with_executor(bound: usize, executor: E) -> Self {
```
carllerche: The generics can be moved to the `with_executor` fn.

PR #166, opened 2019-02-25
Cleanup Buffer

carllerche:
This PR is mostly just about breaking up the single `lib.rs` file into a bunch of mods.

This changes the public API of the crate by:
- Moving all error types into an `error` module
- Moving `ResponseFuture` into a `future` module.

Error and future types are secondary, so moving them into their own modules leaves only important types at the root (improves generated docs).

### DIFF
--- a/tower-buffer/src/worker.rs
+++ b/tower-buffer/src/worker.rs
@@ -1,0 +1,182 @@
+use error::ServiceError;
+use futures::future::Executor;
+use futures::{Async, Future, Poll, Stream};
+use std::sync::Arc;
+use tokio_sync::mpsc;
+use tower_service::Service;
+use {Message, State};
+
+/// Task that handles processing the buffer. This type should not be used
+/// directly, instead `Buffer` requires an `Executor` that can accept this task.
+pub struct Worker<T, Request>
+where
+    T: Service<Request>,
+{
+    current_message: Option<Message<Request, T::Future, T::Error>>,
+    rx: mpsc::Receiver<Message<Request, T::Future, T::Error>>,
+    service: T,
+    finish: bool,
+    failed: Option<Arc<ServiceError<T::Error>>>,
+    state: Arc<State<T::Error>>,
+}
+
+/// This trait allows you to use either Tokio's threaded runtime's executor or the `current_thread`
+/// runtime's executor depending on if `T` is `Send` or `!Send`.
+pub trait WorkerExecutor<T, Request>: Executor<Worker<T, Request>>
+where
+    T: Service<Request>,
+{
+}
+
+impl<T, Request, E: Executor<Worker<T, Request>>> WorkerExecutor<T, Request> for E where
+    T: Service<Request>
+{
+}
+
+impl<T, Request> Worker<T, Request>
+where
+    T: Service<Request>,
+{
+    pub(crate) fn spawn<E>(
+        service: T,
+        rx: mpsc::Receiver<Message<Request, T::Future, T::Error>>,
+        state: Arc<State<T::Error>>,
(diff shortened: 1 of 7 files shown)


carllerche:
I also removed the lazy_cell dep (as I have been burned by the lib in the past).

I opted for `Arc<Mutex<Option<_>>>` as an alternative as it is not in the hot path, so I am not super worried.

tower-buffer/src/worker.rs
```diff
@@ -0,0 +1,215 @@
+use error::ServiceError;
+use futures::future::Executor;
+use futures::{Async, Future, Poll, Stream};
+use message::Message;
+use std::sync::{Arc, Mutex};
+use tokio_sync::mpsc;
+use tower_service::Service;
+
+/// Task that handles processing the buffer. This type should not be used
+/// directly, instead `Buffer` requires an `Executor` that can accept this task.
```
jonhoo: Maybe add a note that this is only `pub` because we're sealing over it? Specifically, the type is _not_ actually exported by `tower-buffer`.

tower-buffer/src/worker.rs
```diff
@@ -0,0 +1,215 @@
...
+    T: Service<Request>,
+{
+}
+
+impl<T, Request, E: Executor<Worker<T, Request>>> WorkerExecutor<T, Request> for E where
+    T: Service<Request>
+{
+}
+
+impl<T, Request> Worker<T, Request>
```
jonhoo: I'm just going to assume that nothing really changed in the below code beyond the move away from `lazy_cell`, and not do an in-depth review. If that's true, feel free to just resolve this. If something _did_ change, can you point me at what?

carllerche: Yes, nothing of substance should have changed. I will attempt to annotate the changes I made.

tower-buffer/src/worker.rs
```diff
@@ -0,0 +1,215 @@
...
+where
+    T: Service<Request>,
+{
+    pub(crate) fn spawn<E>(
+        service: T,
+        rx: mpsc::Receiver<Message<Request, T::Future, T::Error>>,
+        executor: &E,
+    ) -> Result<Handle<T::Error>, T>
```
carllerche: This is changed to initialize the handle & return it instead of having a handle passed in.

tower-buffer/src/worker.rs
```diff
@@ -0,0 +1,215 @@
...
+        // Note that we need to handle the case where some handle is concurrently trying to send us
+        // a request. We need to make sure that *either* the send of the request fails *or* it
+        // receives an error on the `oneshot` it constructed. Specifically, we want to avoid the
+        // case where we send errors to all outstanding requests, and *then* the caller sends its
+        // request. We do this by *first* exposing the error, *then* closing the channel used to
+        // send more requests (so the client will see the error when the send fails), and *then*
+        // sending the error to all outstanding requests.
+        let error = Arc::new(ServiceError::new(method, error));
+
+        let mut inner = self.handle.inner.lock().unwrap();
```
carllerche: Starting here is the code that uses the mutex instead of LazyCell.

tower-buffer/src/worker.rs
```diff
@@ -0,0 +1,215 @@
...
+                    // We don't have any new requests to enqueue.
+                    // So we yield.
+                    return Ok(Async::NotReady);
+                }
+            }
+        }
+    }
+}
+
+impl<E> Handle<E> {
```
carllerche: Handle impl is new.

carllerche:
Addressed feedback.

PR #168, opened 2019-02-25
buffer: replace generic error with Box<Error>

carllerche:
Work towards #131 

Refs: #156

### DIFF
--- a/tower-balance/examples/demo.rs
+++ b/tower-balance/examples/demo.rs
@@ -18,6 +18,7 @@ use futures::{future, stream, Async, Future, Poll, Stream};
 use hdrsample::Histogram;
 use rand::Rng;
 use std::collections::VecDeque;
+use std::fmt;
 use std::time::{Duration, Instant};
 use tokio::{runtime, timer};
 use tower_balance as lb;
@@ -104,9 +105,9 @@ where
     D: Discover + Send + 'static,
     D::Key: Send,
     D::Service: Service<Req, Response = Rsp> + Send,
-    D::Error: Send + Sync,
+    D::Error: ::std::error::Error + Send + Sync + 'static,
     <D::Service as Service<Req>>::Future: Send,
-    <D::Service as Service<Req>>::Error: Send + Sync,
+    <D::Service as Service<Req>>::Error: ::std::error::Error + Send + Sync + 'static,
     C: lb::Choose<D::Key, D::Service> + Send + 'static,
 {
     println!("{}", name);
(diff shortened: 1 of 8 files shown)


tower-buffer/tests/buffer.rs
```diff
@@ -83,29 +84,55 @@ fn when_inner_is_not_ready() {
...
     let mut res1 = service.call("hello");
 
     // Allow the Buffer's executor to do work
     ::std::thread::sleep(::std::time::Duration::from_millis(100));
     with_task(|| {
         let e = res1.poll().unwrap_err();
-        if let error::Error::Closed(e) = e {
-            assert!(format!("{:?}", e).contains("poll_ready"));
-            assert_eq!(e.error(), &tower_mock::Error::Other("foobar"));
+        if let Some(e) = e.downcast_ref::<error::ServiceError>() {
```
carllerche: Most of this junk will go away once `tower-mock` is updated to use `Box<Error>`.

tower-mock/src/lib.rs
```diff
@@ -336,6 +337,32 @@ impl<T, E> Future for ResponseFuture<T, E> {
     }
 }
 
+// ===== impl Error =====
+
+impl<T> fmt::Display for Error<T>
```
carllerche: This change is just to get things to work for now. It will be changed once `tower-mock` is updated to use `Box<Error>`.

carllerche:
This breaks a bunch of the other tower crates... I will be working on that soon.

tower-balance/examples/demo.rs
```diff
@@ -233,12 +234,25 @@ impl Discover for Disco {
     }
 }
 
+#[derive(Debug)]
+pub struct DiscoError;
```
carllerche: This is all pretty terrible, but I *think* most of these changes will go away *once* `Box<Error>` is applied everywhere.

tower-balance/examples/demo.rs
```diff
@@ -305,3 +323,31 @@ where
         Ok(Async::Ready(None))
     }
 }
+
+pub struct Errify<T>(T);
```
LucioFranco: is this something we will always need? or is it just for compat between other services that do not yet use `Box<Error>`?

carllerche: It should go away once everything is `Box<Error>`.

tower-buffer/src/error.rs
```diff
@@ -5,9 +5,9 @@ use std::sync::Arc;
 
 /// An error produced by a `Service` wrapped by a `Buffer`
 #[derive(Debug)]
-pub struct ServiceError<E> {
+pub struct ServiceError {
```
olix0r: should this impl clone now that it can?

carllerche: Generally, errors don't implement `Clone`, so i thought i would avoid it unless needed.

PR #175, opened 2019-02-28
Introduce the `ServiceBuilder`

LucioFranco:
This is a very WIP servicebuilder. The idea behind this is to take a `MakeService` and a set of `Layer`s and combine them into a new `MakeService` that will wrap the `Service` produced from the original `MakeService` with the layer. There are still a few unanswered questions, like the order of services is actually opposite (I think) than what you would think. This also works with servers, and I will add that example soonish.


examples/client.rs
```diff
@@ -0,0 +1,55 @@
...
+
+fn request() -> impl Future<Item = Response<hyper::Body>, Error = ()> {
+    let connector = Connector::new(HttpConnector::new(1));
+    let hyper = Connect::new(connector, Builder::new());
+
+    // let policy = RetryPolicy::new(5);
+    let dst = Destination::try_from_uri(Uri::from_static("http://127.0.0.1:3000")).unwrap();
+
+    let maker = ServiceBuilder::new(hyper)
+        // .add(RetryLayer::new(policy))
```
LucioFranco: retry requires clone, there is something happening here but will need to look into it later. For reference, I got this working before we renamed middleware to layer, so im 100% i can get it to work.

examples/client.rs
```diff
@@ -0,0 +1,55 @@
...
+        // .add(BufferLayer::new(5))
+        .add(InFlightLimitLayer::new(5))
+        .add(RateLimitLayer::new(5, Duration::from_secs(1)))
```
seanmonstar: Just checking, `add` means the new layer will wrap everything before it, right? So this ends up being `RateLimit<InFlightLimit<Hyper>>`, yea?

src/builder.rs
```diff
@@ -0,0 +1,139 @@
...
+
+impl<S, L, Target, Request> ServiceBuilder<S, L, Target, Request>
+where
+    S: Service<Target>,
+    S::Response: Service<Request>,
+    L: Layer<S::Response, Request>,
+    Target: Clone,
+{
+    /// Add a layer to the `ServiceBuilder`.
+    pub fn add<T>(self, layer: T) -> ServiceBuilder<S, Chain<L, T>, Target, Request>
```
seanmonstar: Something we noticed in Linkerd2 was that switching the order of our generics from `and_then` to `push` significantly helped with compile times. I don't know why, but it may be worth checking?

For reference: https://github.com/linkerd/linkerd2-proxy/blob/7e5519619d517d28754c4492887e152700c7e916/lib/stack/src/layer.rs#L22-L46

examples/client.rs
LucioFranco: Yes, I think that is correct. I need to fully test that, @carllerche would know more since he originally wrote the chain code for `tower-wb`

src/builder.rs
seanmonstar: Looking at it again, I think it was changing from `a.and_then(b).and_then(c)` to `c.push(b).push(a)`.

LucioFranco: If push made it easier, I would suspect this has to do how it loops through generics for coherence checks.

hawkw: I think that the really long compile times are largely a result of how Linkerd2's `Layer` can modify the target type, which requires it to be generic over both its own target and the target of the next layer in the stack. I notice that the current `tower-layer` doesn't really support this pattern, so it may not be effected by this issue...

examples/client.rs
carllerche: would `.layer` be clearer?

src/builder.rs
```diff
@@ -0,0 +1,177 @@
...
+    S: Service<Target>,
+{
+    inner: S::Future,
+    layer: Arc<L>,
+    _pd: PhantomData<(Target, Request)>,
+}
+
+/// Errors produced from building a service stack
+#[derive(Debug)]
+pub enum Error<M, L> {
```
carllerche: Instead of an enum, we probably want `T: Into<Box<Error + Send + Sync>>`?

src/builder.rs
carllerche: `tower-layer` should support this use case? The `S` generic is not bound at all, so it can be a service w/ a different `Request` generic. Or am I missing something?

src/builder.rs
```diff
@@ -0,0 +1,177 @@
...
+        }
+    }
+}
+
+/// A no-op middleware.
+///
+/// When wrapping a `Service`, the `Identity` layer returns the provided
+/// service without modifying it.
+#[derive(Debug, Default, Clone)]
+pub struct Identity {
```
carllerche: Any chance you could extract this into a dedicated PR? This should be an obvious win.

src/builder.rs
```diff
@@ -0,0 +1,177 @@
...
+
+/// Configure and build a `MakeService`
+///
+/// `ServiceBuilder` collects layers and a `MakeService` transport
+/// and produces a new `MakeService` that is wrapped by the composed
+/// layer.
+#[derive(Debug)]
+pub struct ServiceBuilder<S, L, Target, Request>
+where
+    S: Service<Target>,
```
carllerche: I don't think you need these bounds at the struct level. You should be able to get away without any bounds or phantom data.

src/builder.rs
```diff
@@ -0,0 +1,177 @@
...
+    S::Response: Service<Request>,
+{
+    maker: S,
+    layer: L,
+    _pd: PhantomData<(Target, Request)>,
+}
+
+/// Composed `MakeService` produced from `ServiceBuilder`
+#[derive(Debug)]
+pub struct ServiceBuilderMaker<S, L, Target, Request> {
```
carllerche: I would probably put these secondary types in a mod and shorten the names.

src/builder.rs
```diff
@@ -0,0 +1,177 @@
...
+        }
+    }
+}
+
+impl<S, L, Target, Request> ServiceBuilder<S, L, Target, Request>
+where
+    S: Service<Target>,
+    S::Response: Service<Request>,
+    L: Layer<S::Response, Request>,
+    Target: Clone,
```
carllerche: Why is `Clone` required here?

src/builder.rs
```diff
@@ -0,0 +1,177 @@
...
+            maker: self.maker,
+            layer: Arc::new(self.layer),
+            _pd: PhantomData,
+        }
+    }
+}
+
+impl<S, L, Target, Request> Service<Target> for ServiceBuilderMaker<S, L, Target, Request>
+where
+    S: Service<Target> + Send + 'static,
```
carllerche: Why bound by `Send + 'static`?

carllerche: Also, shouldn't `S` be bound by `MakeService`?

carllerche:
Looks solid. I left some notes inline.

For completeness, some alternative API direction could be to use `chain` only:

```rust
my_layer
    .chain(other_layer)
    .chain(third_layer)
    .build_make_service(my_make_service)
```

or:

```rust
my_make_service.layer({
    my_layer
        .chain(other_layer)
        .chain(third_layer)
})
```

That said, the `ServiceBuilder` approach has nice precedent from other langs.

davidbarsky:
While we're bikeshedding names, the use of `and_then` for composing layers would be inline with how Rust composes things that look like monads, and would be in line [with Finagle's approach](https://monkey.org/~marius/funsrv.pdf), reproduced below:

```scala
recordHandletime     andThen
traceRequest         andThen
collectJvmStats      andThen
parseRequest         andThen
logRequest           andThen
recordClientStats    andThen
sanitize             andThen
respondToHealthCheck andThen
applyTrafficControl  andThen
virtualHostServer
```

carllerche:
If `and_then` is used, we would probably want to switch the order such that you start by specifying the outer most layer... though that feels a bit confusing?

davidbarsky:
Ah, this is the source of my confusion—the layer _stack_ is LIFO, but layers with `and_then` would be a FIFO, right?

My PoV of `Service` is that it's a fancy future, so some sort of sequential layering _might_ be best? I don't feel too strongly about this.

(I'm sorry if this comment derailed any discussion.)

seanmonstar:
@davidbarsky my [previous comment](https://github.com/tower-rs/tower/pull/175#discussion_r261424363) is also relevant, we noticed compile-time improvements switching from FIFO to LIFO...

davidbarsky:
@seanmonstar thanks for the note. @hawkw's comment sticks out to me here:

> I think that the really long compile times are largely a result of how Linkerd2's Layer can modify the target type, which requires it to be generic over both its own target and the target of the next layer in the stack. I notice that the current tower-layer doesn't really support this pattern, so it may not be effected by this issue...

Has that been confirmed?

src/builder.rs
LucioFranco: so I tried this but I thought I needed them because of this
image

LucioFranco:
@seanmonstar @carllerche @hawkw @davidbarsky I've gone ahead and brought this more current with the newer tower changes including the Box<Error> stuff. I've also cleaned up a bunch of the bounds another review would be helpful! I think we are close, I still haven't figured out if we want to rename it from `add` to say something like `push`. There is also a weird sized error type issue with `tower-hyper` that I will have to figure out at another time but the client code works!

tower-buffer/src/error.rs
```diff
@@ -85,11 +85,12 @@ where
     }
 }
 
-impl<T> std::error::Error for SpawnError<T>
-where
-    T: std::error::Error + 'static,
+impl<T: std::fmt::Debug> std::error::Error for SpawnError<T>
+// where
+//     T: std::error::Error + 'static,
```
LucioFranco: Had trouble with this since `Connection` doesn't impl `std::error::Error`

cc @carllerche

tower-layer/src/util/never.rs
```diff
@@ -0,0 +1,14 @@
+use std::error::Error;
+use std::fmt;
+
+#[derive(Debug)]
+/// An error that can never occur.
+pub enum Never {}
```
LucioFranco: I've add this type here as it seems to be pretty useful for layer impls.

tower/examples/server.rs
```diff
@@ -0,0 +1,81 @@
...
+fn main() {
+    hyper::rt::run(future::lazy(|| {
+        let addr = "127.0.0.1:3000".parse().unwrap();
+        let bind = TcpListener::bind(&addr).expect("bind");
+
+        println!("Listening on http://{}", addr);
+
+        let maker = ServiceBuilder::new(MakeSvc)
+            // TODO(lucio): figure out why this is causing sized issues
+            // .add(InFlightLimitLayer::new(5))
```
LucioFranco: Having a sized issue here when I uncomment this line.

image

tower-layer/src/util/chain.rs
```diff
@@ -10,13 +10,7 @@ pub struct Chain<Inner, Outer> {
...
 }
 
-/// Error's produced when chaining two layers together
-pub enum ChainError<I, O> {
-    /// Error produced from the inner layer call
-    Inner(I),
-    /// Error produced from the outer layer call
-    Outer(O),
-}
+type Error = Box<::std::error::Error + Send + Sync>;
```
seanmonstar: Teeniest of nits: I've been liking the clarity of `Box<dyn Trait>` (easier to differentiate from `Box<Struct>` (like if there were also `struct Error` somewhere)).

tower/examples/server.rs
seanmonstar: It's this line here: https://github.com/tower-rs/tower-hyper/blob/master/src/server/mod.rs#L24

It wants the error type to implement `Error`, and `Box<E>` only implements `Error` if it is `Sized` (not a trait object).

tower/src/builder.rs
```diff
@@ -0,0 +1,151 @@
...
+        }
+    }
+}
+
+/// A no-op middleware.
+///
+/// When wrapping a `Service`, the `Identity` layer returns the provided
+/// service without modifying it.
+#[derive(Debug, Default, Clone)]
+pub struct Identity {
```
seanmonstar: This can be replaced with that other PR you made?

LucioFranco: Yup once that one is merged I can remove this and the never I added.

tower/examples/server.rs
LucioFranco: Okay, so I tried to change this to `Into<Box<dyn Error>>` and got the same thing :/

seanmonstar: Oh here too :D https://github.com/tower-rs/tower-hyper/blob/master/src/server/mod.rs#L81

davidbarsky:
Thanks @LucioFranco! Regarding what you saying:

>  ...still haven't figured out if we want to rename it from add to say something like push.

Did @seanmonstar or @hawkw confirm that the switch from FIFO to LIFO was due to:

> I think that the really long compile times are largely a result of how Linkerd2's Layer can modify the target type, which requires it to be generic over both its own target and the target of the next layer in the stack?

That's a pretty important datapoint that'll lead the design in different ways.

tower/examples/server.rs
davidbarsky: Fixed this one!

tower-layer/src/util/chain.rs
davidbarsky: fixed this one.

tower-layer/src/util/never.rs
carllerche: I believe that there is a “void” crate for those that don’t want to implement a Never type.

davidbarsky:
Given an offline discussion with @LucioFranco, we agreed that remaining tasks are

- Remove `Identity` from this PR once #195 is merged.
- Resolve questions around Layer ordering in the builder. Personally, I'd prefer an ordering _opposite_ of what is in this PR, as layer composition is monadic _and_ this layer implementation doesn't modify a target of a stack (thanks for the context, @hawkw and @seanmonstar).
- Resolve whether we should swap this crate's `Never` type for `Void`.

Additionally, I think an example of testing a composed layer stack (perhaps mocking out the transport?) would be a useful bit of documentation.

tower-buffer/src/error.rs
carllerche: I'm not following this. I forgot the context.

carllerche:
@davidbarsky re: `Never`, I'd like to keep it a private detail. We can keep with our private `Never` error type (don't export it). We can tell users that need something to either use `Void` or define their own `Never` type.

tower-buffer/src/error.rs
carllerche: I accidentally pushed to master... but would this help? https://github.com/tower-rs/tower/commit/733b597f38a1acf1c9a25ed4528a96d168bb4cd8

LucioFranco: Yup, this should fix my issue, ill bring this branch current to master later today. Thanks.

tower-layer/src/util/never.rs
LucioFranco: Not too sure if its worth it to bring in another crate for this?

LucioFranco:
@carllerche not too sure I understand the reasoning behind keeping `Never` private, could you expand on that a bit? Im indifferent either way this just seems easier.

seanmonstar:
The `Never` types are tiny and anyone can make their own. By keeping them private, hopefully we can swap to `!` when it stabilizes in 2034.

LucioFranco:
@seanmonstar really good point, didnt think about swaping to `!`. I will update this based on that. Thanks.

LucioFranco:
Ok, so I've gone ahead and added support for `MakeService` and `Service`. I've also moved the passing in of the said service to `build` from `new`. This should allow `ServiceBuilder` to apply to almost any situation where `tower` can be used!

With `MakeService`
```rust
ServiceBuilder::new()
  .add(BufferLayer::new(5))
  .add(RetryLayer::new(policy))
  .add(InFlightLimitLayer::new(5))
  .add(RateLimitLayer::new(5, Duration::from_secs(1)))
  .build_maker(SomeMakeService);
```

With `Service`
```rust
ServiceBuilder::new()
  .add(BufferLayer::new(5))
  .add(RetryLayer::new(policy))
  .add(InFlightLimitLayer::new(5))
  .add(RateLimitLayer::new(5, Duration::from_secs(1)))
  .build_svc(SomeService);
```

Now I believe the last few remaining issues are:

- [ ] The name of the layering fn
- [ ] The order that services should be added via

For the last one, I believe we should now reverse the order aka the FIFO instead of the current LIFO.

carllerche:
@LucioFranco re: order, how does it work in other langs?

Cargo.toml
```diff
@@ -16,3 +16,7 @@ members = [
   "tower-service-util",
   "tower-timeout",
 ]
+
+[patch.'https://github.com/tower-rs/tower']
+tower-service = { path = "tower-service" }
```
carllerche: Why is this needed?

LucioFranco: ah this was a relic from tower-hyper depending on the git version but its fixed now. Removing it.

Cargo.toml
```diff
@@ -17,3 +17,6 @@ members = [
   "tower-service-util",
   "tower-timeout",
 ]
+
+[patch.'https://github.com/tower-rs/tower']
+tower-retry = { path = "tower-retry" }
```
carllerche: I think you mentioned this in gitter, but why is this patch statement needed? In theory, patch statements only apply to the git repo proper and won't apply to crates depending on tower via git?

LucioFranco: We currently use tower-hyper's fixed retry policy that depends on the git version but locally here we are using the path version and thus there is a trait type mismatch, which this patch solves. We just need to release `tower-retry` on crates and that should solve this patch.

tower/src/builder/mod.rs
```diff
@@ -0,0 +1,198 @@
...
+/// ServiceBuilder::new()
+///     .chain(BufferLayer::new(5))
+///     .chain(InFlightLimitLayer::new(5))
+///     .chain(RateLimitLayer::new(5, Duration::from_secs(1)))
+///     .build_service(MyService);
+/// ```
+#[derive(Debug)]
+pub struct ServiceBuilder<L, S, Request> {
+    layer: L,
+    _pd: PhantomData<(S, Request)>,
```
carllerche: Ideally, there would be no phantom data.

tower/src/builder/mod.rs
```diff
@@ -0,0 +1,198 @@
...
+        ServiceBuilder {
+            layer: Identity::new(),
+            _pd: PhantomData,
+        }
+    }
+}
+
+impl<L, S, Request> ServiceBuilder<L, S, Request> {
+    /// Chain a layer `T` to the `ServiceBuilder`.
+    pub fn chain<T>(self, layer: T) -> ServiceBuilder<Chain<T, L>, S, Request> {
```
carllerche: "chaining" a layer is mixing metaphors. I would vote for naming this fn `layer` and rename any other usage of the "chain" metaphor.

tower/src/builder/service.rs
```diff
@@ -0,0 +1,79 @@
...
+use futures::{Async, Future, Poll};
+use std::marker::PhantomData;
+use std::sync::Arc;
+use tower_layer::Layer;
+use tower_service_util::MakeService;
+use Service;
+
+/// Composed `MakeService` produced from `ServiceBuilder`
+#[derive(Debug)]
+pub struct ServiceBuilderMaker<S, L, Request> {
```
carllerche: What if this was called `MakeLayeredService`?

tower/src/builder/service.rs
```diff
@@ -0,0 +1,79 @@
...
+    maker: S,
+    layer: Arc<L>,
+    _pd: PhantomData<Request>,
+}
+
+/// Async resolve the MakeService and wrap it with the layers
+#[derive(Debug)]
+pub struct MakerFuture<S, L, Target, Request>
```
carllerche: We've been calling the future type `[Item]Future`. Maybe this should be `ServiceFuture`?

seanmonstar:
If we supplied the inner `Service`/`MakeService` initially, would that get rid of the need for `build_service()`/`build_maker()`?

carllerche:
@seanmonstar they build different types. `build_service` -> a service that has the layer applied. `build_maker` -> a `MakeService` that applies the layer to each constructed service.

LucioFranco:
@seanmonstar I think we will either still need to have both those fns or two different builder structs.

tower/src/builder/mod.rs
```diff
@@ -0,0 +1,194 @@
...
+impl<L> ServiceBuilder<L> {
+    /// Layer a new layer `T` onto the `ServiceBuilder`.
+    pub fn layer<T>(self, layer: T) -> ServiceBuilder<Chain<T, L>> {
+        ServiceBuilder {
+            layer: Chain::new(layer, self.layer),
+        }
+    }
+
+    /// Create a `ServiceBuilderMaker` from the composed middleware and transport.
+    pub fn build_maker<M, S, Target, Request>(self, maker: M) -> LayeredMakeService<M, L, Request>
```
olix0r: I was told that i fI wanted to bikeshed I should bikeshed now...  Can we improve on the names `build_maker` and `build_service`? Specifically, we don't refer to "makers" anywhere else.

`build_make_service` is a little more verbose but slightly clearer imo.

Or, we could call these different things entirely, e.g. `build()` and `with_service()`?

LucioFranco: Im totally fine with calling it `build_make_service` but I don't quite follow what your other suggestion was? Are you suggesting `build_with_service`? or change it from build to `with_service` to complete the "builder"?

carllerche: I'm fine w/ `build_make_service() -> impl MakeService` as it matches `build_service() -> impl Service`.

I'm not sure what `build()` vs. `with_service()` would be.

olix0r: I was suggesting the latter, though i think `build_with_service` also sounds fine.

Note, though, that the maker variant also "builds a service", though that service is one that happens to build services (MakeService is itself a Service)...  So we may want to find naming that relates more to the abstract/concrete nature of the service rather than whether it is a service or not...

LucioFranco: Though `MakeService` is a service, we use the trait alias to tell people that it isn't just a regular service, so I feel naming it build make service is clear to tell people that there is a difference. I'm not sure how else we could abstractly convey that. 

With the `build_with_service`, would then it also have to be `build_with_make_service` which seems a bit long.

carllerche:
🎆 🎉 

Merge when ready!

PR #178, opened 2019-03-01
filter: Update error style

carllerche:
This updates the error style used by tower-filter.

This is a continuation of #131.

This PR **does not** migrate `tower-filter` to use an error type of `Box<Error>`. Instead, it defines a [new type](https://github.com/tower-rs/tower/blob/tower-filter-err/tower-filter/src/error.rs#L7-L10) that optionally contains a `Box<Error>`.

The filter layer can be used as part of routing. The router would sequentially attempt to dispatch a request to an inner service. If the request is rejected, it attempts the next one. In this case, allocating a `Box<Error>` for each attempt is not great.

This strategy still fits within the greater picture. `tower_filter::error::Error` implements the error trait, which means that all other layers that take `T: Into<Error>` will still work with `Filter`. Also, only the immediate caller of `Filter` should care about rejection errors. In which case, `Filter` will be referenced explicitly.

### DIFF
--- a/tower-filter/src/error.rs
+++ b/tower-filter/src/error.rs
@@ -22,7 +22,7 @@ impl Error {
     /// Create a new `Error` representing a rejected request.
     pub fn rejected<E>(source: Option<E>) -> Error
     where
-        E: Into<Source>
+        E: Into<Source>,
     {
         Error {
             reason: Reason::Rejected,
@@ -33,7 +33,7 @@ impl Error {
     /// Create a new `Error` representing an inner service error.
     pub fn inner<E>(source: E) -> Error
     where
-        E: Into<Source>
+        E: Into<Source>,
     {
         Error {
             reason: Reason::Inner,
@@ -67,8 +67,7 @@ impl error::Error for Error {
     fn source(&self) -> Option<&(dyn error::Error + 'static)> {
         if let Some(ref err) = self.source {
             Some(&**err)
-        }
-        else {
+        } else {
             None
         }
     }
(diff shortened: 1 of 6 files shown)


tower-filter/src/error.rs
```diff
@@ -0,0 +1,54 @@
...
+pub(crate) type Source = Box<dyn error::Error + Send + Sync>;
+
+impl Error {
+    /// Create a new `Error` representing a rejected request.
+    pub fn rejected() -> Error {
+        Error { source: None }
+    }
+
+    /// Create a new `Error` representing an inner service error.
+    pub fn inner<E>(source: E) -> Error
```
seanmonstar: It seems one can create an "inner" error even in a predicate.

carllerche: This is by design. The predicate can error w/o actually rejecting. In this case, it is represented by `filter::Error::inner(failure)`.

carllerche:
@seanmonstar The goal is for the `Filter` service to be able to reject without allocating. If the error type is switched to `Box<Error>` rejecting a request requires an allocation.

tower-filter/src/error.rs
```diff
@@ -0,0 +1,54 @@
...
+            source.fmt(fmt)
+        } else {
+            write!(fmt, "rejected")
+        }
+    }
+}
+
+impl error::Error for Error {
+    fn source(&self) -> Option<&(dyn error::Error + 'static)> {
+        if let Some(ref err) = self.source {
```
seanmonstar: Either this shouldn't return the inner error, or the `Display` shouldn't output the inner error (only). Otherwise, something that iterates the error source chain and outputs sees this:

```
Error: my inner error
Caused by: my inner error
```

carllerche: @seanmonstar probably should be `Display` that is fixed then?

carllerche:
@seanmonstar fixed

PR #181, opened 2019-03-04
Add Extra for MakeConnection

quininer:
This PR adds an `Extra` associated type to `MakeConnection`. This allows `Connector` to pass some extra information, like ALPN.

see https://github.com/tower-rs/tower-hyper/issues/9

### DIFF
--- a/tower-util/src/make_connection.rs
+++ b/tower-util/src/make_connection.rs
@@ -29,7 +29,7 @@ impl<S, Request> self::sealed::Sealed<Request> for S where S: Service<Request> {
 impl<C, Request, Response, Extra> MakeConnection<Request> for C
 where
     C: Service<Request, Response = (Response, Extra)>,
-    Response: AsyncRead + AsyncWrite
+    Response: AsyncRead + AsyncWrite,
 {
     type Response = Response;
     type Extra = Extra;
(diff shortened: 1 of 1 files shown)


tower-util/src/make_connection.rs
```diff
@@ -11,24 +11,28 @@ pub trait MakeConnection<Request> {
     /// The transport provided by this service
     type Response: AsyncRead + AsyncWrite;
 
+    /// The extra data
```
LucioFranco: I would add a little more here just to explain why we need it.

carllerche:
In general, this problem will not have a great solution until Rust gets specialization.

My personal preference is to define a trait that allows exposing HTTP specific data. This is the strategy I opted for [here](https://github.com/carllerche/tower-web/blob/master/src/net.rs#L13-L17). This trait could exist in the [`tower-http`](https://github.com/tower-rs/tower-http) repository.

`tower-http` can define its own `MakeConnection` that is bound by the `Connection` trait.

All access fns will need to be able to implement a default,`TcpStream` needs to implement `Connection` but also in memory channels.

Thoughts @seanmonstar ?

LucioFranco:
So right now we have this trait proposed that should probably live in tower-http

```
pub trait HttpExtra {
    fn version(&self) -> Option<http::Version>;
}
```

That said, @carllerche do you think its better to have a trait alias in tower-http for `MakeConnection` or just have the `Connect::new` fn in tower-hyper have a bound on `C::Extra`?

quininer:
I personally prefer tuple, but I agree the `MakeConnection` bound do seem clearer.

I am closing this PR now.

PR #182, opened 2019-03-05
Create subdir for tower crate

carllerche:
This avoids having a "free" `src` dir in the root when every other crate is nested.


carllerche:
Consistency... in part for CI as well.

PR #185, opened 2019-03-06
Refresh Reconnect

carllerche:
Switched to `Box<Error>`.

### DIFF
--- a/tower-reconnect/src/lib.rs
+++ b/tower-reconnect/src/lib.rs
@@ -4,11 +4,15 @@ extern crate log;
 extern crate tower_service;
 extern crate tower_util;
 
+pub mod future;
+
+use crate::future::ResponseFuture;
+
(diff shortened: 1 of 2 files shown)


tower-reconnect/src/lib.rs
```diff
@@ -4,11 +4,15 @@ extern crate log;
 extern crate tower_service;
 extern crate tower_util;
 
+pub mod future;
+
+use crate::future::ResponseFuture;
```
seanmonstar: I think this is illegal until you turn on 2018 edition.

carllerche: doesn't seem like it!

carllerche: 2018 isn't enabled in the crate, CI is passing. Is there something else to check?

PR #187, opened 2019-03-07
Add CallAll combinator.

jonhoo:
This PR adds the `CallAll` combinator, which essentially provides

```rust
Stream<Item = Request> + Service<Request> => Stream<Item = Response>
```

That is, given a `Stream` of requests and a `Service` that can accept those requests, it produces a `Stream` of the corresponding responses.

To allow CallAll to return the wrapped `Service` when there are no more requests, CallAll is _really_ more like a [`StateStream`](https://docs.rs/futures-state-stream/), so we may want to consider pulling that in for compatibility. For the time being though, this is simply exposed through a `StateStreamItem` enum.

### DIFF
--- a/tower-util/Cargo.toml
+++ b/tower-util/Cargo.toml
@@ -8,3 +8,6 @@ publish = false
 futures = "0.1"
 tower-service = "0.2.0"
 tokio-io = "0.1"
+
+[dev-dependencies]
+tokio-mock-task = "0.1"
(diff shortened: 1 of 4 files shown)


carllerche:
Instead of making the `Stream::Item` an enum, have you considered adding a fn on `CallAll`: `into_inner(self) -> S` that returns the inner service? That would provide a more ergonomic API when you don't care about getting the service back and it would allow you to get the service back if you do care.

PR #189, opened 2019-03-08
rate-limit: Refresh layout

carllerche:
- Switch to `Box<Error>`
- Break up lib.rs into multiple files.
- Use `tokio::clock::now` instead of `Instant::now`.

### DIFF
--- a/tower-rate-limit/src/error.rs
+++ b/tower-rate-limit/src/error.rs
@@ -1,0 +1,18 @@
+use std::error;
+
+pub(crate) type Error = Box<error::Error + Send + Sync>;
+
+pub(crate) mod never {
+    use std::{error, fmt};
+
+    #[derive(Debug)]
+    pub enum Never {}
+
+    impl fmt::Display for Never {
+        fn fmt(&self, _: &mut fmt::Formatter) -> fmt::Result {
+            unreachable!();
+        }
+    }
+
+    impl error::Error for Never {}
+}
--- a/tower-rate-limit/src/rate.rs
+++ b/tower-rate-limit/src/rate.rs
@@ -1,0 +1,29 @@
+use std::time::Duration;
+
+#[derive(Debug, Copy, Clone)]
+pub struct Rate {
+    num: u64,
+    per: Duration,
+}
+
+impl Rate {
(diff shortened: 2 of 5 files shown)


tower-rate-limit/src/error.rs
```diff
@@ -0,0 +1,18 @@
+use std::error;
+
+pub(crate) type Error = Box<error::Error + Send + Sync>;
```
LucioFranco: I dont feel like this is the style we used in other crates is it?

tower-rate-limit/src/rate.rs
```diff
@@ -0,0 +1,29 @@
...
+    ///
+    /// This function panics if `num` or `per` is 0.
+    pub fn new(num: u64, per: Duration) -> Self {
+        assert!(num > 0);
+        assert!(per > Duration::from_millis(0));
+
+        Rate { num, per }
+    }
+
+    pub(crate) fn num(&self) -> u64 {
```
LucioFranco: do we want to remove the `pub(crate)` and have this be `pub`? I could see the use case where you might create a rate and want to access what its set too.

tower-rate-limit/src/error.rs
seanmonstar: How so? Looks similar to me...

carllerche: What style?

LucioFranco: the `pub(crate)` but it seems inflight its private. its a non issue was just curious if we maybe wanted to allow these error trait aliases to be avail to users of the crate.

tower-rate-limit/src/rate.rs
carllerche: I *think* if we enable the `unreachable_pub` lint, it will complain if it is not `pub(crate)`.

tower-rate-limit/src/error.rs
carllerche: We've been keeping the `Box<Error>` alias private. Rustdoc will expand it anyway.

LucioFranco: works for me 👍

PR #190, opened 2019-03-09
buffer: fix panics if worker executor drops unexpectedly

seanmonstar:


jonhoo:
This seems to change a bunch of things around, partially related to removing `DirectService` (so we no longer need `method`). What was the actual bug?

seanmonstar:
The tests triggered the panics. They happened if the worker was dropped early without an error. I was seeing it when the Linkerd2 proxy was shutting down, as the runtime dropped.

jonhoo:
I guess what I meant was that it's hard to tease out what exactly the underlying problem was given all the changes. What triggered the panic?

seanmonstar:
The unreachable in future, and the expect in poll_ready.

PR #191, opened 2019-03-11
Reorganize tower-timeout

carllerche:
- Change error bound: `Error: From<S::Error>` (for `?` operator)
- Split types into dedicated files
- Move `ResponseFuture` into public `future` mod.

### DIFF
--- a/tower-timeout/src/error.rs
+++ b/tower-timeout/src/error.rs
@@ -1,0 +1,17 @@
+//! Error types
+
+use std::{error, fmt};
+
+pub(crate) type Error = Box<error::Error + Send + Sync>;
+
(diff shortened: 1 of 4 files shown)


seanmonstar:
FWIW, `From<E> for Box<dyn Error>` is going to be a smaller set of types than `Into<Box<dyn Error>>` (realistically, probably only excluding `&'static str`/`String`, unless other people are implementing `Into<Box<..>>` for custom types that don't otherwise implement `Error`. 🤷‍♂️

carllerche:
@seanmonstar Oh? I was mostly changing to let the `?` work. Why is there no `From<String>` impl?

PR #192, opened 2019-03-12
Add hedge retry middleware

adleong:
Add tower-hedge, a middleware that pre-emptively retries requests which have been outstanding for longer than a given latency percentile.  If either of the original future or the retry future completes, that value is used.  For more information about hedge requests, see: [The Tail at Scale](https://cseweb.ucsd.edu/~gmporter/classes/fa17/cse124/post/schedule/p74-dean.pdf)

Signed-off-by: Alex Leong <alex@buoyant.io>

### DIFF
--- a/tower-hedge/src/lib.rs
+++ b/tower-hedge/src/lib.rs
@@ -1,0 +1,246 @@
+#[macro_use]
+extern crate futures;
+extern crate hdrhistogram;
+#[macro_use]
+extern crate log;
+extern crate tokio_timer;
+extern crate tower_service;
+
+use futures::{Async, Future, Poll};
+use hdrhistogram::Histogram;
+use tokio_timer::{clock, Delay};
+use tower_service::Service;
+
+use std::sync::{Arc, Mutex};
+use std::time::{Duration, Instant};
+
+mod rotating;
+
+use rotating::Rotating;
+
+/// A "retry policy" to classify if a request should be pre-emptively retried.
+pub trait Policy<Request>: Sized {
+    /// Check the policy if a certain request should be pre-emptively retried.
+    ///
+    /// This method is passed a reference to the original request.
+    fn can_retry(&self, req: &Request) -> bool;
+    /// Tries to clone a request before being passed to the inner service.
+    ///
+    /// If the request cannot be cloned, return `None`.
+    fn clone_request(&self, req: &Request) -> Option<Request>;
+}
+
+/// A middleware that pre-emptively retries requests which have been outstanding
+/// for longer than a given latency percentile.  If either of the original
+/// future or the retry future completes, that value is used.
+#[derive(Clone)]
+pub struct Hedge<P, S> {
+    policy: P,
+    service: S,
+    latency_percentile: f32,
+    /// Only retry if there are at least this many data points in the histogram.
+    min_data_points: u64,
+    /// A rotating histogram is used to track response latency.
+    pub latency_histogram: Arc<Mutex<Rotating<Histogram<u64>>>>,
+}
+
+/// The Future returned by the Hedge service.
+pub struct ResponseFuture<P, S, Request>
+where
+    P: Policy<Request>,
+    S: Service<Request>,
+{
+    /// If the request was clonable, a clone is stored.
+    request: Option<Request>,
+    /// The time of the original call to the inner service.  Used to calculate
+    /// response latency.
+    start: Instant,
+    hedge: Hedge<P, S>,
+    orig_fut: S::Future,
+    hedge_fut: Option<S::Future>,
+    /// A future representing when to start the hedge request.
+    delay: Option<Delay>,
+}
+
+// ===== impl Hedge =====
+
+impl<P, S> Hedge<P, S> {
+    pub fn new<Request>(
+        policy: P,
+        service: S,
+        latency_percentile: f32,
+        rotation_period: Duration,
+        min_data_points: u64,
+    ) -> Self
+    where
+        P: Policy<Request> + Clone,
+        S: Service<Request>,
+    {
+        let new: fn() -> Histogram<u64> = || {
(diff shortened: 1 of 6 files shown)


tower-hedge/src/lib.rs
```diff
@@ -0,0 +1,257 @@
...
+/// A "retry policy" to classify if a request should be pre-emptively retried.
+pub trait Policy<Request>: Sized {
+    /// Check the policy if a certain request should be pre-emptively retried.
+    ///
+    /// This method is passed a reference to the original request.
+    fn can_retry(&self, req: &Request) -> bool;
+    /// Tries to clone a request before being passed to the inner service.
+    ///
+    /// If the request cannot be cloned, return `None`.
+    fn clone_request(&self, req: &Request) -> Option<Request>;
```
seanmonstar: It looks like `clone_request` is called at the beginning, and then `can_retry` is called after the delay triggers. Do you envision use cases where the knowledge of whether a certain request should be retried would only be known later? Or would that knowledge be generally known when `clone_request` is being called?

tower-hedge/src/lib.rs
```diff
@@ -0,0 +1,257 @@
...
+/// future or the retry future completes, that value is used.
+#[derive(Clone)]
+pub struct Hedge<P, S> {
+    policy: P,
+    service: S,
+    latency_percentile: f32,
+    /// Only retry if there are at least this many data points in the histogram.
+    min_data_points: u64,
+    /// A rotating histogram is used to track response latency.
+    pub latency_histogram: Arc<Mutex<Rotating<Histogram<u64>>>>,
```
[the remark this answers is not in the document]
adleong: I did this for testing purposes so that a test can pre-populate the histogram.  I'm open to other solutions...

tower-hedge/src/lib.rs
```diff
@@ -0,0 +1,257 @@
...
+                return p;
+            }
+            // Original future is pending, but hedge hasn't started.  Check
+            // the delay.
+            let delay = match self.delay.as_mut() {
+                Some(d) => d,
+                // No delay, can't retry.
+                None => return Ok(Async::NotReady),
+            };
+            match delay.poll() {
```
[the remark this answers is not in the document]
adleong: Good catch.  I'll re-structure this so that the delay is `take()`n and only put back if the Delay is NotReady.

tower-hedge/src/lib.rs
```diff
@@ -0,0 +1,257 @@
...
+                Ok(Async::Ready(_)) => {
+                    try_ready!(self.hedge.poll_ready());
+                    if let Some(req) = self.request.take() {
+                        if self.hedge.policy.can_retry(&req) {
+                            // Start the hedge request.
+                            trace!(
+                                "Issuing hedge request after {:?}",
+                                clock::now() - self.start
+                            );
+                            self.request = self.hedge.policy.clone_request(&req);
```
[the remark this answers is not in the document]
adleong: :+1:

tower-hedge/src/lib.rs
adleong: For example, retry budgets should be checked after the delay has triggered, when the hedge request is about to be issued.

tower-hedge/src/lib.rs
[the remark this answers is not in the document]
seanmonstar: My only concern is that this exposes the inner type. If that dependency has a breaking change, we must also issue one to upgrade.

Looking at the finagle `BackupRequestFilter` tests, it looks like they "warm up" the service before running a test?

tower-hedge/src/rotating.rs
```diff
@@ -0,0 +1,76 @@
...
+use std::time::{Duration, Instant};
+
+use tokio_timer::clock;
+
+/// This represents a "rotating" value which stores two T values, one which
+/// should be read and one which should be written to.  Every period, the
+/// read T is discarded and replaced by the write T.  The idea here is that
+/// the read T should always contain a full period (the previous period) of
+/// write operations.
+pub struct Rotating<T> {
```
carllerche: It seems like more effort than it is worth to make this generic over `T` as the only thing it is every used for is `Histogram<u64>`?

tower-hedge/src/lib.rs
```diff
@@ -0,0 +1,258 @@
...
+    ///
+    /// If the request cannot be cloned, return `None`.
+    fn clone_request(&self, req: &Request) -> Option<Request>;
+}
+
+/// A middleware that pre-emptively retries requests which have been outstanding
+/// for longer than a given latency percentile.  If either of the original
+/// future or the retry future completes, that value is used.
+#[derive(Clone)]
+pub struct Hedge<P, S> {
```
carllerche: It would be cool if `Hedge` could be a layer that always fires off two requests and the `Policy` describes how to build the second request.

In the specific case that you are handling here, could the histogram / latency tracking be moved into the policy? Can the policy trait be:

```rust
pub trait Policy<Request> {
    type Response;
    type Error;
    type Future;

    fn hedge(&mut self, req: &Request) -> Self::Future;
}
```

This looks *very* much like `Service`, but it takes a `&Request` instead. I wonder if there could be a way to somehow "just" use `Service` as a policy, but that isn't initially important.

The way you may be able to implement hedge w/ the histogram would be to return a response future that waits for the delay before sending the request. It could be that there should be a `tower-delay` layer that does this. We can punt though.

Does structuring it like this seem possible (or a good idea)?

adleong: Yes, I think this sounds possible.  However, you'd still need some middleware above or below to measure response latency and feed that data to the policy.  Is it better to assemble this out of small components (latency histogram layer, delay, percentile based policy, and the hedge itself) or have an all-in-one solution like I have here?

adleong:
Replaced by https://github.com/tower-rs/tower/pull/236

PR #193, opened 2019-03-13
Initial introduction of tower-service-util

carllerche:
Currently, all "utility" extensions to `tower_service::Service` exist in `tower-util`. The problem is that layer crates want to depend on these utilities [1], but `tower-util` also wants to depend on the layer crates [2]. This creates a circular dependency.

The solution is to move all utilities that depend only on `tower-service` and crates external to Tower into a crate that all Tower sub crates can depend on. There are two options:

1) Move the utilities into `tower-service` and expose them via a feature flag (the [tokio-buf](https://github.com/tokio-rs/tokio/blob/92d51202ef5e6b99004d77300de1aba0f8835513/tokio-buf/src/lib.rs) strategy).
2) Move them into a new crate: `tower-service-util`.

Given how fundamental and important stability is for the `tower-service` crate, this PR takes the second approach.

This work is in preparation of removing `tower-util` in favor of the `tower` crate.

This PR is not yet complete.

[1]: `MakeService` and `MakeConnection` are used heavily in the layer crates.
[2]: `ServiceExt` wants to gain layer specific fns, like `buffer(n: usize) -> Buffer<Self>`.


seanmonstar:
A possible alternative is to have the `tower` crate be the one that adds the useful extension methods using layers?

carllerche:
@seanmonstar I'm not following your proposal and how it solves the cyclical dependency problem.

carllerche:
I agree w/ @seanmonstar's naming strategy, but the rename will be punted to later:

* `tower-service-util` -> `tower-util`
* `tower-util` -> `tower`.

PR #195, opened 2019-03-15
layer: Add an identity layer

LucioFranco:
Tied to #175

### DIFF
--- a/tower-layer/src/util/identity.rs
+++ b/tower-layer/src/util/identity.rs
@@ -1,0 +1,47 @@
+use std::error::Error;
+use std::fmt;
+use tower_service::Service;
+use Layer;
+
+/// A no-op middleware.
+///
+/// When wrapping a `Service`, the `Identity` layer returns the provided
+/// service without modifying it.
+#[derive(Debug, Default, Clone)]
+pub struct Identity {
+    _p: (),
+}
+
+impl Identity {
+    /// Create a new `Identity` value
+    pub fn new() -> Identity {
+        Identity { _p: () }
+    }
+}
+
+/// Decorates a `Service`, transforming either the request or the response.
+impl<S, Request> Layer<S, Request> for Identity
+where
+    S: Service<Request>,
+{
+    type Response = S::Response;
+    type Error = S::Error;
+    type LayerError = Never;
+    type Service = S;
+
+    fn layer(&self, inner: S) -> Result<Self::Service, Self::LayerError> {
+        Ok(inner)
+    }
+}
+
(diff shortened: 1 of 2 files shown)


tower-layer/src/util/mod.rs
```diff
@@ -23,3 +23,34 @@ pub trait LayerExt<S, Request>: Layer<S, Request> {
...
+}
+
+/// Decorates a `Service`, transforming either the request or the response.
+impl<S, Request> Layer<S, Request> for Identity
+where
+    S: Service<Request>,
+{
+    type Response = S::Response;
+    type Error = S::Error;
+    type LayerError = ();
```
seanmonstar: Should this be some never type?

LucioFranco:
@seanmonstar yeah I can add that, the goal of this was a suggestion from carl to extract easy to merge parts from the ServiceBuilder PR.

carllerche:
@seanmonstar it's pretty useful as a seed to the service builder or other places where you want a "default" layer.

tower-layer/src/util/mod.rs
```diff
@@ -47,10 +50,22 @@ where
...
     type Service = S;
 
     fn layer(&self, inner: S) -> Result<Self::Service, Self::LayerError> {
         Ok(inner)
     }
 }
+
+#[derive(Debug)]
+/// An error that can never occur.
+pub enum Never {}
```
carllerche: This probably should be a hidden type: public in a private mod.

LucioFranco: I actually had this here so that other crates can use it, is that a bad idea?

LucioFranco: because like 50% of layers dont produce an error so this would be useful for them to not have to always define a never type.

carllerche: What do you think @seanmonstar?

LucioFranco:
@carllerche so I think to move this forward we should just use the `void` crate?

tower-layer/src/util/identity.rs
```diff
@@ -0,0 +1,47 @@
...
+    type Service = S;
+
+    fn layer(&self, inner: S) -> Result<Self::Service, Self::LayerError> {
+        Ok(inner)
+    }
+}
+
+/// An error that can never occur.
+#[derive(Debug)]
+pub enum Never {}
```
carllerche: `Never` should be private.

carllerche: Ah, I missed that it was in a private mod.

PR #198, opened 2019-03-18
Polish call all

carllerche:
Posting this early for feedback. The work is done, but I need to add some tests.

This moves `CallAll` into `tower-service-util` (which will become `tower-util` at some point). It also adds an unordered variant where responses are returned in the order in which they are completed. The unordered variant is created via a combinator on the ordered one:

```rust
my_service.call_all(my_requests).unordered()
```


tower-service-util/src/call_all/common.rs
```diff
@@ -0,0 +1,99 @@
...
+    S: Stream,
+    S::Error: Into<Error>,
+    Q: Queue<Svc::Future>,
+{
+    type Item = Svc::Response;
+    type Error = Error;
+
+    fn poll(&mut self) -> Poll<Option<Self::Item>, Self::Error> {
+        loop {
+            let res = self.queue.poll().map_err(Into::into);
```
jonhoo: I don't think you need the `map_err` here. `?` should automatically invoke `into`. And at that point, you can just put `self.queue.poll()?` directly in the `if let` below like in the original implementation.

tower-service-util/src/call_all/common.rs
```diff
@@ -0,0 +1,99 @@
...
+            if self.eof {
+                if self.queue.is_empty() {
+                    return Ok(Async::Ready(None));
+                } else {
+                    return Ok(Async::NotReady);
+                }
+            }
+
+            // Then, see that the service is ready for another request
+            try_ready!(self.service.poll_ready().map_err(Into::into));
```
jonhoo: Same here -- I don't think you need the `map_err`. I think `try_ready!` also automatically applies `into`.

tower-service-util/src/call_all/common.rs
```diff
@@ -0,0 +1,99 @@
...
+
+            // If it is, gather the next request (if there is one)
+            match self.stream.poll().map_err(Into::into)? {
```
jonhoo: And again re `into`.

tower-service-util/src/call_all/common.rs
carllerche: Unfortunately, you do :( Because `?` only works with `Box<Error>: From<E>` and not with `Into<Box<Error>>`

tower-service-util/src/call_all/common.rs
```diff
@@ -0,0 +1,99 @@
...
+/// TODO: Dox
+#[derive(Debug)]
+pub(crate) struct CallAll<Svc, S, Q> {
+    service: Svc,
+    stream: S,
+    queue: Q,
+    eof: bool,
+}
+
+pub(crate) trait Queue<T: Future> {
```
jonhoo: `Queue` is a little weird of a name for this, since `FutureUnordered` isn't a `Queue` (i.e., FIFO).

jonhoo:
LGTM overall!

tower-service-util/src/call_all/common.rs
carllerche: I renamed -> `Drive` 🤷‍♂️

tower-service-util/src/call_all/common.rs
jonhoo: Can't we just invert the bounds to match that instead then?
```rust
impl<Svc, S, Q> Stream for CallAll<Svc, S, Q>
where
    Error: From<Svc::Error>,
    Error: From<S::Error>,
    ...
```

jonhoo:
@davidbarsky Always add docs when adding a feature is an adage we should adopt in tower I think. Will save us from much headache down the line.

tower-service-util/src/call_all/common.rs
carllerche: Then you can't use `String` / `&'static str` as errors.

jonhoo: Ah, I guess this is pretty much https://github.com/rust-lang/rust/pull/58974 /cc @seanmonstar. I suppose we'll leave it this way for the time being.

PR #199, opened 2019-03-18
First Draft of Blocking Service

davidbarsky:
This is a draft of a blocking service that accepts a regular, non-blocking Tower Service and blocks on it until it resolves as a `Result<T, E>`. The idea is to provide [Reqwest](https://github.com/seanmonstar/reqwest)-like functionality while allowing end-programmers to make use of Tower's middleware in a synchronous context. 

**Future Work**

- Once #175 is merged, `BlockingService` can be modified to accept a `ServiceBuilder`. 
- At the moment, I'm not sure `BlockingService` should have a corresponding `Layer` implementation because I think `BlockingService` would be modifying the target type of the composed layers.
- More tests & examples.


tower-blocking-service/src/lib.rs
```diff
@@ -0,0 +1,47 @@
...
+
+trait BlockingService<Request> {
+    type Response;
+    type Error;
+
+    fn call(&mut self, req: Request) -> Result<Self::Response, Self::Error>;
+}
+
+struct BlockingSvc<T, U, E> {
+    inner: BoxService<T, U, E>,
```
davidbarsky: I think this needs to be Boxed because the result of a `ServiceBuilder` isn't nameable at the moment.

tower-blocking-service/src/lib.rs
```diff
@@ -0,0 +1,47 @@
...
+impl<Request, U, E> BlockingService<Request> for BlockingSvc<Request, U, E>
+where
+    E: Send + 'static,
+    U: Send + 'static,
+{
+    type Response = U;
+    type Error = E;
+
+    fn call(&mut self, req: Request) -> Result<Self::Response, Self::Error> {
+        RUNTIME.lock().unwrap().block_on(self.inner.call(req))
```
davidbarsky: I think I'll need to `poll_ready` on the inner service before calling `Service::call`, but I'm not sure if should return an error if the service isn't ready, or busy loop until it is. I'm leaning towards the former.

tower-blocking-service/src/lib.rs
```diff
@@ -0,0 +1,47 @@
...
+use tower_service::Service;
+use tower_service_util::BoxService;
+
+lazy_static! {
+    static ref RUNTIME: Arc<Mutex<Runtime>> = Arc::new(Mutex::new(
+        Runtime::new().expect("Runtime could not be created")
+    ));
+}
+
+trait BlockingService<Request> {
```
davidbarsky: This trait is probably removable.

carllerche:
Could the `lazy_static` be avoided in favor of `blocking`?

davidbarsky:
@carllerche You're referring to [`tokio_threadpool::blocking`](https://docs.rs/tokio-threadpool/0.1.12/tokio_threadpool/fn.blocking.html), right? My impression is that `tokio_threadpool::blocking` + [`futures::future::poll_fn`](https://docs.rs/futures/0.1.25/futures/future/fn.poll_fn.html) are designed for embedded blocking code within a larger non-blocking application (due to the `Poll` return type), while `Runtime::block_on` is meant for transforming non-blocking code into blocking code. I view `BlockingService` as something that tackles the _latter_ use-case.

tower-blocking-service/src/lib.rs
davidbarsky: Removed it.

tower-blocking-service/src/lib.rs
davidbarsky: Added a poll_ready.

tower-blocking-service/Cargo.toml
```diff
@@ -0,0 +1,11 @@
+[package]
+name = "tower-blocking-service"
+version = "0.1.0"
+authors = ["David Barsky <dbarsky@amazon.com>"]
```
LucioFranco: probably want to make this tower contributors like the other crates.

Cargo.toml
```diff
@@ -3,6 +3,7 @@
 members = [
   "tower",
   "tower-balance",
+  "tower-blocking-service",
```
LucioFranco: I think this could just live in tower itself

tower-blocking-service/src/lib.rs
```diff
@@ -0,0 +1,26 @@
...
+        S::Future: Send + 'static,
+    {
+        Self {
+            inner: BoxService::new(inner),
+            rt: Runtime::new().unwrap(),
+        }
+    }
+
+    pub fn call(&mut self, req: Request) -> Result<Response, Err> {
+        self.inner.poll_ready()?;
```
carllerche: This would need to be blocked on as well.

tower-blocking-service/src/lib.rs
```diff
@@ -0,0 +1,26 @@
+use tokio::runtime::current_thread::Runtime;
+use tower_service::Service;
+use tower_service_util::BoxService;
+
+pub struct BlockingSvc<Request, Response, Err> {
+    inner: BoxService<Request, Response, Err>,
```
carllerche: Why does this need to be `BoxService`? I don't think you need anything here besides:

```rust
pub struct Something<T> {
    service: T,
    rt: Runtime,
}
```

tower-blocking-service/src/lib.rs
```diff
@@ -0,0 +1,26 @@
+use tokio::runtime::current_thread::Runtime;
+use tower_service::Service;
+use tower_service_util::BoxService;
+
+pub struct BlockingSvc<Request, Response, Err> {
```
carllerche: The name `BlockingService` made me think that it was a wrapper that took a "sync service" and made it async via a threadpool strategy.

I'm not sure what name we could use for the other way around though.

Cargo.toml
carllerche: I agree.

davidbarsky: Moved it into tower.

tower-blocking-service/src/lib.rs
davidbarsky: Done.

tower-blocking-service/src/lib.rs
davidbarsky: fixed.

tower-blocking-service/src/lib.rs
davidbarsky: I was considering `SyncService` but that might conflict weirdly with `UnsyncBoxService`.

davidbarsky:
Here are the remaining tasks on this:

- [ ] Land https://github.com/tower-rs/tower/pull/175, which will allow us to pass a composed Tower service stack into `BlockingService`.
- [ ] Land https://github.com/tokio-rs/tokio/pull/993/, which would allow us to be generic over the runtime (multi-threaded vs. single-threaded).
- [ ] Documentation & Examples.
- [ ] Tests.

LucioFranco:
This is some good initial work but I am gonna start preparing the master branch for std::future::Future so I am gonna close this. Feel free to reopen against the proper branch.

davidbarsky:
Thanks for the heads up. With async/await on stable, I feel that this PR is less necessary.

PR #204, opened 2019-03-20
Introduce tower-load-shed

seanmonstar:
Provides a middleware that immediately fails any request if the
underlying service isn't ready yet. This is useful when used in
conjunction with `InFlightLimit` or others, so that requests are
rejected immediately, instead of keeping track of extra pending
requests.


seanmonstar:
I wasn't certain about the best name here. I'm open to changing to something else. Names I considered:

- tower-load
- tower-overload
- tower-shed
- tower-unavailable

carllerche:
@seanmonstar re: naming, I don't think `tower-overload` is right. My initial thought was that the crate did stress testing somehow. I would go with:

- tower-load-shed (preferred)
- tower-shed

tower-overload/src/lib.rs
```diff
@@ -0,0 +1,47 @@
...
+    S::Error: Into<Error>,
+{
+    type Response = S::Response;
+    type Error = Error;
+    type Future = ResponseFuture<S, Req>;
+
+    fn poll_ready(&mut self) -> Poll<(), Self::Error> {
+        // We always report ready, and check inner `poll_ready` in the response
+        // future instead.
+        Ok(().into())
```
carllerche: I expected this implementation to call `poll_ready` on the inner service, and track if `Ready` was returned or not. It would always return `Ready` itself, but track the inner ret val for when `call` is invoked. If the inner service returned `Ready`, the request would be forwarded. If not, the request errors (overloaded).

LucioFranco:
I'm +1 on `tower-load-shed`

tower-overload/Cargo.toml
```diff
@@ -0,0 +1,14 @@
+[package]
+name = "tower-overload"
+version = "0.1.0"
+authors = ["Sean McArthur <sean@seanmonstar.com>"]
```
LucioFranco: should this be renamed to tower contributors?

seanmonstar: Heh, that's just a result of `cargo new`. I clicked around, but couldn't find another place where we say "tower contributors". What email is used?

seanmonstar:
Updated to tower-load-shed, and no longer needs `Service + Clone`, but instead tracks readiness via `poll_ready` but always reporting `Ready`.

tower-overload/Cargo.toml
carllerche: At this point, I'm pretty lax about it as we haven't shipped. Before release, I was planning on setting up an email forward. I have `tower-rs.com`, so you can use `team@tower-rs.com` if you want (i have not setup the email forward yet though).

tower-load-shed/src/lib.rs
```diff
@@ -0,0 +1,80 @@
...
+
+extern crate futures;
+extern crate tower_layer;
+extern crate tower_service;
+
+use futures::Poll;
+use tower_service::Service;
+
+pub mod error;
+mod future;
```
carllerche: This probably should be pub... it is pub everywhere else. Do we want to explicitly make response futures not nameable?

seanmonstar: I don't care too much either way. I don't think people should name this one, because it will be annoying if you change around the layers, but it's easy to mark public.

carllerche: I would either like to make this public or make all the other ones private. Which do you think is best?

seanmonstar: I made the re-export public.

PR #205, opened 2019-03-23
buffer: switch to TypedExecutor

carllerche:
Switch tower-buffer to use `TypedExecutor`.

For better or for worse (which is TBD), `TypedExecutor` takes `&mut self`. This makes things a bit trickier in the layer impl, because `layer` takes `&self`.

To work around this, `Layer` now requires the executor to also be `Clone`.

Perhaps having `TypedExecutor` take `&mut self` is not the right choice. However, its been pushed to crates.io and `tokio_executor::Executor`, which has been around for a while also takes `&mut self`.

### DIFF
--- a/tower-buffer/Cargo.toml
+++ b/tower-buffer/Cargo.toml
@@ -5,11 +5,11 @@ authors = ["Carl Lerche <me@carllerche.com>"]
 publish = false
 
 [dependencies]
-futures = "0.1"
+futures = "0.1.25"
 tower-service = "0.2.0"
 tower-layer = { version = "0.1", path = "../tower-layer" }
-tokio-executor = "0.1"
-tokio-sync = "0.1"
+tokio-executor = "0.1.7"
+tokio-sync = "0.1.0"
 
 [dev-dependencies]
 tower-mock = { version = "0.1", path = "../tower-mock" }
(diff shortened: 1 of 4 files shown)


seanmonstar:
Why switch? What is gained? It seems a downside is that less things implement `TypedExecutor`.

carllerche:
Imo the things that matter will implement it. The trait is gone in futures 0.3 so there is no long term option to stick with it. 

TypedExecutor resolves the error type issue. Also whether or not it should take &mut is a real long term question that applies to Executor as well. 

Given that tower-buffer has not shipped yet, it seems good to aim for a long term solution.

seanmonstar:
There's `Spawn` in futures 0.3, which is definitely a less useful version than 0.1's `Executor`, but it's not gone completely. I don't see an issue with the error type nor `&mut`, so it seems like overall an annoyance when someone (like us) tries to plug in an executor and is reminded we didn't implement all 3 traits.

carllerche:
The Spawn trait is the same as tokio_executor::Executor. As of now, there
is no alternative to TypedExecutor in 0.3.

I think there is value to push forward to validate the TypedExecutor trait
in order to learn if any changes will be required come the next breaking
change.

carllerche:
Not to mention, `TypedExecutor` already shipped in tokio-executor. So, there are **are** 3 traits. Tokio will be deprecating the `futures::Executor` implementation, so there is no reason for tower to ship using a trait that is being deprecated by Tokio.

PR #210, opened 2019-03-26
Export all crates from the tower crate

LucioFranco:


tower/src/lib.rs
```diff
@@ -1,14 +1,38 @@
...
+extern crate tower_discover;
+#[cfg(feature = "make_service")]
+extern crate tower_reconnect;
+
+pub mod builder;
+#[cfg(feature = "layers")]
+pub mod layer;
+#[cfg(feature = "make_service")]
+pub mod make_service;
+pub mod service_util;
```
carllerche: I don't think we need this named `service_util`. The types that matter should be exported in `tower::util`. service-util will be renamed tower-util.

PR #211, opened 2019-03-27
balance: cleanup example

carllerche:
This simplifies the balance example using functionality that has recently landed.

### DIFF
--- a/tower-balance/examples/demo.rs
+++ b/tower-balance/examples/demo.rs
@@ -19,12 +19,12 @@ use hdrsample::Histogram;
 use rand::Rng;
 use std::time::{Duration, Instant};
 use tokio::{runtime, timer};
+use tower::ServiceExt;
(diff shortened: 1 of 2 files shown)


tower-balance/examples/demo.rs
```diff
@@ -89,46 +88,64 @@ fn main() {
...
+            let delay = timer::Delay::new(start + latency);
+
+            delay.map(move |_| {
+                let latency = Instant::now() - start;
+                Rsp { latency }
+            })
+        });
 
-    Disco { changes }
+        InFlightLimit::new(svc, ENDPOINT_CAPACITY)
```
LucioFranco: layer? 😄 or servicebuilder?

carllerche: I will probably update this to `ServiceBuilder` once the misc re-exports / tweaks are applied.

PR #218, opened 2019-03-29
Rename `EitherService` -> `Either`

carllerche:
This removes the dependency on `either` and provides our own enum.

I had originally wanted to name the service `Either`, but doing so would have conflicted with `Either`. Depending on `either` adds a dependency and is of dubious value as we are not able to implement either `Service` or `Future` on it.

Instead, the PR provides an `Either` enum with the necessary implementations.


carllerche:
This is a first step to remove the `Service` suffix from service implementations.

PR #221, opened 2019-03-30
Support weighted balancing

olix0r:
In order to implement red-line testing, blue-green deployments, and
other operational use cases, many service discovery and routing schemes
support weighting.

In this iteration, we provide a decorator type, `Weighted`, that may be
used to wrap load-bearing services to alter their load according to a
weight.  The `Weighted` type may also be used to wrap `Discover`
implementations, in which case it will wrap all new services with
`Weighted`.

The demo has been updated to show the effects of node-weighting:
```
REQUESTS=100000
CONCURRENCY=50
ENDPOINT_CAPACITY=50
MAX_ENDPOINT_LATENCIES=[1ms, 10ms, 10ms, 10ms, 10ms, 100ms, 100ms, 100ms, 100ms, 1000ms, ]
WEIGHTS=[ 1, 0.5, 0.5, 1.5, 1.5, 0.5, 0.5, 1.5, 1.5,  1, ]
P2C+PeakEWMA
  [00] 19814  19.8%
  [01] 14508  14.5%
  [02] 14068  14.1%
  [03] 14806  14.8%
  [04] 14567  14.6%
  [05]  5527   5.5%
  [06]  5536   5.5%
  [07]  5553   5.6%
  [08]  5539   5.5%
  [09]    82   0.1%
  wall   30s
  p50     5ms
  p90    55ms
  p95    78ms
  p99    96ms
  p999  111ms
P2C+PeakEWMA weighted
  [00] 19619  19.6%
  [01] 12273  12.3%
  [02] 13131  13.1%
  [03] 16470  16.5%
  [04] 16322  16.3%
  [05]  3580   3.6%
  [06]  3649   3.6%
  [07]  7475   7.5%
  [08]  7368   7.4%
  [09]   113   0.1%
  wall   30s
  p50     5ms
  p90    55ms
  p95    78ms
  p99    96ms
  p999  168ms
```


tower-balance/src/weight.rs
```diff
@@ -0,0 +1,167 @@
...
+#[derive(Clone, Debug, PartialEq, PartialOrd)]
+pub struct Weighted<T> {
+    inner: T,
+    weight: Weight,
+}
+
+#[derive(Debug)]
+pub struct WithWeighted<T>(T);
+
+pub trait HasWeight {
```
hawkw: This might be overly nitpicky, but the distinction between `HasWeight` and `Weighted` is not particularly clear to me from the names...I might have called this trait `Weighted` and the service type `WeightedService`...TIOLI of course.

tower-balance/src/weight.rs
```diff
@@ -0,0 +1,167 @@
+use futures::{Async, Poll};
+use std::ops;
+use tower_discover::{Change, Discover};
+use tower_service::Service;
+
+use Load;
+
+/// A weight on [0.0, ∞].
```
seanmonstar: It looks like `Weight::MIN` is `0.0`. What happens then? With the `impl Div<Weight> for Cost` and `Count`, it looks like that would yield `NAN`. What does that mean? (I hate IEEE 754 for introducing NAN....)

olix0r: We treat this as `INFINITY`

PR #222, opened 2019-03-31
re-export layers from `tower` crate.

carllerche:
This is intended to be a first step in providing a comprehensive `tower` crate. The details of how the crate is organized will be iterated on in future PRs.


tower/src/lib.rs
```diff
@@ -1,3 +1,7 @@
+// Allows refining features in the future without breaking backwards
+// compatibility
+#![cfg(feature = "full")]
```
LucioFranco: not sure I understand the goal of this feature? What does it disable if i set no default features?

carllerche: It makes the crate empty if `no-default-features` is set. It's a bit paranoid, but technically if types are removed from `no-default-features` but enabled as part of the default feature set, it is a breaking change. It would be very dumb for someone to do so, but at least this guards against it...

LucioFranco: Not sure I understand the difference between that and say just exporting `Service` in a no-default-features setup.

carllerche: There isn’t much difference... I’ve been thinking that the proposed strategy could be a general pattern.

At any point in the future we could include Service with no-default-features and it is backwards compat but not the other way around. 

In our case, including Service is pretty safe.

tower/src/balance.rs
```diff
@@ -0,0 +1 @@
+pub use tower_balance::*;
```
seanmonstar: Is it preferable to have module with glob re-exports, versus `pub extern crate tower_balance as balance`?

seanmonstar: I don't care too much, just seemed like a re-export of the whole crate was easier. :shrug:

PR #223, opened 2019-04-02
retry: Call poll_ready on the inner service

LucioFranco:


carllerche:
test? :)

tower-rate-limit/src/lib.rs
```diff
@@ -82,7 +82,7 @@ where
 
     fn poll_ready(&mut self) -> Poll<(), Self::Error> {
         match self.state {
-            State::Ready { .. } => return Ok(().into()),
+            State::Ready { .. } => return self.inner.poll_ready().map_err(Into::into),
```
hawkw: TIOLI: you could probably use 
```suggestion
            State::Ready { .. } => return Ok(self.inner.poll_ready()?),
```
rather than the explicit `into`, but I am not sure if this is actually better style or not...

seanmonstar: It doesn't work, because `Into` does not imply `From` (the other way around works).

LucioFranco: I personally also find the map_err stuff to be a bit easier to read.

hawkw: @seanmonstar ah yeah, nevermind!

PR #224, opened 2019-04-03
tower: Reexport all layers and add layer::util mod

LucioFranco:
This makes it easier to import all the layers when using the service builder.  This also moves `Chain` and `Identity` to live under `tower::layer::util::{Chain, Identity}`.

I personally find the export of layers here as well to be really handy and a lot easier to write. It also allows you to bulk import like so `tower::layer::{TimeoutLayer, RetryLayer}`.

### DIFF
--- a/tower/src/layer.rs
+++ b/tower/src/layer.rs
@@ -1,9 +1,19 @@
 //! A collection of `Layer` based tower services
 
 pub use tower_layer::Layer;
-// TODO: Should this be re-exported?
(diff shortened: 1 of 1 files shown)


carllerche:
I was hoping to avoid this (and avoid spreading layer related types across multiple mods).

I was planning on adding fns on `ServiceBuilder` for each layer that is included in `tower` proper. This avoids requiring an import at all.

PR #227, opened 2019-04-03
Provide layer specific fns on `ServiceBuilder`

carllerche:
This allows avoiding the layer type imports.

### DIFF
--- a/tower/src/builder/mod.rs
+++ b/tower/src/builder/mod.rs
@@ -4,10 +4,20 @@ mod service;
 
 pub use self::service::{LayeredMakeService, ServiceFuture};
 
+use buffer::BufferLayer;
+use filter::FilterLayer;
+use in_flight_limit::InFlightLimitLayer;
+use load_shed::LoadShedLayer;
+use rate_limit::RateLimitLayer;
+use retry::RetryLayer;
+use timeout::TimeoutLayer;
+
 use tower_layer::Layer;
 use tower_service::Service;
 use tower_util::layer::{Chain, Identity};
 use tower_util::MakeService;
+
+use std::time::Duration;
 
 pub(super) type Error = Box<::std::error::Error + Send + Sync>;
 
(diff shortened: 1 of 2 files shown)


tower/src/builder/mod.rs
```diff
@@ -172,6 +182,41 @@ impl<L> ServiceBuilder<L> {
         }
     }
 
+    /// Buffer requests
+    pub fn buffer(self, bound: usize) -> ServiceBuilder<Chain<BufferLayer, L>> {
+        self.layer(BufferLayer::new(bound))
+    }
+
+    /// Filter requests
+    pub fn filter<U>(self, predicate: U) -> ServiceBuilder<Chain<FilterLayer<U>, L>> {
```
LucioFranco: should we have a where clause here to bound what `U` should be to not require the user to go to `FilterLayer::new` to find the bound?

tower/src/builder/mod.rs
```diff
@@ -172,6 +182,41 @@ impl<L> ServiceBuilder<L> {
...
+        self.layer(LoadShedLayer::new())
+    }
+
+    /// Limit requests to at most `num` per the given duration
+    pub fn rate_limit(self, num: u64, per: Duration) -> ServiceBuilder<Chain<RateLimitLayer, L>> {
+        self.layer(RateLimitLayer::new(num, per))
+    }
+
+    /// Retry failed requests.
+    pub fn retry<P>(self, policy: P) -> ServiceBuilder<Chain<RetryLayer<P>, L>> {
```
LucioFranco: same here we should bound this by Policy

tower/src/builder/mod.rs
carllerche: We cannot add bounds because we don't know the type of `Request`.

carllerche:
I added some documentation covering ordering and the implicit trait bounds.

PR #229, opened 2019-04-04
tower: Add `buffer_with_executor` to the `ServiceBuilder`

LucioFranco:
Not sure if this is the right approach to this but I can see myself using this in the near future since I will want to spawn a task without being inside a futures context.

That said, it may be better to just import tower_buffer and call the with_executor layer but this seems to work with the annotation.


carllerche:
I considered this when I did the original work but landed on requiring manually constructing the layer. 

Is there a reason why you are not in a spawnable context?

PR #235, opened 2019-04-04
Consolidate `limit` layers

### LINKED ISSUE #225: Consider renaming `InFlightLimit` -> `LimitInFlight`
The same would apply to `RateLimt` -> `LimitRate`.

Having the name **start** with the action performed reads better IMO.

Also, we should determine if layers are named as verbs or nouns. Many names can read as both.

carllerche:
- `InFlightLimit` and `RateLimit` are moved into `tower-limit` crate.
- `InFlightLimit` is renamed to `ConcurrencyLimit`.

Fixes #225


seanmonstar:
I see the purpose, which is to have it be verb-noun, but "rate limit" is *the* way the industry calls that technique...

carllerche:
@seanmonstar 

"Rate limit" refers to the maximum rate that a service accepts, i.e. the maximum number of requests per some time period. The thing that enforces the rate limit is the "rate limiter". Idiomatically, we are avoiding -er suffixes for layers.

IMO the only options are to either rename it "tower-limit-rate" or switch all layers to nouns, which would mean a lot of -er suffixes (tower-rate-limiter, tower-balancer, ...).

Also, "limit-rate" clearly ties to the "rate limit" term.

seanmonstar:
> or switch all layers to nouns

Only if it's a hard requirement that all layers must be verbing-nouns...

My feeling is `concurrency_limit(n)` and `rate_limit(x)` read nicely.

carllerche:
I actually had the opposite reaction (my brain did not know how to read it)
which prompted this.

carllerche:
Anyway, I say survey the thought of others. If I’m in the minority then we
can keep it RateLimit

LucioFranco:
I agree with sean on this one, I prefer `rate_limit` and `in_flight_limit`, I'm personally not very hard set on keeping things nouns or verbs, whatever reads easier to me is what I prefer.

jonhoo:
I think either is fine -- no strong preference.

davidbarsky:
I'm in favor of these changes because `LimitConcurrency` more accurately describes what `InFlightLimit` _does_—concurrency control. Rate limiting is but a subset of that, and I'm a fan of applying control theory wherever I go.

carllerche:
@davidbarsky the main point of debate is `RateLimit` vs. `LimitRate` (and `ConcurrencyLimit` vs. `LimitConcurrency`).

davidbarsky:
@carllerche: sorry, scatterbrained due to work. I like placing `Limit` first because I like the sense that it’s a verb—that always did stick out as a weird change despite `RateLimit` being the norm—internal consistency within Tower matters more than compliance with (adhoc)  has industry norms.

seanmonstar:
I don't want to dig in, I just felt that "rate limit" is what people would expect to search for, not "limit rate". 🤷‍♂️

carllerche:
@seanmonstar I don't want to dig in either :P

olix0r:
@carllerche asked me to chime on this, so let's quickly explore the grammatical underpinnings here...

In "rate limit" and "concurrency limit", I see "rate" and "concurrency" as adjectives -- they describe the kind of limit (noun). To verbify this, we'd "_enforce_ a request rate limit" or we could "_limit_ the rate of requests".

So, with this in mind, I'd lean towards "limit rate" and not "rate limit".

That said, I deeply don't care ;)

adleong:
For what it's worth, I feel like "rate limit" has become an accepted verb-phrase.  e.g. "We must rate limit the requests."  Therefore, I would lean toward calling "rate limit" which is the more common term and is commonly used as a verb.

"Concurrency limit" follows the same structure and while it's not a commonly used phrase, it could be used the same way as a verb also.  e.g. "We must concurrency limit the requests."

olix0r:
@adleong verbing nouns weirds language

adleong:
I think the confusion arises because "limit" can be either a noun ("The limit is seven.") or a verb ("You should limit exposure.")  Therefore "rate limit" can be a verb-phrase with "rate" being an adverb and "limit" being a verb ("We must rate limit the requests.") or it can be a noun-phrase with "rate" being an adjective and "limit" being a noun ("The rate limit is 100rps.")

hawkw:
+1 that "rate-limit" is a verb.

PR #236, opened 2019-04-05
Add hedge retry middleware

adleong:
Add tower-hedge, a middleware that pre-emptively retries requests which have been outstanding for longer than a given latency percentile.  If either of the original future or the retry future completes, that value is used.  For more information about hedge requests, see: [The Tail at Scale](https://cseweb.ucsd.edu/~gmporter/classes/fa17/cse124/post/schedule/p74-dean.pdf)

This PR replaces https://github.com/tower-rs/tower/pull/192 and is a more modular solution.

Signed-off-by: Alex Leong <alex@buoyant.io>


tower-hedge/src/lib.rs
```diff
@@ -0,0 +1,199 @@
+extern crate futures;
```
seanmonstar: If you slap these attributes at the top, it catches a few common things we want to try to always provide:

```rust
#![deny(warnings)]
#![deny(missing_docs)]
#![deny(missing_debug_implementations)]
```

tower-hedge/src/lib.rs
```diff
@@ -0,0 +1,199 @@
...
+        policy,
+        histo,
+        min_data_points,
+    };
+    Select::new(select_policy, recorded_a, delayed)
+}
+
+// TODO: Remove when Duration::as_millis() becomes stable.
+fn duration_as_millis(d: Duration) -> u64 {
+    d.as_secs() * 1_000 + d.subsec_millis() as u64
```
seanmonstar: We may want to use checked arithmetic here, to protect again overflows in production (overflow detection is disabled when `--release`).

tower-hedge/src/rotating_histogram.rs
```diff
@@ -0,0 +1,71 @@
...
+        trace!("Rotated {:?} points into read", self.read.len());
+        self.write.clear();
+    }
+
+    fn clear(&mut self) {
+        self.read.clear();
+        self.write.clear();
+    }
+
+    fn duration_as_nanos(d: &Duration) -> u64 {
```
seanmonstar: Perhaps this can be a free function similar to `duration_as_millis` above? Also, same thing about checked arithmetic.

tower-hedge/src/lib.rs
```diff
@@ -0,0 +1,199 @@
...
+    /// clone_request is called when the request is first received to determine
+    /// if the request is retryable.
+    fn clone_request(&self, req: &Request) -> Option<Request>;
+    /// can_retry is called after the hedge timeout to determine if the hedge
+    /// retry should be issued.
+    fn can_retry(&self, req: &Request) -> bool;
+}
+
+#[derive(Clone)]
+pub struct PolicyPredicate<P>(P);
```
seanmonstar: Are these types supposed to be part of the public API? Or are they only `pub` because they needed to be to satisfy `pub type Service = ..`? If the latter, we could make these all private by change `type Service` into an opaque wrapper `struct` instead.

tower-hedge/src/lib.rs
```diff
@@ -0,0 +1,199 @@
...
+    policy: P,
+    histo: Histo,
+    min_data_points: u64,
+}
+
+/// A middleware that pre-emptively retries requests which have been outstanding
+/// for longer than a given latency percentile.  If either of the original
+/// future or the retry future completes, that value is used.
+pub fn service<S, P, Request>(
+    service: S,
```
seanmonstar: This is more of a personal preference: since there are lot of arguments here, what do you think of a `Builder` style instead? For instance:

```rust
let svc = tower::hedge::service(svc, policy, 10, 0.9, Duration::from_secs(10));

// vs

let svc = tower::hedge::Builder::new(svc, policy)
    .min_data_points(10)
    .latency_percentile(0.9)
    .period(Duration::from_secs(10))
    .build();
```

tower-hedge/src/delay/mod.rs
```diff
@@ -0,0 +1,95 @@
...
+    fn poll_ready(&mut self) -> Poll<(), Self::Error> {
+        self.service
+            .poll_ready()
+            .map_err(|e| Error::ServiceError(e.into()))
+    }
+
+    fn call(&mut self, request: Request) -> Self::Future {
+        let deadline = tokio_timer::clock::now() + self.policy.delay(&request);
+        ResponseFuture {
+            service: self.service.clone(),
```
carllerche: The instance that `poll_ready` was called on needs to be the one passed to `ResponseFuture`. You probably want to call `clone`, then use `mem::swap` to place the cloned service onto `self`.

If tests pass using an up-to-date version of `tower_test::mock`, then there is a bug in the mock implementation. It should catch this case.

tower-hedge/src/delay/error.rs
```diff
@@ -0,0 +1,27 @@
+//! Error types
+
+use std::fmt;
+
+#[derive(Debug)]
+pub enum Error {
```
carllerche: Do you need this enum? Can `Delay`'s error be `Box<Error + Send + Sync>`?

tower-hedge/src/delay/error.rs
```diff
@@ -0,0 +1,27 @@
...
+            Error::TimerError(e) => Some(e),
+            Error::ServiceError(e) => Some(&**e),
+        }
+    }
+}
+
+impl fmt::Display for Error {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match self {
+            Error::TimerError(e) => write!(f, "timer error in delay: {:?}", e),
```
carllerche: The `Display` implementation should not forward to the inner error if `source` also returns the inner error. The user is expected to iterate the `source` chain if they want the error message to be of the format "top level error: lower level cause".

tower-hedge/src/lib.rs
```diff
@@ -0,0 +1,199 @@
...
+mod rotating_histogram;
+mod select;
+
+use delay::Delay;
+use latency::Latency;
+use rotating_histogram::RotatingHistogram;
+use select::Select;
+
+type Histo = Arc<Mutex<RotatingHistogram>>;
+pub type Service<InnerSvc, P> = select::Select<
```
carllerche: The current convention is to name the service after the function (ideally a verb), so in this case, `Hedge` and to make it a type that contains the inner composition. So:

```rust
pub struct Hedge<T, P> {
    inner: Service<T, P>, // where `Service` is the current alias
```

tower-hedge/src/lib.rs
```diff
@@ -0,0 +1,199 @@
...
+pub struct SelectPolicy<P> {
+    policy: P,
+    histo: Histo,
+    min_data_points: u64,
+}
+
+/// A middleware that pre-emptively retries requests which have been outstanding
+/// for longer than a given latency percentile.  If either of the original
+/// future or the retry future completes, that value is used.
+pub fn service<S, P, Request>(
```
carllerche: This would be `fn new` on `Hedge`.

tower-hedge/src/lib.rs
```diff
@@ -0,0 +1,199 @@
...
+use delay::Delay;
+use latency::Latency;
+use rotating_histogram::RotatingHistogram;
+use select::Select;
+
+type Histo = Arc<Mutex<RotatingHistogram>>;
+pub type Service<InnerSvc, P> = select::Select<
+    SelectPolicy<P>,
+    Latency<Histo, InnerSvc>,
+    Delay<DelayPolicy, Filter<Latency<Histo, InnerSvc>, PolicyPredicate<P>>>,
```
carllerche: Very nice to see `Filter` used here :)

tower-hedge/src/lib.rs
```diff
@@ -0,0 +1,199 @@
...
+                error!("Failed to write to hedge histogram: {:?}", e);
+            })
+    }
+}
+
+impl<P, Request> tower_filter::Predicate<Request> for PolicyPredicate<P>
+where
+    P: Policy<Request>,
+{
+    type Future = future::Either<
```
carllerche: I've been thinking that `Predicate` should not be async. There is no use case yet. This would simplify things here. Thoughts?

tower-hedge/src/lib.rs
carllerche:
```rust
    pub fn millis(duration: Duration) -> u64 {
        // Round up.
        let millis = (duration.subsec_nanos() + NANOS_PER_MILLI - 1) / NANOS_PER_MILLI;
        duration.as_secs().saturating_mul(MILLIS_PER_SEC).saturating_add(u64::from(millis))
}
```

tower-hedge/src/lib.rs
adleong: Yeah, this being non-async would be more convenient.  An async version could be useful for something that makes calls out to an external policy service perhaps... but maybe that could be a different middleware.

tower-hedge/src/lib.rs
adleong: Good call, thanks.  I've added opaque wrappers for the Service and Future types.

tower-hedge/src/lib.rs
adleong: Builders feel a bit boilerplate-y to me and 5-6 arguments isn't *that* many...

tower-hedge/src/lib.rs
adleong: Adding debug implementations for everything turned out to be a pain, so I didn't do it.  I've derived debug in a bunch more places though.

seanmonstar: FWIW, in some cases I just provide a debug impl that outputs the name of the struct and nothing else. I'd recommend having them on services, requests, and responses. Futures are whatever.

PR #237, opened 2019-04-05
Create `tower-test` and include tower-mock.

carllerche:
tower-mock is deleted in favor of having a single test crate. Once `tokio-test` is released, I would expect that it will be re-exported as part of `tower-test`.

Some changes are made to mock:

- `mock::pair()` is used to get the two handles and `Mock::new` is deleted.
- `Respond` is renamed to `SendResponse`.
- The fns are renamed `send_response` and `send_error`.
- The `Request` struct is removed in favor of just using a tuple.

This crate also includes a new macro: `assert_request_eq!` for working with the mock handle:

```rust
assert_request_eq!(mock_handle, "hello").send_response("world");
```

Not all the crates have been updated yet.


tower-test/src/lib.rs
```diff
@@ -0,0 +1,8 @@
+//! Mock `Service` that can be used in tests.
+
+extern crate futures;
+extern crate tokio_sync;
+extern crate tower_service;
+
+mod macros;
```
LucioFranco: does this support getting the macro via an import?

carllerche: Not sure... I would guess so? It would be worth a test. We also probably should update everything to 2018...

carllerche: If it is not, we can update it with #238.

tower-test/src/macros.rs
```diff
@@ -0,0 +1,15 @@
+#[macro_export]
+macro_rules! assert_request_eq {
```
LucioFranco: might want docs here?

carllerche: Yeah. Probably... thanks for calling that out.

PR #239, opened 2019-04-08
Rename `Chain` -> `Stack`

### LINKED ISSUE #233: Consider renaming `Chain` -> `Stack`
"chain" is mixing metaphors.

carllerche:
Try to reduce metaphores.

Fixes #233

### DIFF
--- a/tower/src/layer.rs
+++ b/tower/src/layer.rs
@@ -3,6 +3,6 @@
 pub use tower_layer::Layer;
 
 pub mod util {
-    pub use tower_util::layer::Stack;
     pub use tower_util::layer::Identity;
+    pub use tower_util::layer::Stack;
 }
--- a/tower/src/layer.rs
+++ b/tower/src/layer.rs
@@ -3,7 +3,7 @@
 pub use tower_layer::Layer;
 
 pub mod util {
-    pub use tower_util::layer::Chain;
+    pub use tower_util::layer::Stack;
     pub use tower_util::layer::Identity;
 }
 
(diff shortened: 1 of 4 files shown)


tower/src/layer.rs
```diff
@@ -14,12 +14,12 @@ pub trait LayerExt<S, Request>: Layer<S, Request> {
     /// `middleware` to services being wrapped.
     ///
     /// This defines a middleware stack.
-    fn chain<T>(self, middleware: T) -> util::Chain<Self, T>
+    fn chain<T>(self, middleware: T) -> util::Stack<Self, T>
```
hawkw: It's possible we've already discussed this, but I feel like the type and method ought to share the same name?
```suggestion
    fn stack<T>(self, middleware: T) -> util::Stack<Self, T>
```

LucioFranco: 👍 on this

carllerche: Ah yes... but I was thinking about getting rid of that fn. I don't think it is needed w/ `ServiceBuilder`.

carllerche: Thanks for catching this though.

LucioFranco: I still think it could be useful 🤷‍♂️ I don't see why there is a reason to remove it over keeping it.

carllerche:
I'm going to just remove `LayerExt`.

tower/src/layer.rs
carllerche: Can always bring it back later. I rather stay on the conservative side initially.

LucioFranco:
Not sure if there is a benefit to removing it? I personally can see people maybe using it 🤷‍♂️

tower/src/layer.rs
carllerche: There is only 1 fn right now... `Stack::new` can be called directly, and `ServiceBuilder` handles the use case too.

LucioFranco:
but if removing it means that we might add it later then im fine with that.

PR #247, opened 2019-04-10
buffer: Add BufferLazy to lazily spawn the worker

### LINKED ISSUE #242: buffer: Provide mechanism to lazily spawn the worker
Since `DefaultExecutor::current` is lazy, it should be possible to delay the spawning of the worker task until the point of the first poll_ready/call.

LucioFranco:
This provides a "lazy" buffer that will only spawn the background worker on the first call to `poll_ready`. The advantage of this is that it allows one to create the service stack outside of a futures context and then call that service stack from within a futures context will provide all the benefits of a `Service + Clone`.

Closes #242


LucioFranco:
after discussion with carl on gitter I'm gonna close this and move this into our repo locally while we still discover this path.

PR #254, opened 2019-04-16
WIP: Add Routing middleware

Abogical:
Idea by @carllerche.
I'm still failing at implementing the `Layer` trait for the last couple of days. I can't have a structure that can recursively layer itself, as `Service` expects a mutable reference and `Layer` expects a non-mutable reference. This doesn't seem to have a clean solution, so I'm not sure if this middleware would make things any easier.

### DIFF
--- a/tower-route/src/lib.rs
+++ b/tower-route/src/lib.rs
@@ -1,0 +1,37 @@
+extern crate futures;
+extern crate tower_service;
+extern crate void;
+
+use futures::{future::Either, Async, Poll};
+use tower_service::Service;
+
+pub trait Accepts<Request> {
+    fn accepts(&self, request: &Request) -> bool;
+}
+
+pub struct Route<A, B> {
+    service: A,
+    other: B,
+}
+
+impl<Request, A, B> Service<Request> for Route<A, B>
+where
+    A: Service<Request> + Accepts<Request>,
+    B: Service<Request, Response = A::Response, Error = A::Error>,
+{
+    type Response = A::Response;
+    type Error = A::Error;
+    type Future = Either<A::Future, B::Future>;
+
+    fn poll_ready(&mut self) -> Poll<(), Self::Error> {
+        Ok(Async::Ready(()))
+    }
+
+    fn call(&mut self, request: Request) -> Self::Future {
+        if self.service.accepts(&request) {
+            Either::A(self.service.call(request))
+        } else {
+            Either::B(self.other.call(request))
+        }
+    }
+}
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,6 +15,7 @@ members = [
   "tower-test",
   "tower-timeout",
   "tower-util",
+  "tower-route"
 ]
 
 [patch.'https://github.com/tower-rs/tower']
--- a/tower-route/Cargo.toml
+++ b/tower-route/Cargo.toml
@@ -1,0 +1,11 @@
+[package]
+name = "tower-route"
+version = "0.1.0"
+authors = ["Abdel-Rahman A. <abogical@gmail.com>"]
+edition = "2018"
+
+[dependencies]
+futures = "0.1.26"
+tower-service = "0.2.0"
+tower-layer = { path = "../tower-layer" }
+void = "1.0.2"


tower-route/src/lib.rs
```diff
@@ -0,0 +1,42 @@
...
+
+use futures::{Async, Poll};
+use tower_service::Service;
+
+pub trait Acceptor<Request> {
+    fn accepts(&self, request: &Request) -> bool;
+}
+
+pub struct Route<'a, A, B> {
+    service: &'a mut A,
```
LucioFranco: We probably want this to own A and B?

tower-route/src/lib.rs
```diff
@@ -0,0 +1,42 @@
...
+    other: &'a mut B,
+}
+
+impl<'a, A, B, Request, Response, Future, Error> Service<Request> for Route<'a, A, B>
+where
+    Future: futures::Future<Item = Response, Error = Error>,
+    A: Service<Request, Response = Response, Future = Future, Error = Error> + Acceptor<Request>,
```
LucioFranco: might make more sense to have a separate acceptor struct, instead of forcing this to be bundled with `Service`? This wouldn't say allow you to erase the type into a `impl Service` I believe.

tower-route/src/lib.rs
```diff
@@ -0,0 +1,42 @@
...
+    B: Service<Request, Response = Response, Future = Future, Error = Error>,
+{
+    type Response = Response;
+    type Error = Error;
+    type Future = Future;
```
LucioFranco: you probably want this to be `type Future = Either<A::Future, B::Future>`.

tower-route/src/lib.rs
Abogical: Sorry for the delay. I've made the other suggestions but I'm not exactly sure what you mean by this one.

Abogical:
@LucioFranco @carllerche 

I'm having the idea to replace the `Acceptor` trait with a trait that simply returns `Option<Service>` instead of just a boolean. It simply maps requests to a service, and returns `None` if it doesn't accept the request. This way the Service doesn't have to implement `Acceptor`. I may call this trait `ServiceMapper` or something like that. What do you think?

```Rust
trait ServiceMapper{
    fn map(request: &Request) -> Option<Service>
}
```

tower-route/src/lib.rs
```diff
@@ -0,0 +1,43 @@
...
+pub trait Acceptor<Request> {
+    fn accepts(&self, request: &Request) -> bool;
+}
+
+pub struct Route<A, B> {
+    service: A,
+    other: B,
+}
+
+impl<A, B, Request, Future, Response, Error> Service<Request> for Route<A, B>
```
carllerche: I don’t think you need most of these generics. You can say B: Service<Request, Reaponse = A::Response, ...>

You also don’t want the future types to be the same or each service won’t be able to have different behavior :-)

tower-route/src/lib.rs
```diff
@@ -0,0 +1,43 @@
...
+    Future: futures::Future<Item = Response, Error = Error>,
+    A: Service<Request, Future = Future, Response = Response, Error = Error> + Acceptor<Request>,
+    B: Service<Request, Future = Future, Response = Response, Error = Error>,
+{
+    type Response = Response;
+    type Error = Error;
+    type Future = Either<A::Future, B::Future>;
+
+    fn poll_ready(&mut self) -> Poll<(), Self::Error> {
+        match self.service.poll_ready() {
```
carllerche: For readiness, the strategy to use is poll_ready is always ready, and the response future uses Oneshot in tower-util.

carllerche:
@Abogical We initially had a strategy like that. See: https://github.com/linkerd/linkerd2-proxy/tree/master/lib/router

This is valid, but covers a different use case: routing between many services of the **same type**.

In this PR, we are trying to add a router for services of **different** types. Because of this, the proposed fn (`fn map(request: &Request) -> Option<Service>`) doesn't work.

tower-route/src/lib.rs
Abogical: IIUC, why would you want to want to wrap the services with Oneshot? I'll have to borrow the services out of their `Route` struct to construct it.

Abogical: The problem with `Oneshot` is that it must own the services inside the route in order to construct it. So the services must also implement `Copy`. Is that ok?

LucioFranco:
Hey! I have an example of this type of routing implemented within tonic. I am going to close this PR feel free to reopen against the appropriate branch as master will become the std::future::Future branch.

PR #256, opened 2019-04-19
tower: Fix client example

LucioFranco:
So the return type has changed to `LiftBody` which honestly we should just rename to `tower::body::Body`. LiftBody impls HttpBody and wraps a Payload type.


carllerche:
Superseded by #261

PR #257, opened 2019-04-19
buffer: make Buffer::new infallible

seanmonstar:
Creating a buffer can internally fail to spawn a worker. Before, that
error was returned immediately from `Buffer::new`. This changes `new` to
always return a `Buffer`, and the spawn error is encountered via
`poll_ready`.

### DIFF
--- a/tower-buffer/src/layer.rs
+++ b/tower-buffer/src/layer.rs
@@ -41,6 +41,10 @@ where
     type Service = Buffer<S, Request>;
 
     fn layer(&self, service: S) -> Result<Self::Service, Self::LayerError> {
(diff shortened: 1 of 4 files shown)


davidbarsky:
I'm in favor of this change, in that failing to spawn a worker is a runtime issue, not a construction-time issue. Is this motivated by the Layer changes?

I'd also like to hear what @jonhoo thinks.

seanmonstar:
Yes, this is inspired by noticing that the only layer that could fail is buffer.

jonhoo:
The one downside to this is that it's now slightly harder to track down _where_ we were when we failed to spawn, since the place we get the error is now disassociated from where the error occurred.

PR #260, opened 2019-04-23
Remove reconnect from tower facade

seanmonstar:


carllerche:
Could you add `publish = false` to the crate?

seanmonstar:
It's already there.

PR #275, opened 2019-04-27
first pass at adding map request functionality

### LINKED ISSUE #272: Implement mapping combinators in tower
I would like to see tower implement a mapping combinator into the `ServiceBuilder` api. Something similar to what we say in PR #251 would be quite nice.

```rust
let mut service = ServiceBuilder::new()
            .map_request(|request: &'static str| request.to_string())
            .layer(FilterLayer::new(|_request: &String| Ok(())))
            .concurrency_limit(100)
            .buffer(5)
            .service(my_service);
```

I've been hacking on it locally, and it seems pretty easy to add in. I'm currently just fighting some generic type declarations. I'm currently playing with the idea of the mapping function being `Fn(Request) -> Future` (where I'm a little fuzzy on the exact return type).

drusellers:
closes #272 

Still wrapping my head around futures in rust, but this should give us something to discuss more concretely.


tower-map/src/lib.rs
```diff
@@ -0,0 +1,18 @@
...
+#![cfg_attr(test, deny(warnings))]
+#![allow(elided_lifetimes_in_paths)]
+
+//! Tower middleware for converting the request type
+
+mod layer;
+mod service;
+mod error;
+
+pub use error::Error;
```
carllerche: This does not need to be pub.

tower-map/src/service.rs
```diff
@@ -0,0 +1,47 @@
...
+  /// Create a new rate limiter
+  pub fn new(m: M, inner: S) -> Self {
+    MapRequest { m, inner }
+  }
+}
+
+impl<S, M, F, O, Request> Service<Request> for MapRequest<S, M>
+where
+  // Func(Request) -> Future<S::Request, ?>
+  M: FnMut(Request) -> F,
```
carllerche: with futures,  mapping is not asynchronous. AndThen is used for asynchronous mapping. should we follow this pattern? I.e. MapRequest would not be Async and we would have a different util for Async mapping.

tower-map/src/service.rs
```diff
@@ -0,0 +1,47 @@
...
+    self.inner.poll_ready().map_err(Into::into)
+  }
+
+  fn call(&mut self, request: Request) -> Self::Future {
+    // convert the inbond request to the downstream
+    // request. Then handle that the conversion
+    // returns a future as well.
+    let transformed = (self.m)(request).into_future();
+
+    transformed.and_then(|x| self.inner.call(x)).map_err(|e| e.into())
```
carllerche: Closures cannot be named. This means they cannot be used to build a future that must be named as an associated type. This is why we impl all our futures by hand.

drusellers: I was wondering why that was. Thank you. :)

tower-map/src/service.rs
drusellers: @carllerche i'll make it synchronous for now, that would make things a bit easier. We can as you say, have a different utility for async mapping.

drusellers:
> Also, I wonder if we should define our own Fn trait with a blanket impl for ‘T: Fn(...) -> ...’ this way we can use MapRequest while still being able to name the type.

That sounds like a win. It lets us easily track usage and pin down exactly what we want the function to be without having to remember the "longer" type description. I'll read up on this and see about making it happen.

In the mean time I believe I have the other changes your requested implemented.

tower-util/src/map/layer.rs
```diff
@@ -0,0 +1,26 @@
...
+impl<M> MapRequestLayer<M> {
+  /// Create a new `MapRequestLayer` from a closure
+  pub fn new(m: M) -> Self {
+    MapRequestLayer { m }
+  }
+}
+
+impl<S, M> Layer<S> for MapRequestLayer<M>
+where
+  M: Clone,
```
LucioFranco: I wonder if boxing is better here?

drusellers: @LucioFranco i'll leave this question for wiser Rust people than myself. Happy to do what y'all think is best. Working on adding the custom Fn trait now. :)

drusellers:
@carllerche re the trait for the Fn would that be something akin to this (this is a struct not a trait)

```
/// A `Service` implemented by a closure.
#[derive(Copy, Clone, Debug)]
pub struct ServiceFn<T> {
    f: T,
}
```

Having a hard time googling it and wondering if it just means a type alias?

LucioFranco:
Looks like this PR has gone a bit stale! We are transitioning master to support std::future::Future, feel free to reopen this against the proper branch.

PR #281, opened 2019-05-09
balance: Configure weights from keys, not services

olix0r:
The initial weighted balancing implementation required that the
underlying service implement `HasWeight`.

Practically, this doesn't work that well, since this may force
middlewares to implement this trait as well.

To fix this, we change the type bounds so that _keys_, not services,
must implement `HasWeight`.

This has a drawback, though, in that Weight, which contains a float,
cannot implement `Hash` or `Eq`, which is required by the balancer. This
tradeoff seems manageable, though (and is already addressed in linkerd,
for instance).

### DIFF
--- a/tower-balance/examples/demo.rs
+++ b/tower-balance/examples/demo.rs
@@ -4,10 +4,14 @@ use env_logger;
 use futures::{future, stream, Future, Stream};
 use hdrsample::Histogram;
 use rand::{self, Rng};
-use std::{cmp, hash};
 use std::time::{Duration, Instant};
+use std::{cmp, hash};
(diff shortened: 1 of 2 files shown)


seanmonstar:
> This has a drawback, though, in that Weight, which contains a float,
cannot implement Hash or Eq, which is required by the balancer.

What if `Weight` didn't wrap a float, but a `u64` instead?

olix0r:
@seanmonstar Yeah, I agree that's a good change we should pursue. I think that's sufficiently separate, though, that we can merge this and address that in a followup.

PR #282, opened 2019-05-10
balance: Internally represent weights as integers

olix0r:
The balancer's `Weight` type represents a weight value as a float
because, conceptually, we want to think of a unit weight of 1.0...

However, storing a float means that this type cannot be hashed, and
therefore all types that hold a Weight cannot implement Hash. Given that
`Balance` expects that keys impl `Hash` and `Eq`, this is especially
cumbersome.

This branch changes the internal representation of weights to be
unsigned integers with a unit value of `10_000` (corresponding to
1.0000).


olix0r:
superseded by #286

PR #283, opened 2019-05-25
Spawn ready: drives a service's readiness on an executor

olix0r:
Some layers cannot guarantee that they will poll inner services in a
timely fashion. For instance, the balancer polls its inner services to
check for readiness, but it does so randomly. If its inner service
must be polled several times to become ready, e.g., because it's driving
the initiation of a TLS connection, then the balancer may not drive the
handshake to completion.

The `SpawnReady` layer ensures that its inner service is driven to
readiness by spawning a background task.

### DIFF
--- a/tower-spawn-ready/src/future.rs
+++ b/tower-spawn-ready/src/future.rs
@@ -1,0 +1,78 @@
+use crate::error::Error;
+use futures::{Async, Future, Poll};
+use tokio_executor::TypedExecutor;
+use tokio_sync::oneshot;
+use tower_service::Service;
+use tower_util::Ready;
+
+pub struct BackgroundReady<T, Request>
+where
+    T: Service<Request>,
+    T::Error: Into<Error>,
+{
+    ready: Ready<T, Request>,
+    tx: Option<oneshot::Sender<Result<T, Error>>>,
+}
+
+/// This trait allows you to use either Tokio's threaded runtime's executor or
+/// the `current_thread` runtime's executor depending on if `T` is `Send` or
+/// `!Send`.
+pub trait BackgroundReadyExecutor<T, Request>: TypedExecutor<BackgroundReady<T, Request>>
+where
+    T: Service<Request>,
+    T::Error: Into<Error>,
+{
+}
+
+impl<T, Request, E: TypedExecutor<BackgroundReady<T, Request>>> BackgroundReadyExecutor<T, Request>
+    for E
+where
+    T: Service<Request>,
+    T::Error: Into<Error>,
+{
+}
+
+impl<T, Request> BackgroundReady<T, Request>
+where
+    T: Service<Request>,
+    T::Error: Into<Error>,
+{
+    pub(crate) fn new(service: T) -> (Self, oneshot::Receiver<Result<T, Error>>) {
+        let (tx, rx) = oneshot::channel();
+        let bg = Self {
+            ready: Ready::new(service),
+            tx: Some(tx),
+        };
+        (bg, rx)
+    }
+}
+
+impl<T, Request> Future for BackgroundReady<T, Request>
+where
+    T: Service<Request>,
+    T::Error: Into<Error>,
+{
+    type Item = ();
+    type Error = ();
+
+    fn poll(&mut self) -> Poll<(), ()> {
+        match self.tx.as_mut().expect("illegal state").poll_close() {
+            Ok(Async::Ready(())) | Err(()) => return Err(()),
+            Ok(Async::NotReady) => {}
+        }
+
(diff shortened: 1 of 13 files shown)


tower-spawn-ready/src/future.rs
```diff
@@ -0,0 +1,78 @@
...
+/// the `current_thread` runtime's executor depending on if `T` is `Send` or
+/// `!Send`.
+pub trait BackgroundReadyExecutor<T, Request>: TypedExecutor<BackgroundReady<T, Request>>
+where
+    T: Service<Request>,
+    T::Error: Into<Error>,
+{
+}
+
+impl<T, Request, E: TypedExecutor<BackgroundReady<T, Request>>> BackgroundReadyExecutor<T, Request>
```
hawkw: Nit: Can the bound on `E` go in the `where` clause instead? The rustfmt formatting here is kinda weird, and I think shortening this line would make it less so. Also, it's a little easier to read when all the bounds are in the same place, IMO.

tower-spawn-ready/src/service.rs
```diff
@@ -0,0 +1,110 @@
...
+    T: Service<Request> + Send,
+    T::Error: Into<Error>,
+    E: BackgroundReadyExecutor<T, Request>,
+{
+    /// Creates a new `SpawnReady` wrapping `service`.
+    ///
+    /// `executor` is used to spawn a new `BackgroundReady` task that is
+    /// dedicated to driving the inner service to readiness.
+    ///
+    /// `bound` gives the maximal number of requests that can be queued for the service before
```
hawkw: It looks like there's no longer a parameter called `bound`?

tower-spawn-ready/src/service.rs
```diff
@@ -0,0 +1,110 @@
...
+                        .expect("illegal state")
+                        .poll_ready()
+                        .map_err(Into::into)?
+                        .is_ready()
+                    {
+                        return Ok(Async::Ready(()));
+                    }
+
+                    let (bg, rx) = BackgroundReady::new(svc.take().expect("illegal state"));
+                    self.executor.spawn(bg).map_err(|_| SpawnError::new())?;
```
hawkw: IMO, I think it might be good to propagate information from the executor's `SpawnError` here, or at least log it --- I'd like to know if my spawns are failing due to capacity or if it's being called outside of an executor...

tower-spawn-ready/src/service.rs
```diff
@@ -0,0 +1,110 @@
...
+
+                    Inner::Future(rx)
+                }
+                Inner::Future(ref mut fut) => match try_ready!(fut.poll()) {
```
hawkw: TIOLI: I'm pretty sure this match could just be
```rust
Inner::Future(ref mut fut) => {
    let svc = try_ready!(fut.poll())?;
    Inner::Service(Some(svc))
}
```
or even
```rust
Inner::Future(ref mut fut) => Inner::Service(Some(try_ready!(fut.poll())?)),
```
although that's admittedly a lot to read...

tower-spawn-ready/src/future.rs
```diff
@@ -0,0 +1,78 @@
...
+    for E
```
hawkw: Nit: The way rustfmt chose to format this is a little weird IMO. Maybe moving the bounds on `E` into the `where` clause would shorten the first line, making it not get rustfmt-ed this way? Also, IMO, it's a little easier to grok when all the bounds are in the same place.

tower-spawn-ready/src/service.rs
hawkw: It looks like this no longer takes a parameter called `bound` --- is this comment stale?

tower-spawn-ready/src/future.rs
```diff
@@ -0,0 +1,78 @@
...
+    T::Error: Into<Error>,
+{
+}
+
+impl<T, Request> BackgroundReady<T, Request>
+where
+    T: Service<Request>,
+    T::Error: Into<Error>,
+{
+    pub(crate) fn new(service: T) -> (Self, oneshot::Receiver<Result<T, Error>>) {
```
carllerche: bikeshed: I've been tending to move away from `new` fns that return tuples and favoring mod fns in those cases... this is crate private, so not a big deal.

tower-spawn-ready/src/service.rs
olix0r: FWIW, this is copied directly from `tower-buffer`...

hawkw: oh, okay...I may open a ticket to propagate spawn error reasons in buffer as well, then.

PR #285, opened 2019-05-28
Extract tower-load from tower-balance

olix0r:
The tower-balance crate includes the `Load` and `Instrument` traits,
which are likely useful outside of balancers; and certainly have no
tight coulping with any specific balancer implementation. This change
extracts these protocol-agnostic traits into a dedicated crate.

The `Load` trait includes a latency-aware _PeakEWMA_ load strategy as
well as a simple _PendingRequests_ strategy for latency-agnostic
applications.

The `Instrument` trait is used by both of these strategies to track
in-flight requests without knowing protocol details. It is expected that
protocol-specific crates will provide, for instance, HTTP
time-to-first-byte latency strategies.

A default `NoInstrument` implementation tracks the a request until its
response future is satisfied.

This crate should only be published once tower-balance is published.


tower-load/src/lib.rs
```diff
@@ -0,0 +1,28 @@
...
+    constant::Constant,
+    instrument::{Instrument, InstrumentFuture, NoInstrument},
+    peak_ewma::{PeakEwma, PeakEwmaDiscover},
+    pending_requests::{PendingRequests, PendingRequestsDiscover},
+};
+
+/// Exposes a load metric.
+pub trait Load {
+    /// A comparable load metric. Lesser values are "preferable" to greater values.
+    type Metric: PartialOrd + std::fmt::Debug;
```
carllerche: You may want to consider removing the `Debug` bound from the core trait.

Issue #286, opened 2019-05-28
tower-balance: Observations from testing


olix0r:
We've recently been running tests on the Linkerd proxy that exercise the load
balancer in larger clusters (of 30+ endpoints). At the same time, I've been
exploring the existing endpoint-weighting scheme.

In doing so, I've realized that the balancer is currently O(n), though it is
intended to be effectively O(1). Furthermore, the existing weighting scheme
is complex to instrument in practice, and is of questionable value in its
current form.

All of this leads to me believe that we should drastically simplify the
balancer:

1. Do not attempt to unify (Weighted) P2C and Round-Robin under one
   implementation. Each strategy benefits from being able to use its own data
   structure. For now, I propose that we simply drop the Round-Robin logic. It
   can easily be added later if it's desirable.
2. The balancer cannot be responsible for driving the readiness of all of its
   constituent services. The P2C balancer is intended to sample _two_
   endpoints. In the current implementation, we _always_ poll _all_ unready
   inner services, which leads to poor behavior as the balancer scales.
   Something like #283 is necessary to relax the balancer's readiness-polling
   guarantees.
3. We initially decided that all endpoint-service errors should be treated as fatal
   to the balancer. It now seems more appropriate to let the balancer handle these
   failures by dropping the failed service from the balancer. 
4. The `Load` and `Instrument` traits are not balancer-specific and should
   become more generally useful abstractions in a dedicated crate.
5. The balancer should expose a Layer that layers over inner layers that
   produce `Discover`-typed results.
6. The `Pool` implementation is factored inconveniently, especially in light of how
   tower-layer has evolved: it requires direct access to a balancer implementation,
   accessing its `discover` field directly. I think that it should probably
   be implemented as a `Discover` proxy-type that is constructed with a
   `Watch<Load>`. The pool doesn't rely on any specific balancer behavior, so it
   shouldn't dictate use of a specific balancer implementation.

I have a series of changes that I would like to pursue to this end:
- https://github.com/tower-rs/tower/pull/283 Enables endpoint stacks to be driven to readiness without being actively polled, i.e. by the balancer.
- https://github.com/tower-rs/tower/pull/285 extracts the Load & Instrument traits into a dedicated crates; and removes the current `Weight` implementation (which is not really what we want).
- I have another (followup) branch that removes the `choose` trait/module, leaving only a `P2CBalance` implementation.
- I could use some help figuring out the path forward for Pool.

carllerche:
[...]

PR #288, opened 2019-05-30
balance: Specialize the balancer for P2C

olix0r:
As described in #286, `Balance` had a few problems:
- it is responsible for driving all inner services to readiness, making
  its `poll_ready` O(n) and not O(1);
- the `choose` abstraction was a hinderance. If a round-robin balancer
  is needed it can be implemented separately without much duplicate
  code; and
- endpoint errors were considered fatal to the balancer.

This changes replaces `Balance` with `p2c::Balance` and removes the
`choose` module.

Endpoint service failures now cause the service to be removed from the
balancer gracefully.

Endpoint selection is now effectively constant time, though it biases
for availability in the case when random selection does not yield an
available endpoint.

`tower-test` had to be updated so that a mocked service could fail after
advertising readiness.


tower-balance/src/layer.rs
```diff
@@ -0,0 +1,49 @@
...
+        let rng = SmallRng::from_rng(rng)?;
+        Ok(Self {
+            rng,
+            _marker: PhantomData,
+        })
+    }
+}
+
+impl<D: Discover> Layer<D> for P2CBalanceLayer<D> {
+    type Service = P2CBalance<D>;
```
olix0r: bleh, this really should build a MakeP2CBalance or whatever...

tower-balance/src/lib.rs
```diff
@@ -1,337 +1,287 @@
...
-            None => return None,
-            Some((_, svc)) => match svc.poll_ready() {
-                Ok(Async::Ready(())) => return Some(Ok(Async::Ready(()))),
-                Err(e) => return Some(Err(e)),
-                Ok(Async::NotReady) => {}
-            },
+        match self.endpoints.len() {
+            0 => Ok(Async::NotReady),
+            1 => {
+                // If there's only one endpoint, ignore its but require that it
```
hawkw: Nit: I think there's a word missing here --- "[...] ignore its but require [...]"?

tower-balance/src/lib.rs
```diff
@@ -1,337 +1,287 @@
...
-    type Future = ResponseFuture<Svc::Future>;
+    type Response = <D::Service as Service<Request>>::Response;
+    type Error = error::Error;
+    type Future =
+        future::MapErr<<D::Service as Service<Request>>::Future, fn(Svc::Error) -> error::Error>;
 
     /// Prepares the balancer to process a request.
     ///
-    /// When `Async::Ready` is returned, `chosen_ready_index` is set with a valid index
+    /// When `Async::Ready` is returned, `chosen` is set with a valid index
```
hawkw: I think `chosen` was renamed to `ready_index`?

tower-balance/src/test.rs
```diff
@@ -1,136 +1,142 @@
...
 use crate::*;
 
-type Error = Box<dyn std::error::Error + Send + Sync>;
+//type Error = Box<dyn std::error::Error + Send + Sync>;
 
-struct ReluctantDisco(VecDeque<Change<usize, ReluctantService>>);
-
-struct ReluctantService {
-    polls_until_ready: usize,
+macro_rules! assert_ready {
```
hawkw: TIOLI: I think the `tokio-test` crate also provides similar macros we could use here?

tower-balance/src/lib.rs
```diff
@@ -1,337 +1,287 @@
...
-}
-
-impl<D, C> Balance<D, C>
-where
-    D: Discover,
-    C: Choose<D::Key, D::Service>,
-{
-    /// Creates a new balancer.
-    pub fn new(discover: D, choose: C) -> Self {
+    pub fn new(discover: D, rng: SmallRng) -> Self {
```
hawkw: This is quite nitpicky, but I'm not sure whether it's more idiomatic for this to be `new`, or to make this `new_with_rng` and call `from_entropy` `new`?

tower-balance/src/lib.rs
```diff
@@ -1,337 +1,287 @@
...
-                    };
-                    // XXX is it safe to just drop the Service? Or do we need some sort of
-                    // graceful teardown?
-                    // TODO: poll_close
+        loop {
+            match try_ready!(self.discover.poll().map_err(|e| error::Discover(e.into()))) {
+                Change::Insert(key, svc) => drop(self.endpoints.insert(key, svc)),
+                Change::Remove(rm_key) => {
+                    // Update the ready index to account for reordering of endpoints.
+                    let orig_sz = self.endpoints.len();
```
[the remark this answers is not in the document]
olix0r: No, because the if-let clause modifies the size

tower-balance/src/lib.rs
olix0r: This isn't meant to be primary entry-point (the layer is), so i ended up with this. In the proxy, for instance, we weren't reusing the RNGs and were creating a new seed for each balancer (because the `new` hid this detail).

tower-balance/src/test.rs
olix0r: citation needed

hawkw: https://github.com/tokio-rs/tokio/blob/84d5a7f5a0660dbe1e2be0e831e715f13cb59a2b/tokio-test/src/macros.rs#L3-L55

hawkw: you'd have to use `assert_ready!(svc.poll_ready())` rather than just `assert_ready!(svc)` though. also, not sure if it's worth it (it probably isn't).

olix0r: Oooooh tokio-test, not tower-test!

tower-balance/src/lib.rs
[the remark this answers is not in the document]
hawkw: ah, yeah, missed that. never mind!

olix0r:
Per offline discussion, @carllerche suggested some potential improvements that I have written up in https://github.com/tower-rs/tower/issues/292. We're going to merge this for the time being and revisit #292 before balance is released to crates.

jonhoo:
I just realized that this _does_ introduce a requirement for inner services to implement `Load`, which seems potentially unfortunate. With round-robin, that requirement is relaxed.

olix0r:
@jonhoo yeah, that is a notable change... I think it would be pretty easy to introduce a `rr:Balance` that does this (and could actually avoid "incorrectness" in the face of discovery updates -- the prior implementation did not attempt to preserve order).

olix0r:
@jonhoo though, note the `Constant` load type can be used to paper over this wrinkle

PR #293, opened 2019-06-12
balance: Only balance over ready endpoints

### LINKED ISSUE #292: balance: Investigate FuturesUnordered-style polling for inner services
In #286, we decided that the balancer must not be responsible for driving all of its inner services to readiness. The reality is probably a bit more nuanced...

Ideally, the balancer would maintain a `FuturesUnordered` style structure internally containing non-ready service; and we'd be able to know exactly which inner services should be polled for readiness (and then performing p2c over only ready services).

We can't just use FuturesUndordered directly, as we have no means to force a given readiness future to fail when the service is removed from discovery. This is necessary so that the FuturesUnordered doesn't leak services that never become ready (i.e. due to reconnect logic) but are administratively removed (i.e. via tower-discover).

olix0r:
In 03ec4aa, the balancer was changed to make a quick endpoint decision.
This, however, means that the balancer can return NotReady when it does
in fact have a ready endpoint. Furthermore, testing shows that while
this approach does have about 10% beter CPU utilization, it has about
25% worse throughput when communicating with a busy service.

This changes the balancer to segregate unready endpoints, only
performing p2c over ready endpoints. Unready endpoints are tracked with
a FuturesUnordered that supports premature eviction via oneshots.

The main downside of this change is that the Balancer must become
generic over the Request type.

Fixes #292

### DIFF
--- a/tower-balance/Cargo.toml
+++ b/tower-balance/Cargo.toml
@@ -29,7 +29,6 @@ log = "0.4.1"
 rand = "0.6.5"
 tokio-sync = "0.1.3"
 tokio-timer = "0.2.4"
-tower = { version = "0.1.0", path = "../tower" }
 tower-discover = "0.1.0"
 tower-layer = "0.1.0"
 tower-load = { version = "0.1.0", path = "../tower-load" }
@@ -43,7 +42,7 @@ hdrsample = "6.0"
 quickcheck = { version = "0.6", default-features = false }
 tokio = "0.1.7"
 tokio-executor = "0.1.2"
-tower = { version = "0.1", path = "../tower" }
-tower-buffer = { version = "0.1", path = "../tower-buffer" }
-tower-limit = { version = "0.1", path = "../tower-limit" }
-tower-test = { version = "0.1.0", path = "../tower-test" }
+tower = { version = "*", path = "../tower" }
+tower-buffer = { version = "*", path = "../tower-buffer" }
+tower-limit = { version = "*", path = "../tower-limit" }
+tower-test = { version = "*", path = "../tower-test" }
--- a/tower-balance/Cargo.toml
+++ b/tower-balance/Cargo.toml
@@ -27,7 +27,9 @@ futures = "0.1.26"
 indexmap = "1.0.2"
 log = "0.4.1"
 rand = "0.6.5"
+tokio-sync = "0.1.3"
 tokio-timer = "0.2.4"
+tower = { version = "0.1.0", path = "../tower" }
 tower-discover = "0.1.0"
 tower-layer = "0.1.0"
 tower-load = { version = "0.1.0", path = "../tower-load" }
--- a/tower-balance/src/p2c/make.rs
+++ b/tower-balance/src/p2c/make.rs
@@ -1,25 +1,32 @@
 use super::Balance;
 use futures::{try_ready, Future, Poll};
 use rand::{rngs::SmallRng, FromEntropy};
+use std::marker::PhantomData;
 use tower_discover::Discover;
 use tower_service::Service;
 
 /// Makes `Balancer`s given an inner service that makes `Discover`s.
 #[derive(Clone, Debug)]
-pub struct BalanceMake<S> {
+pub struct BalanceMake<S, Req> {
     inner: S,
     rng: SmallRng,
+    _marker: PhantomData<fn(Req)>,
 }
 
 /// Makes a balancer instance.
-pub struct MakeFuture<F> {
+pub struct MakeFuture<F, Req> {
     inner: F,
     rng: SmallRng,
+    _marker: PhantomData<fn(Req)>,
 }
 
-impl<S> BalanceMake<S> {
+impl<S, Req> BalanceMake<S, Req> {
     pub(crate) fn new(inner: S, rng: SmallRng) -> Self {
-        Self { inner, rng }
+        Self {
+            inner,
+            rng,
+            _marker: PhantomData,
+        }
     }
 
     /// Initializes a P2C load balancer from the OS's entropy source.
@@ -28,14 +35,15 @@ impl<S> BalanceMake<S> {
     }
 }
 
-impl<S, Target> Service<Target> for BalanceMake<S>
+impl<S, Target, Req> Service<Target> for BalanceMake<S, Req>
 where
     S: Service<Target>,
     S::Response: Discover,
+    <S::Response as Discover>::Service: Service<Req>,
 {
-    type Response = Balance<S::Response>;
+    type Response = Balance<S::Response, Req>;
     type Error = S::Error;
-    type Future = MakeFuture<S::Future>;
+    type Future = MakeFuture<S::Future, Req>;
 
     fn poll_ready(&mut self) -> Poll<(), Self::Error> {
         self.inner.poll_ready()
@@ -45,16 +53,18 @@ where
         MakeFuture {
             inner: self.inner.call(target),
             rng: self.rng.clone(),
+            _marker: PhantomData,
         }
     }
 }
 
-impl<F> Future for MakeFuture<F>
+impl<F, Req> Future for MakeFuture<F, Req>
 where
     F: Future,
     F::Item: Discover,
+    <F::Item as Discover>::Service: Service<Req>,
 {
-    type Item = Balance<F::Item>;
+    type Item = Balance<F::Item, Req>;
     type Error = F::Error;
 
     fn poll(&mut self) -> Poll<Self::Item, Self::Error> {
--- a/tower-balance/src/p2c/service.rs
+++ b/tower-balance/src/p2c/service.rs
@@ -26,9 +26,9 @@ use tower_util::Ready;
 pub struct Balance<D: Discover, Req> {
     discover: D,
 
-    ready: IndexMap<D::Key, D::Service>,
+    ready_services: IndexMap<D::Key, D::Service>,
 
-    unready: stream::FuturesUnordered<CancelReady<D::Key, D::Service, Req>>,
+    unready_services: stream::FuturesUnordered<UnreadyService<D::Key, D::Service, Req>>,
     cancelations: IndexMap<D::Key, oneshot::Sender<()>>,
 
     /// Holds an index into `endpoints`, indicating the service that has been
@@ -38,8 +38,11 @@ pub struct Balance<D: Discover, Req> {
     rng: SmallRng,
 }
 
+/// A Future that becomes satisfied when an `S`-typed service is ready.
+///
+/// May fail due to cancelation, i.e. if the service is removed from discovery.
 #[derive(Debug)]
-struct CancelReady<K, S, Req> {
+struct UnreadyService<K, S, Req> {
     key: Option<K>,
     cancel: oneshot::Receiver<()>,
     ready: tower_util::Ready<S, Req>,
@@ -60,9 +63,9 @@ where
         Self {
             rng,
             discover,
-            ready: IndexMap::default(),
+            ready_services: IndexMap::default(),
             cancelations: IndexMap::default(),
-            unready: stream::FuturesUnordered::new(),
+            unready_services: stream::FuturesUnordered::new(),
             next_ready_index: None,
         }
     }
@@ -74,7 +77,7 @@ where
 
     /// Returns the number of endpoints currently tracked by the balancer.
     pub fn len(&self) -> usize {
-        self.ready.len() + self.unready.len()
+        self.ready_services.len() + self.unready_services.len()
     }
 
     // XXX `pool::Pool` requires direct access to this... Not ideal.
@@ -115,7 +118,7 @@ where
     fn push_unready(&mut self, key: D::Key, svc: D::Service) {
         let (tx, rx) = oneshot::channel();
         self.cancelations.insert(key.clone(), tx);
-        self.unready.push(CancelReady {
+        self.unready_services.push(UnreadyService {
             key: Some(key),
             ready: Ready::new(svc),
             cancel: rx,
@@ -124,10 +127,10 @@ where
 
     fn evict(&mut self, key: &D::Key) {
         // Update the ready index to account for reordering of ready.
-        if let Some((idx, _, _)) = self.ready.swap_remove_full(key) {
+        if let Some((idx, _, _)) = self.ready_services.swap_remove_full(key) {
             self.next_ready_index = self
                 .next_ready_index
-                .and_then(|i| Self::repair_index(i, idx, self.ready.len()));
+                .and_then(|i| Self::repair_index(i, idx, self.ready_services.len()));
             debug_assert!(!self.cancelations.contains_key(key));
         } else if let Some(cancel) = self.cancelations.remove(key) {
             let _ = cancel.send(());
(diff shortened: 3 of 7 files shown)


tower-balance/src/p2c/service.rs
```diff
@@ -21,26 +23,47 @@ use tower_service::Service;
...
     /// chosen to dispatch the next request.
-    ready_index: Option<usize>,
+    next_ready_index: Option<usize>,
 
     rng: SmallRng,
 }
 
-impl<D: Discover> Balance<D> {
+#[derive(Debug)]
+struct CancelReady<K, S, Req> {
```
carllerche: This name confused me. If I am reading this correctly, this struct waits on a service to become ready, and is cancellable. The name made me think the action taken by this struct was to cancel readiness.

carllerche:
Also, ccing @jonhoo

tower-balance/Cargo.toml
```diff
@@ -27,7 +27,9 @@ futures = "0.1.26"
 indexmap = "1.0.2"
 log = "0.4.1"
 rand = "0.6.5"
+tokio-sync = "0.1.3"
 tokio-timer = "0.2.4"
+tower = { version = "0.1.0", path = "../tower" }
```
jonhoo: Doesn't this result in a circular dependency?

tower-balance/src/p2c/make.rs
```diff
@@ -1,25 +1,32 @@
...
 use tower_discover::Discover;
 use tower_service::Service;
 
 /// Makes `Balancer`s given an inner service that makes `Discover`s.
 #[derive(Clone, Debug)]
-pub struct BalanceMake<S> {
+pub struct BalanceMake<S, Req> {
     inner: S,
     rng: SmallRng,
+    _marker: PhantomData<fn(Req)>,
```
jonhoo: Why `fn(Req)` over just `Req`?

tower-balance/src/p2c/service.rs
```diff
@@ -51,33 +74,81 @@ impl<D: Discover> Balance<D> {
...
-                Change::Insert(key, svc) => drop(self.endpoints.insert(key, svc)),
-                Change::Remove(rm_key) => {
-                    // Update the ready index to account for reordering of endpoints.
-                    if let Some((rm_idx, _, _)) = self.endpoints.swap_remove_full(&rm_key) {
-                        self.ready_index = self
-                            .ready_index
-                            .and_then(|i| Self::repair_index(i, rm_idx, self.endpoints.len()));
-                    }
+                Change::Remove(key) => {
+                    trace!("remove");
```
jonhoo: Did you mean to include this and the `trace!` below?

tower-balance/src/p2c/service.rs
```diff
@@ -21,26 +23,47 @@ use tower_service::Service;
...
+
+enum Error<E> {
+    Inner(E),
+    Canceled,
+}
+
+impl<D, Req> Balance<D, Req>
+where
+    D: Discover,
+    D::Service: Service<Req>,
```
jonhoo: I don't think you need this bound here?

tower-balance/src/p2c/service.rs
```diff
@@ -51,33 +74,81 @@ impl<D: Discover> Balance<D> {
...
+            let _ = cancel.send(());
+        }
+    }
+
+    fn poll_unready(&mut self) {
+        loop {
+            match self.unready.poll() {
+                Ok(Async::NotReady) | Ok(Async::Ready(None)) => return,
+                Ok(Async::Ready(Some((key, svc)))) => {
+                    trace!("endpoint ready");
```
jonhoo: Same here -- were this `trace!` and the `info!` below intentionally included?

tower-balance/src/p2c/service.rs
```diff
@@ -51,33 +74,81 @@ impl<D: Discover> Balance<D> {
...
+                    let _cancel = self.cancelations.remove(&key);
+                    debug_assert!(_cancel.is_some(), "missing cancelation");
+                    self.ready.insert(key, svc);
+                }
+                Err((key, Error::Canceled)) => debug_assert!(!self.cancelations.contains_key(&key)),
+                Err((key, Error::Inner(e))) => {
+                    info!("dropping failed endpoint: {}", e.into());
```
```suggestion
                    debug!("dropping failed endpoint: {}", e.into());
```

tower-balance/src/p2c/service.rs
jonhoo: Related to this, I'd like to see some docs on `CancelReady` to explain why it's needed and how it's used + maintained.

tower-balance/src/p2c/service.rs
```diff
@@ -106,120 +177,106 @@ impl<D: Discover> Balance<D> {
...
+                let (key, svc) = self
+                    .ready
+                    .swap_remove_index(index)
+                    .expect("invalid ready index");
+                self.push_unready(key, svc);
+                None
+            }
+            Err(e) => {
+                // failed, so drop it.
+                info!("evicting failed endpoint: {}", e.into());
```
```suggestion
                debug!("evicting failed endpoint: {:?}", e.into());
```

tower-balance/src/p2c/service.rs
```diff
@@ -106,120 +177,106 @@ impl<D: Discover> Balance<D> {
...
-        Svc: Service<Request> + Load,
-        Svc::Error: Into<error::Error>,
-    {
-        let (_, svc) = self.endpoints.get_index_mut(index).expect("invalid index");
-        try_ready!(svc.poll_ready());
-        Ok(Async::Ready(svc.load()))
+    /// Accesses a ready endpoint by index and returns its current load.
+    fn poll_ready_index_load(&mut self, index: usize) -> Option<<D::Service as Load>::Metric> {
+        let (_, svc) = self.ready.get_index_mut(index).expect("invalid index");
+        let load = match svc.poll_ready() {
```
jonhoo: I wonder if it's even necessary to `poll_ready` here. Doesn't the `Service` only end up in `self.ready` if it has already returned `Ready` from `poll_ready`?

tower-balance/src/p2c/service.rs
```diff
@@ -230,45 +287,58 @@ where
...
-                Err(e) => {
-                    drop(self.endpoints.swap_remove_index(index));
-                    info!("evicting failed endpoint: {}", e.into());
-                }
-            }
-
-            self.ready_index = None;
-        }
+        self.poll_unready();
+        debug!("ready={}; unready={}", self.ready.len(), self.unready.len());
```
```suggestion
        trace!("ready={}; unready={}", self.ready.len(), self.unready.len());
```

tower-balance/src/p2c/service.rs
```diff
@@ -230,45 +287,58 @@ where
...
-                Ok(Async::NotReady) => {}
-                Err(e) => {
-                    drop(self.endpoints.swap_remove_index(index));
-                    info!("evicting failed endpoint: {}", e.into());
-                }
-            }
-
-            self.ready_index = None;
-        }
+        self.poll_unready();
```
jonhoo: I wonder whether it'd be better to only `poll_unready` if there aren't enough things in `self.ready`?

tower-balance/src/p2c/service.rs
```diff
@@ -230,45 +287,58 @@ where
...
-            n => n / 2,
-        };
-        for _ in 0..tries {
-            if let Async::Ready(idx) = self.poll_ready_index().map_err(Into::into)? {
-                trace!("ready: {:?}", idx);
-                self.ready_index = Some(idx);
+        if let Some(index) = self.next_ready_index {
+            trace!("preselected ready_index={}", index);
+            debug_assert!(index < self.ready.len());
+            // Ensure the selected endpoint is still ready.
```
jonhoo: Are we expecting endpoints to _stop_ being ready after they have indicated that they are ready? /cc @carllerche

tower-balance/src/p2c/service.rs
```diff
@@ -230,45 +287,58 @@ where
...
-        let tries = match self.endpoints.len() {
-            0 => return Ok(Async::NotReady),
-            1 => 1,
-            n => n / 2,
-        };
-        for _ in 0..tries {
-            if let Async::Ready(idx) = self.poll_ready_index().map_err(Into::into)? {
-                trace!("ready: {:?}", idx);
-                self.ready_index = Some(idx);
+        if let Some(index) = self.next_ready_index {
```
jonhoo: I'm not sure I follow this. Why would `self.next_ready_index` every be `Some` here? Don't we always immediately use an endpoint when we select one?

tower-balance/src/p2c/service.rs
jonhoo: Oh, this is for cancellation, isn't it? I think it'd be useful to document somewhere exactly how this flow works. In particular, this code wouldn't be sufficient _in general_ if services could stop being ready, but it is sufficient _in this case_ because we know that only we (`&mut self`) could have cancelled the `Service`.

tower-balance/src/p2c/service.rs
jonhoo: I commented on this below, but I now realize (I think) that this is due to cancellations. Explaining that in a comment would be helpful.

tower-balance/src/p2c/service.rs
jonhoo: Maybe `RemovableService` is a better name?

jonhoo:
Sorry it took me so long to get to this @olix0r! This looks like an _excellent_ change, though I do have some questions I'd like to resolve before merging. Most of them are around adding comments for some of the more complex interactions (and to check my own understanding).

tower-balance/src/p2c/service.rs
olix0r: @jonhoo It's possible, for instance if a failure detector or circuit breaker kicks in, for a service to go from ready to notready or ready to error. We take a belt-and-suspenders approach to avoid on the assumption that redundant polling will be cheap.

tower-balance/src/p2c/service.rs
olix0r: We have no guarantees that poll will only be called once before call.

jonhoo: Oh, this is me being stupid. `poll_ready` will be called first, and it is what sets `next_ready_index`, which is used by `call`. Never mind me :sweat_smile:

tower-balance/src/p2c/service.rs
jonhoo: Right, I guess my point is more that if a service can go from ready to not ready at any time, then no amount of calling `poll_ready` is sufficient. The service could always become not ready _just_ after you last call `poll_ready`. At some point, you have to leave it to `call` to actually give out the error. From this I think it follows that we are here specifically trying to deal with the case of evicted services, because for those we _know_ that this check is sufficient. This in turn makes me wonder whether we could have a fast-path here where we only do this second `poll_ready` if there has been at least one eviction?

tower-balance/Cargo.toml
olix0r: Cargo culted from tower-buffer, tower-test, & tower-util...  I'm happy to do whatever is the best practice; though, for the time being, `tower` does not export tower-balance...

carllerche: `tower` can be in the `dev-dependencies` section safely, but it probably should not depend directly on `tower`.

carllerche: In general, I want to update the dev-dep on "facade" crates to use a `*` version dependency. Without that, there will be conflicts when we try to issue breaking change releases...

tower-balance/src/p2c/make.rs
olix0r: We do this here (and elsewhere) so that `Req` doesn't get the `Sync/Send/'static` requirements of the Service.  I.e. `Req` need not be `Send` just because `Service` is, etc.

I'm pretty sure I picked up this habit from @seanmonstar...

jonhoo:
@olix0r depending on how busy you are, I can also take on pushing this over the finish line as I'd like to have it for some work-related stuff :)

olix0r:
@jonhoo I should have a chance today to push another iteration. Thanks, though!

tower-balance/src/p2c/service.rs
olix0r: Landed on `UnreadyService` to be consistent with the (updated) field name `unready_services`.

tower-balance/src/p2c/service.rs
olix0r: It's needed so that `UnreadyService<..>: Future`, which is required by `FuturesUnordered`

tower-balance/src/p2c/service.rs
olix0r: This adds some complexities that I'd prefer to defer for now. Specifically, this means that new instances added to the load balancer won't be considered until all existing nodes are exhausted. I think this would be surprising in practice and difficult for users to debug.

Instead, we'll probably want to consider an alternate _aperture_ load balancer at some point in the future, which works more as you describe -- fitting the number of nodes being considered to the load requirements.

As for cost of polling unready, in practice I think the endpoint stack should use a spawn-ready layer so that the balancer isn't responsible for driving all endpoints to readiness, and is only notified as they are fully ready.

jonhoo: Yeah, that's a good point. In fact, it's similar to exactly what I ended up doing in [`tokio-tower`](https://github.com/tower-rs/tokio-tower) -- the service is spawned and provides a single-element queue of requests precisely so that the service is driven elsewhere.

tower-balance/src/p2c/service.rs
olix0r: While it's always technically true that no amount of polling can _guarantee_ readiness, I do think it's important that the target service returns ready immediately before _call_ is invoked. so that the balancer task is not interrupted between the readiness and the call.

If we're concerned about the cost of this, let's craft some benchmarks and optimize in followups.

olix0r:
@jonhoo I do take your general point about this polling perhaps excessively. I do think  it's important for us to eagerly poll a selected endpoint immediately before a request is dispatched to it, but i do think we can accomplish this with a slightly more relaxed strategy. Let me try something...

jonhoo:
I'd also be okay with merging this as-is now and then do optimizations in a follow-up PR. Up to you :)

olix0r:
@jonhoo PTAL; that change allowed us to remove some of the uglier code in there (repairing indexes).

tower-balance/Cargo.toml
```diff
@@ -41,7 +42,7 @@ hdrsample = "6.0"
...
 tokio = "0.1.7"
 tokio-executor = "0.1.2"
-tower = { version = "0.1", path = "../tower" }
-tower-buffer = { version = "0.1", path = "../tower-buffer" }
-tower-limit = { version = "0.1", path = "../tower-limit" }
-tower-test = { version = "0.1.0", path = "../tower-test" }
+tower = { version = "*", path = "../tower" }
+tower-buffer = { version = "*", path = "../tower-buffer" }
+tower-limit = { version = "*", path = "../tower-limit" }
+tower-test = { version = "*", path = "../tower-test" }
```
olix0r: @carllerche like this?

carllerche: Yes, that looks like what I had in mind.

Full disclosure: I haven't actually updated any other project ot do this... so this will be the experiment!

PR #295, opened 2019-06-28
Update to std::future::Future

LucioFranco:


leaxoy:
How about change generic type `Request` to trait associated type

davidbarsky:
@leaxoy Tower used to treat request as an associated type, but this was changed here a few months back: https://github.com/tower-rs/tower/pull/109

leaxoy:
@davidbarsky thanks

carllerche:
What are the remaining steps here?

LucioFranco:
@carllerche ok sorry for late response, currently only timeout and service are updated. Im keeping this light for right now as my goal is to get grpc updated then merge this with just the service and timeout changes so we can update the other layers in parallel.

PR #298, opened 2019-07-05
Early push to bring tracing into tower

jonhoo:


tower-balance/Cargo.toml
```diff
@@ -22,10 +22,14 @@ categories = ["asynchronous", "network-programming"]
 edition = "2018"
 publish = false
 
+[features]
+log = ["tracing/log"]
+default = []
```
hawkw: IMO, the `log` feature should _probably_ be on by default, given that most users currently expect `tower` to emit `log` records, and removing it will break them...

tower-balance/examples/demo.rs
```diff
@@ -38,7 +37,12 @@ struct Summary {
 }
 
 fn main() {
-    env_logger::init();
+    tracing::subscriber::set_global_default(
+        tracing_fmt::FmtSubscriber::builder()
+            .on_event(tracing_fmt::default::Format::default())
```
hawkw: I think this is unnecessary, you can just say
```rust
tracing_fmt::FmtSubscriber::builder().finish()
```
or even
```rust
tracing_fmt::FmtSubscriber::default()
```
(i think)

tower-buffer/src/service.rs
```diff
@@ -87,13 +87,21 @@ where
     }
 
     fn call(&mut self, request: Request) -> Self::Future {
+        let span = tracing::Span::current();
+        let _guard = span.enter();
```
hawkw: TIOLI: Is it expected that events will occur in the `oneshot::channel` constructor? If not, we can probably just use the span as the parent for the `trace!` call on line 99, and then include it in the message without cloning it.

tower-buffer/src/worker.rs
```diff
@@ -157,30 +165,43 @@ where
...
                     if let Some(ref failed) = self.failed {
+                        tracing::trace!("notifying about worker failure");
                         let _ = msg.tx.send(Err(failed.clone()));
                         continue;
                     }
 
                     // Wait for the service to be ready
+                    tracing::trace!(
+                        { resumed = !first },
+                        "worker received request; waiting for service readiness"
```
hawkw: Nit/TIOLI: IMO this might read better as 
```rust
                    tracing::trace!(
                        resumed = !first,
                        message = "worker received request; waiting for service readiness",
                    );
```
but, up to you.

tower-buffer/src/worker.rs
```diff
@@ -157,30 +165,43 @@ where
...
                             //
                             // An error means the request had been canceled in-between
                             // our calls, the response future will just be dropped.
+                            tracing::trace!("returning response future");
                             let _ = msg.tx.send(Ok(response));
                         }
                         Ok(Async::NotReady) => {
+                            tracing::trace!("service not ready; delay");
                             // Put out current message back in its slot.
+                            drop(_guard);
```
hawkw: The guard should be dropped when we return two lines later; this drop doesn't seem strictly necessary...

tower-buffer/src/worker.rs
jonhoo: My one concern with this change is that it hardcodes an assumption that the string message is equivalent to a field named `message`. Not sure that that's something `tracing` guarantees?

tower-buffer/src/worker.rs
jonhoo: It's necessary because the guard borrows `msg.span`, and we're trying to move `msg` on the next line.

tower-buffer/src/worker.rs
hawkw: > My one concern with this change is that it hardcodes an assumption that the string message is equivalent to a field named `message`. Not sure that that's something `tracing` guarantees?

It should be guaranteed to have the same behaviour as using the `log`-like syntax; the macros will desugar that to a field named `message` with a `format_args!` value identically to the `key=value` syntax above.

jonhoo: Is that documented anywhere?

hawkw: if you weren't aware of it, it is probably under-documented >.<

jonhoo: So, to clarify, I was aware that the current implementation exposes it as a field called `message`. I did not know that `tracing` _guarantees_ that this will be the case (i.e., it's a part of the library's stable API).

jonhoo:
I don't think there's a strong reason not to land something like this soon, and then we can iterate on it going forward.

hawkw:
I would love to be using this in linkerd, fwiw.

tower-buffer/src/worker.rs
jonhoo: @hawkw I think I'm only waiting on an answer to ^ before this is ready to merge.

tower-buffer/src/worker.rs
hawkw: whoops, missed that!

tower-buffer/src/worker.rs
hawkw: I think we've committed to `message` being part of the stable API at this point --- if we ever change the handling of textual messages, we'll ensure that it is at least back-compatible with subscribers that expect a message field. This should, perhaps, be documented in `tracing`.

tower-buffer/src/service.rs
```diff
@@ -93,7 +93,9 @@ where
         // outside of task context.
         let (tx, rx) = oneshot::channel();
 
-        match self.tx.try_send(Message { request, tx }) {
+        let span = tracing::Span::current();
```
hawkw: TIOLI: Might be nice to have a comment here for future readers noting that we capture the current span here to propagate it to the buffer worker?

tower-buffer/src/worker.rs
```diff
@@ -157,30 +165,43 @@ where
 
         loop {
             match try_ready!(self.poll_next_msg()) {
-                Some(msg) => {
+                Some((msg, first)) => {
+                    let _guard = msg.span.enter();
                     if let Some(ref failed) = self.failed {
+                        tracing::trace!("notifying about worker failure");
```
hawkw: Nit: notifying _what_ about worker failure? Maybe
```suggestion
                        tracing::trace!("notifying caller about worker failure");
```
or something?

tower-buffer/src/worker.rs
```diff
@@ -157,30 +165,43 @@ where
...
                     }
 
                     // Wait for the service to be ready
+                    tracing::trace!(
+                        resumed = !first,
+                        message = "worker received request; waiting for service readiness"
+                    );
                     match self.service.poll_ready() {
                         Ok(Async::Ready(())) => {
+                            tracing::debug!("processing request");
```
hawkw: Nit/TIOLI: I might consider changing this to something like:
```suggestion
                            tracing::debug!(service.ready = true, message = "processing request");
```
That way, if we add a similar `service.ready` boolean field to the event that we record when the service isn't ready, a hypothetical subscriber could track the state of the service.

tower-buffer/src/worker.rs
```diff
@@ -157,30 +165,43 @@ where
...
 
                             // Send the response future back to the sender.
                             //
                             // An error means the request had been canceled in-between
                             // our calls, the response future will just be dropped.
+                            tracing::trace!("returning response future");
                             let _ = msg.tx.send(Ok(response));
                         }
                         Ok(Async::NotReady) => {
+                            tracing::trace!("service not ready; delay");
```
hawkw: As I mentioned above, what if this was
```suggestion
                            tracing::trace!(service.ready = false, message = "delay");
```

tower-balance/src/pool.rs
```diff
@@ -71,14 +73,17 @@ where
                     return Ok(Async::NotReady);
                 }
 
+                tracing::trace!("decided to add service to loaded pool");
```
hawkw: TIOLI: should we also record the number of services here?
```suggestion
                tracing::trace!(pool.services = self.services, message = "decided to add service to loaded pool");
```

tower-balance/src/pool.rs
```diff
@@ -71,14 +73,17 @@ where
...
                 try_ready!(self.maker.poll_ready());
+                tracing::trace!("making new service");
                 // TODO: it'd be great if we could avoid the clone here and use, say, &Target
                 self.making = Some(self.maker.make_service(self.target.clone()));
             }
         }
 
         if let Some(mut fut) = self.making.take() {
             if let Async::Ready(s) = fut.poll()? {
+                tracing::trace!("finished creating new service");
```
hawkw: Maybe we should record the number of services here as well?
```suggestion
                tracing::trace!(pool.services = self.services, message = "finished creating new service");
```

tower-balance/src/pool.rs
```diff
@@ -95,6 +100,7 @@ where
             Level::Normal => Ok(Async::NotReady),
             Level::Low if self.services == 1 => Ok(Async::NotReady),
             Level::Low => {
+                tracing::trace!("removing service for over-provisioned pool");
```
hawkw: ...and here?
```suggestion
                tracing::trace!(pool.services = self.services - 1, message = "removing service for over-provisioned pool");
```

(instead of `self.services - 1` we could also just move this down to after we actually remove the service)

tower-balance/src/pool.rs
```diff
@@ -284,13 +290,19 @@ where
 
             let discover = self.balance.discover_mut();
             if self.ewma < self.options.low {
+                if discover.load != Level::Low {
+                    tracing::trace!({ ewma = %self.ewma }, "pool is over-provisioned");
```
hawkw: TIOLI: what if all of these conditionally-recorded traces were replaced with a single
```rust
tracing::trace!(ewma = %self.ewma, %discover.load, %discover.services)
```
or something?

jonhoo: That would result in a _lot_ of mostly unnecessary traces (the ewma is changing for _every_ request, whereas the load estimate changes only very rarely). Not sure about this one.

hawkw: Ah, yeah, you're right. Never mind, let's not do that one.

tower-balance/src/p2c/service.rs
```diff
@@ -197,14 +195,7 @@ where
...
 
-                trace!(
-                    "load[{}]={:?}; load[{}]={:?} -> ready={}",
-                    aidx,
-                    aload,
-                    bidx,
-                    bload,
-                    ready
-                );
+                trace!({ a.idx = aidx, a.load = ?aload, b.idx = bidx, b.load = ?bload, ready = ?ready }, "choosing by load");
```
hawkw: Oh, one final thing, extremely minor: I notice the way we name the fields for indices in this file is kinda inconsistent: on line 174, we drop the `idx` from the variable names but refer to the repaired index as `i`, while here, we use `a.idx` and `b.idx`, and on line 288 we have a field called `index`. Maybe we should make these more consistent?

tower-balance/src/pool.rs
```diff
@@ -71,14 +73,23 @@ where
...
+                tracing::trace!("making new service");
                 // TODO: it'd be great if we could avoid the clone here and use, say, &Target
                 self.making = Some(self.maker.make_service(self.target.clone()));
             }
         }
 
         if let Some(mut fut) = self.making.take() {
             if let Async::Ready(s) = fut.poll()? {
+                tracing::trace!(
+                    pool.services = self.services,
```
hawkw: The event when deleting a service occurs _after_ the service is deleted and `self.services` is incremented, while this one occurs before. I think we should move this to after the increment so we report the number of services after creating the new service.

PR #300, opened 2019-07-09
Service combinator utilities

samscott89:
I mostly put this together because I wanted `map_err`. But once I started it seemed like a reasonable thing to extend out.

This basically implements super simple `Service` mapping, mostly with a one-to-one equivalency with the `futures` variants.

Open to thoughts/suggestions. And rationale for why this makes no sense :) Especially in the context of #214, and perhaps #272. 

To clarify, this is for simple futures-based combinators on services' responses. Not about combining services, nor modifying requests.

### DIFF
--- a/tower-util/src/combinator/or_else.rs
+++ b/tower-util/src/combinator/or_else.rs
@@ -1,0 +1,58 @@
+use super::*;
+
+/// `Service` analog to the `futures::future::OrElse` combinator.
+///
+/// TODO: Currently restricted to not modifying the error type,
+/// since we need `poll_ready` and `call` to have the same type.
+/// But it doesn't make sense to `or_else` the `poll_ready` type.
+pub struct OrElse<T, F> {
+    inner: T,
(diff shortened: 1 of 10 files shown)


tower-util/src/combinator/or_else.rs
```diff
@@ -0,0 +1,58 @@
+use super::*;
+
+/// `Service` analog to the `futures::future::OrElse` combinator.
+///
+/// TODO: Currently restricted to not modifying the error type,
+/// since we need `poll_ready` and `call` to have the same type.
+/// But it doesn't make sense to `or_else` the `poll_ready` type.
```
olix0r: Elsewhere in cases like, we tend to have the `Error` type be `Box<dyn std::errror::Error + Send + Sync>`, and then place requirements on the inner services like: `T::Error: Into<Box<dyn std::errror::Error + Send + Sync>>`. This tends to be more ergonomic.

PR #301, opened 2019-07-10
 Notify Pool when Services are dropped

jonhoo:
Prior to this change, when `Balance` dropped a failing service, `Pool` would not be notified of this fact. This meant that it never updated `.services`, and so it might not add a new backing `Service` (e.g., due to `max_services`) even though no working backing exist.

With this change, dropped services notify the `Pool` so that it knows to re-check its limits.

This PR also (finally) adds tests for `Pool`, and fixes some missing wakeup paths that were discovered as a result of those tests. In addition, it moves `Pool` to the new `Box<dyn Error>` error pattern.

### DIFF
--- a/tower-balance/Cargo.toml
+++ b/tower-balance/Cargo.toml
@@ -34,6 +34,7 @@ tower-layer = "0.1.0"
 tower-load = { version = "0.1.0", path = "../tower-load" }
 tower-service = "0.2.0"
 tower-util = "0.1.0"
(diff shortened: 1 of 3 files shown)


carllerche:
cc @adleong @olix0r

jonhoo:
@olix0r I'd love to merge this even given your pending changes, as otherwise `pool` is broken on `master` :)

PR #302, opened 2019-07-12
WIP RFC: tower-ready-cache

olix0r:
The kernel of the p2c balancer that manages service readiness seems likely to be useful to build other kinds of dispatchers like @jonhoo's lazy service pool and @adleong's weighted distributor.

This draft PR separates this kernel into a dedicated tower-ready-cache crate so we can start to flesh out these datails.

As discussed in https://github.com/tower-rs/tower/issues/286


tower-balance/src/p2c/service.rs
```diff
@@ -77,12 +56,7 @@ where
 
     /// Returns the number of endpoints currently tracked by the balancer.
     pub fn len(&self) -> usize {
-        self.ready_services.len() + self.unready_services.len()
-    }
-
-    // XXX `pool::Pool` requires direct access to this... Not ideal.
-    pub(crate) fn discover_mut(&mut self) -> &mut D {
-        &mut self.discover
+        self.ready_services.ready_len() + self.ready_services.unready_len()
```
jonhoo: Maybe just provide a `len()` function too? :p

tower-balance/src/p2c/service.rs
```diff
@@ -26,10 +24,7 @@ use tower_util::Ready;
 pub struct Balance<D: Discover, Req> {
     discover: D,
 
-    ready_services: IndexMap<D::Key, D::Service>,
-
-    unready_services: stream::FuturesUnordered<UnreadyService<D::Key, D::Service, Req>>,
-    cancelations: IndexMap<D::Key, oneshot::Sender<()>>,
+    ready_services: ReadyCache<D::Key, D::Service, Req>,
```
jonhoo: I wonder if this should now just be called `services`?

tower-ready-cache/src/lib.rs
```diff
@@ -0,0 +1,276 @@
...
+        self.cancelations.contains_key(key)
+    }
+
+    /// Attempts to cancel a service in the unready set.
+    ///
+    /// The service will be removed from the cache when `poll_unready` is called.
+    ///
+    /// Returns true iff the referenced service was unready and it was
+    /// succsesfully canceled.
+    pub fn cancel_unready<Q: Hash + Equivalent<K>>(&mut self, key: &Q) -> bool {
```
jonhoo: It doesn't look like this method is used externally anywhere (rather, everyone calls `evict`). Are you expecting this to ever be called specifically?

tower-ready-cache/src/lib.rs
```diff
@@ -0,0 +1,276 @@
...
+        if let Some(c) = self.cancelations.swap_remove(key) {
+            c.send(()).is_ok()
```
jonhoo: I wonder if dropping `c` is sufficient here?

tower-ready-cache/src/lib.rs
```diff
@@ -0,0 +1,276 @@
...
+    ///
+    /// The service will be promoted to the ready set as `poll_unready` is invoked.
+    ///
+    /// Note that this does not replace a service currently in the ready set with
+    /// an equivalent key until this service becomes ready or the existing
+    /// service becomes unready.
+    pub fn push_unready(&mut self, key: K, svc: S) {
+        let (tx, rx) = oneshot::channel();
+        if let Some(c) = self.cancelations.insert(key.clone(), tx) {
+            let _ = c.send(());
```
jonhoo: Same here, I _think_ this can just `drop(c)`

tower-ready-cache/src/lib.rs
```diff
@@ -0,0 +1,276 @@
...
+                    .expect("invalid ready index");
+                Err(e)
+            }
+        }
+    }
+
+    /// Swap-removes the indexed service from the ready set.
+    ///
+    /// Panics if the provided `index` is out of range.
+    pub fn process_ready_index<T, P>(&mut self, index: usize, process: P) -> T
```
jonhoo: This method should probably be given a more helpful name.

tower-ready-cache/src/lib.rs
olix0r: The UnreadyService future currently only treats successful cancelations as cancelations. Drops (which cause errors) are ignored.

It's probably best to make the error case panic completely, as it's not part of the intended lifecycle.

tower-ready-cache/src/lib.rs
```diff
@@ -0,0 +1,276 @@
...
+                debug!("evicting failed endpoint: {:?}", e);
+                self.ready_services
+                    .swap_remove_index(index)
+                    .expect("invalid ready index");
+                Err(e)
+            }
+        }
+    }
+
+    /// Swap-removes the indexed service from the ready set.
```
jonhoo: I think this warrants some further explanation. In particular, this could be mis-read to imply that it removes the _service_ after calling `process`, which isn't the case. Instead, it moves the service to the unready set.

tower-ready-cache/src/lib.rs
```diff
@@ -0,0 +1,276 @@
...
+    {
+        let (key, mut svc) = self
+            .ready_services
+            .swap_remove_index(index)
+            .expect("index out of range");
+
+        let t = (process)(&mut svc);
+
+        // If a new version of this service has been added to the
+        // unready set, don't overwrite it.
```
jonhoo: This case is weird to me. How would that happen?

tower-ready-cache/src/lib.rs
```diff
@@ -0,0 +1,276 @@
...
+        if !self.unready_contains(&key) {
+            self.push_unready(key, svc);
```
jonhoo: What if the service is immediately ready again even after `process`? Wouldn't it be a mistake to place it back in `unready`?

jonhoo:
I'm not sure I see exactly how this factors into `pool`, but I like the refactoring.

tower-ready-cache/src/lib.rs
olix0r: `ReadyCache::push_unready` may be called for a key that is already in the ready set and the ready instance will not be replaced until the unready instance becomes available.

Imagine that you get an update from service discovery that reconfigures a connection. I wouldn't want to be forced to tear down an old connection until the new connection is ready.

tower-ready-cache/src/lib.rs
olix0r: The contract is that `poll_unready` must be called following this.

This is much like poll_ready vs call -- `call` can never invoke `poll` (as call need not be invoked from a task context).  process need not be called from a task context.

olix0r:
@jonhoo I'm imagining that we could implement `pool` roughly as follows:

* A `Pool` service holds a `ReadyCache`.
* In `Pool::poll_ready`
  * `ReadyCache::poll_unready()` is called make services available.
  * If `ReadyCache::ready_len` is greater than 0, it returns ready immediately
  * Otherwise, a new service instance is provisioned and added to the ready cache.

The pool would probably want to enforce a maximum number of services that will be created at once...

I haven't thought through this in a very detailed way, but I think this could help accomplish something similar to what's there now without having to use any of the Discover API...

jonhoo:
Ohh, yes, I see what you mean. I think we'd still ideally want pool to support using a balancer though. Something like P2C still makes sense through a pool. Not sure how we abstract that though.

tower-ready-cache/src/lib.rs
```diff
@@ -224,24 +224,21 @@ where
     /// Swap-removes the indexed service from the ready set.
     ///
     /// Panics if the provided `index` is out of range.
-    pub fn process_ready_index<T, P>(&mut self, index: usize, process: P) -> T
-    where
-        P: FnOnce(&mut S) -> T,
-    {
+    pub fn call_ready_index(&mut self, index: usize, req: Req) -> S::Future {
```
jonhoo: I almost wonder whether the `next_ready_index` handling should also be folded into this.

olix0r: Interesting idea. We could make `poll_ready_index` part of the required API so that it is responsible for stashing the next ready index if it was ready... but I think we should _not_ make the index selection part of this module

jonhoo: Sorry, yes, that's what I meant -- the index stashing, not the index _selection_.

tower-balance/src/pool.rs
```diff
@@ -14,311 +14,117 @@
...
+        }
 
-                // don't reset the EWMA -- in theory, poll_ready should now start returning
-                // `Ready`, so we won't try to launch another service immediately.
-                // we clamp it to high though in case the # of services is limited.
-                self.ewma = self.options.high;
-            } else {
-                discover.load = Level::Normal;
-            }
+        let fut = self.make_service.make_service(());
```
jonhoo: This will create new services _very_ aggressively. What's the reasoning behind not maintaining a load estimate?

olix0r: That depends entirely on how inner services advertise readiness. If, for instance, an inner service uses a buffer, a new service would only ever be created when the buffer is at capacity.

This is still an experiment, obviously, but i'm curious if we avoid having `pool` dependent on a specific load metric implementation.  (Note, for instance, that the pool ewma doesn't really take into account time-to-first-byte or errors or other data that may be measured to gauge health/load).

olix0r:
Going to close this out for now. #303 includes the ready-cache refactor without changes to pool. #304 splits out the breaker prototype. I'll continue experimenting with the pool as the other PRs progress.

PR #303, opened 2019-07-18
Introduce tower-ready-cache

olix0r:
In #293, `balance` was refactored to manage dispatching requests over a
set of equivalent inner services that may or may not be ready.

This change extracts the core logic of managing a cache of ready
services into a dedicated crate, leaving the balance crate to deal with
node selection.


tower-balance/src/p2c/service.rs
```diff
@@ -104,82 +77,39 @@ where
...
-                }
-                Err((key, Error::Canceled)) => debug_assert!(!self.cancelations.contains_key(&key)),
-                Err((key, Error::Inner(e))) => {
-                    let error = e.into();
-                    debug!({ %error }, "dropping failed endpoint");
-                    let _cancel = self.cancelations.swap_remove(&key);
-                    debug_assert!(_cancel.is_some());
-                }
+            if let Err(e) = self.services.poll_pending() {
+                debug!("dropping endpoint: {:?}", e);
```
```suggestion
                debug!("dropping failed endpoint: {:?}", e);
```

tower-balance/src/p2c/service.rs
```diff
@@ -104,82 +77,39 @@ where
...
-                .next_ready_index
-                .and_then(|i| Self::repair_index(i, idx, self.ready_services.len()));
-            debug_assert!(!self.cancelations.contains_key(key));
-        } else if let Some(cancel) = self.cancelations.remove(key) {
-            let _ = cancel.send(());
-        }
-    }
-
     fn poll_unready(&mut self) {
+        // Note: this cannot perturb the order of ready services.
```
jonhoo: I'm not sure I follow this comment.

tower-balance/src/p2c/service.rs
```diff
@@ -195,46 +125,26 @@ where
...
                 let ready = if aload <= bload { aidx } else { bidx };
 
-                trace!({ a.idx = aidx, a.load = ?aload, b.idx = bidx, b.load = ?bload, ready = ?ready }, "choosing by load");
+                trace!(
+                    {
+                        a.index = aidx,
+                        a.load = ?aload,
+                        b.index = bidx,
+                        b.load = ?bload,
+                        ready = if ready == aidx { "a" } else { "b" },
```
```suggestion
                        chosen = if ready == aidx { "a" } else { "b" },
```

tower-balance/src/p2c/service.rs
olix0r: This isn't relevant in the balancer anymore. This was referring to the fact that poll_pending will only interact with the pending set and not the ready set, so the next-ready-index can't be impacted by this...

tower-balance/src/p2c/service.rs
```diff
@@ -255,77 +165,39 @@ where
 
     /// Prepares the balancer to process a request.
     ///
-    /// When `Async::Ready` is returned, `ready_index` is set with a valid index
-    /// into `ready` referring to a `Service` that is ready to disptach a request.
+    /// When `Async::Ready` is returned, `next_ready_index` is set with a valid
```
jonhoo: I don't think this code maintains `next_ready_index` explicitly any more?

tower-balance/src/p2c/service.rs
```diff
@@ -255,77 +165,39 @@ where
...
             // detector has changed the state of the service, it may be evicted
-            // from the ready set so that P2C can be performed again.
-            if let Some(index) = self.next_ready_index {
-                trace!({ next.idx = index }, "preselected ready_index");
-                debug_assert!(index < self.ready_services.len());
-
-                if let Ok(Async::Ready(())) = self.poll_ready_index_or_evict(index) {
+            // from the ready set so that another service can be selected.
+            if let Some(index) = index {
+                if let Ok(Async::Ready(())) = self.services.poll_next_ready_index(index) {
```
jonhoo: I assume this sets `next_ready_index` internally in `self.services` again given that we "take" it above?

tower-ready-cache/src/service.rs
```diff
@@ -0,0 +1,376 @@
...
+enum PendingError<K, E> {
+    Canceled(K),
+    Inner(K, E),
+}
+
+/// A Future that becomes satisfied when an `S`-typed service is ready.
+///
+/// May fail due to cancelation, i.e. if the service is removed from discovery.
+#[derive(Debug)]
+struct Pending<K, S, Req> {
```
jonhoo: Hmm, this is still a weird name to me.

tower-balance/src/p2c/service.rs
olix0r: Yeah. I'm updating the docs for poll_next_ready_index to be:

```rust
    /// Configures the ready service to be used by `call`.
    ///
    /// This may be called instead of `Service::poll_ready`.
    ///
    /// If the service is no longer ready and there are no equivalent unready
    /// services, it is moved back into the ready set and `Async::NotReady` is
    /// returned.
    ///
    /// If the service errors, it is removed and dropped and the error is returned.
    ///
    /// Otherwise, the `next_ready_index` is set with the provided index and
    /// `Async::Ready` is returned.
```

If you have suggestions on improving this, please have at it.

tower-ready-cache/src/service.rs
```diff
@@ -0,0 +1,376 @@
...
+    pub fn get_ready_index(&self, idx: usize) -> Option<(&K, &S)> {
+        self.ready.get_index(idx)
+    }
+
+    /// Obtains a mutable reference to a service in the ready set by index.
+    pub fn get_ready_index_mut(&mut self, idx: usize) -> Option<(&mut K, &mut S)> {
+        self.ready.get_index_mut(idx)
+    }
+
+    /// Evicts an item from the cache.
```
jonhoo: Should document the return value.

tower-ready-cache/src/service.rs
```diff
@@ -0,0 +1,376 @@
...
+    pub fn evict<Q: Hash + Equivalent<K>>(&mut self, key: &Q) -> bool {
+        let canceled = if let Some(c) = self.cancelations.swap_remove(key) {
```
jonhoo: I'd love a small comment here about what eviction does. Specifically, something that points out that there are two cases: the service is in the pending set (not ready), or in the ready set. If the former, we cancel it, and it will be evicted when it becomes ready (why?). If the latter, we just evict it immediately.

tower-balance/src/p2c/service.rs
jonhoo: What does "equivalent unready service" mean here?

Also, is the concept of `next_ready_index` documented anywhere?

tower-ready-cache/src/service.rs
```diff
@@ -0,0 +1,380 @@
...
+enum PendingError<K, E> {
+    Canceled(K),
+    Inner(K, E),
+}
+
+/// A Future that becomes satisfied when an `S`-typed service is ready.
+///
+/// May fail due to cancelation, i.e. if the service is removed from discovery.
+#[derive(Debug)]
+struct EvictableReady<K, S, Req> {
```
jonhoo: Isn't eviction specifically for _unready_ services?

tower-ready-cache/src/service.rs
```diff
@@ -0,0 +1,380 @@
...
+            ready_index = Some(0);
+        }
+    }
+
+    fn call(&mut self, req: Req) -> Self::Future {
+        self.call_next_ready(req).map_err(Into::into)
+    }
+}
+
+// === Pending ===
```
jonhoo: This name also needs to be updated I think.

tower-ready-cache/src/service.rs
```diff
@@ -0,0 +1,380 @@
...
+{
+    /// Pushes a new service onto the unready set.
+    ///
+    /// The service will be promoted to the ready set as `poll_pending` is invoked.
+    ///
+    /// Note that this does not replace a service currently in the ready set with
+    /// an equivalent key until this service becomes ready or the existing
+    /// service becomes unready.
+    pub fn push_service(&mut self, key: K, svc: S) {
+        let (tx, cancel) = oneshot::channel();
```
jonhoo: I wonder if there's a way we can avoid creating a new oneshot each time a service is pushed. Might be tricky without exposing the inner wrapper though.

tower-ready-cache/src/service.rs
```diff
@@ -0,0 +1,380 @@
...
+                Ok(Async::Ready(None)) => return Ok(Async::Ready(())),
+                Ok(Async::Ready(Some((key, svc)))) => {
+                    trace!("endpoint ready");
+                    self.cancelations.remove(&key).expect("cancel sender lost");
+                    self.ready.insert(key, svc);
+                }
+                Err(PendingError::Canceled(_)) => {
+                    debug!("endpoint canceled");
+                    // Either the cancelation has already been removed, or a
+                    // replacement unready service has already been pushed.
```
jonhoo: I think you can make a stronger statement here, namely "The cancellation for this service was removed in order to cause this cancellation."

tower-ready-cache/src/service.rs
```diff
@@ -0,0 +1,380 @@
...
+                    return Err(error::Failed(key, e.into()));
+                }
+            }
+        }
+    }
+
+    /// Configures the ready service to be used by `call`.
+    ///
+    /// This may be called instead of invoking `Service::poll_ready`
+    pub fn poll_next_ready<Q: Hash + Equivalent<K>>(
```
jonhoo: This probably also deserves some more docs.

tower-ready-cache/src/service.rs
```diff
@@ -0,0 +1,380 @@
...
+        &mut self,
+        key: &Q,
+    ) -> Poll<(), error::Failed<K>> {
+        match self.ready.get_full_mut(key) {
+            None => Ok(Async::NotReady),
+            Some((index, _, _)) => self.poll_next_ready_index(index),
+        }
+    }
+
+    /// Consume the `next_ready_index`.
```
jonhoo: A note about the implication of this on `call`, and the expected pattern of calling `poll_next_ready_index` subsequently would be good here.

tower-ready-cache/src/service.rs
```diff
@@ -0,0 +1,380 @@
...
+    pub fn take_next_ready_index(&mut self) -> Option<usize> {
+        self.next_ready_index.take()
+    }
+
+    /// Configures the ready service to be used by `call`.
+    ///
+    /// This may be called instead of `Service::poll_ready`.
+    ///
+    /// If the service is no longer ready and there are no equivalent unready
```
jonhoo: What does "equivalent unready services" mean here?

tower-ready-cache/src/service.rs
```diff
@@ -0,0 +1,380 @@
...
+                    .swap_remove_index(index)
+                    .expect("invalid ready index");
+                Err(error::Failed(key, e.into()))
+            }
+        }
+    }
+
+    /// Calls the next ready service.
+    ///
+    /// `poll_next_ready_index` must have returned Ready before this is invoked;
```
jonhoo: I _think_ it's okay to call any of the `poll_*ready*` methods before calling this, no? `poll_next_ready` for example.

tower-ready-cache/src/service.rs
jonhoo: Maybe by having a private function that pushes an `EvictableReady` directly that we use internally (in `call_next_ready` for example), and then a `pub` like this one that creates a new one.

tower-ready-cache/src/service.rs
jonhoo: `poll_pending` is specifically _not_ sufficient though.

tower-ready-cache/src/service.rs
```diff
@@ -0,0 +1,380 @@
...
+                match self.poll_next_ready_index(index) {
+                    Ok(Async::NotReady) => {}
+                    Ok(Async::Ready(())) => return Ok(Async::Ready(())),
+                    Err(e) => return Err(e.into()),
+                }
+            }
+
+            if self.ready_len() == 0 {
+                if self.pending_len() == 0 {
+                    // If the cache is totally exhausted, we have no means to notify the
```
jonhoo: missing word

tower-ready-cache/src/service.rs
```diff
@@ -0,0 +1,380 @@
...
+    K: Clone + Eq + Hash,
+    S: Service<Req>,
+    S::Error: Into<error::Error>,
+    error::Failed<K>: Into<error::Error>,
+{
+    type Response = S::Response;
+    type Error = error::Error;
+    type Future = future::MapErr<S::Future, fn(S::Error) -> Self::Error>;
+
+    fn poll_ready(&mut self) -> Poll<(), Self::Error> {
```
jonhoo: Should this document that it's not okay to call any other methods on `self` between getting `Ready` from this and calling `call`?

tower-ready-cache/src/service.rs
```diff
@@ -0,0 +1,380 @@
...
+    fn poll(&mut self) -> Poll<Self::Item, Self::Error> {
+        if self.cancel.poll().expect("cancel sender lost").is_ready() {
+            let key = self.key.take().expect("polled after ready");
+            return Err(PendingError::Canceled(key));
+        }
+
+        match self.ready.poll() {
+            Ok(Async::NotReady) => Ok(Async::NotReady),
+            Ok(Async::Ready(svc)) => {
+                let key = self.key.take().expect("polled after ready");
```
jonhoo: This _could_ also be a poll after `Err` technically. Same with the one below.

tower-ready-cache/src/service.rs
olix0r: I believe it's a general contract for service that poll_ready be called immediately before `call` (@carllerche ?), but I'll document something to that effect.

tower-ready-cache/src/service.rs
olix0r: `EvictableReady` wraps the `tower_util::Ready` future to interact with `evict`.

jonhoo: I think there is some overloading of terms here with "ready" also referring to a _service that is ready_. Perhaps just `UnreadyService`?

tower-balance/Cargo.toml
```diff
@@ -29,16 +29,17 @@ default = ["log"]
 [dependencies]
 futures = "0.1.26"
 indexmap = "1.0.2"
-tracing = "0.1"
 rand = "0.6.5"
```
jonhoo: Actually, can you bump `rand` while you're at it?

tower-ready-cache/src/service.rs
carllerche: It probably would be possible if we don't use `FuturesUnordered` and instead implement something custom.

olix0r: I modified the ready-cache to only create a single oneshot per service, reusing it throughout its lifetime. This should generally amortize the cost of using a oneshot here.

olix0r:
> Instead, the ready cache could "just" have `poll_ready_by_key(...)` and `call_by_key(...)`?

it's important for at least the balancer to be able to just treat the cache as an indexable array that supports removal by key, so we would always want to allow by_index access -- otherwise the balancer would need to duplicate the list of keys and keep it in sync with the cache in order to be able to implement p2c.

As to whether `next_ready_index` belongs in this or not, I can see both sides of the argument. It seems rare that the access pattern would be other than what's implemented here, so bundling the behavior is convenient; but it may add some contract complexity for the ready cache.

olix0r:
I'm reviving this PR.

LucioFranco:
I'd like to get this merged and clean up other PRs then move 0.1 support to its own branch and 0.3 onto master.

olix0r:
@LucioFranco @jonhoo @carllerche OK. I'm fairly happy with this now. I've removed any concept of a `next_ready` from the cache (and moved this up into the balancer); and as a result of that, the cache no longer implements `Service` directly. I've added a lengthy doc comment on `ReadyCache` describing its contract.   I've also tried to clean up naming, referring only to _pending_ and _ready_ throughout.

Please take a look! I'm especially interested to know where the contract is insufficiently documented.

tower-balance/src/p2c/service.rs
```diff
@@ -98,88 +73,47 @@ where
...
-                }
-                Err((key, Error::Canceled)) => debug_assert!(!self.cancelations.contains_key(&key)),
-                Err((key, Error::Inner(e))) => {
-                    let error = e.into();
-                    debug!({ %error }, "dropping failed endpoint");
-                    let _cancel = self.cancelations.swap_remove(&key);
-                    debug_assert!(_cancel.is_some());
-                }
+            if let Err(error) = self.services.poll_pending() {
+                debug!(%error, "dropping failed endpoint");
```
jonhoo: I'm confused by this loop — why does it break immediately upon success? Shouldn't it break only when `poll_pending` returns `Async::NotReady`?

tower-balance/src/p2c/service.rs
```diff
@@ -253,79 +165,41 @@ where
...
-            // from the ready set so that P2C can be performed again.
-            if let Some(index) = self.next_ready_index {
-                trace!({ next.idx = index }, "preselected ready_index");
-                debug_assert!(index < self.ready_services.len());
-
-                if let Ok(Async::Ready(())) = self.poll_ready_index_or_evict(index) {
-                    return Ok(Async::Ready(()));
+            // from the ready set so that another service can be selected.
+            if let Some(index) = self.ready_index.take() {
+                match self.services.check_ready_index(index) {
```
jonhoo: This "extra" ready check is still so weird to me, and I worry that it might have adverse consequences as calling `poll_ready` now needs to be idempotent. Maybe that's fine, I'm just not sure how much the extra check buys us..

tower-balance/src/pool/make_ready.rs
```diff
@@ -0,0 +1,63 @@
+#![allow(missing_docs)]
```
jonhoo: Why?

tower-balance/src/pool/make_ready.rs
```diff
@@ -0,0 +1,63 @@
...
+#![doc(hidden)]
```
jonhoo: Why are these needed?

tower-buffer/src/worker.rs
```diff
@@ -176,11 +176,11 @@ where
                     // Wait for the service to be ready
                     tracing::trace!(
                         resumed = !first,
-                        message = "worker received request; waiting for service readiness"
```
jonhoo: These seem like mostly orthogonal changes to this PR?

tower-balance/src/p2c/service.rs
jonhoo: I see and follow the argument made in the API docs for `ReadyCache` for why you need to check ready again, it's just a weird contract to me. You always run the risk of the service _just_ becoming not ready between the last time you called `poll_ready` and `call`. Does calling `poll_ready` again just before `call` serve that much of a purpose? And is it worth the cost of calling `poll_ready` again?

tower-buffer/src/worker.rs
olix0r: Yeah, the tracing syntax was improved. I can minimize the diff.

tower-balance/src/p2c/service.rs
olix0r: > `poll_ready` now needs to be idempotent.

This is already the case! You already have to expect that `poll_ready` may be called redundantly -- consider a Future that needs multiple services to be ready before it can take an action: there will almost certainly be redundant calls:
```rust
try_ready!(a.poll_ready())
try_ready!(b.poll_ready())
a.call(c).join(b.call(d))
```

> You always run the risk of the service just becoming not ready between the last time you called poll_ready and call

Sure. There will always be the potential for a resource to become unavailable just as its used. It doesn't mean that we shouldn't try to minimize that potential, though.

Specifically, we have decent guarantees that the thread has not changed/reactor state between poll_ready and call because we know the task has not been interrupted! If we poll_ready, and then let the thread/reactor do additional processing before call, it seems _much_ more likely for the state to change. The readiness contract is: given the most up-to-date state of the runtime, can we proceed? That state _can't_ change between poll_ready and call if they run in a single task invocation, really. (Yes, the external world can change, but we're operating under the best guarantees the local task can offer us).

> Does calling poll_ready again just before call serve that much of a purpose? And is it worth the cost of calling poll_ready again?

This depends entirely on the cost of calling `poll_ready`. I posit that implementors _should_ make `poll_ready` cheap to call, especially if it's already in a ready state. As discussed above, I think this guidance holds regardless of the semantics of ReadyCache.

--

On further thought, it is not _required_ that a ready service be checked before call is invoked. But it is _necessary_ (for our needs in the balancer, at least) to have a means to check a ready service redundantly. The balancer, specifically, could have services that become ready and remain unused for _minutes_. It's absolutely necessary to have means to re-check for readiness in this case.

jonhoo: Okay, I think that answer persuades me. Maybe we could add a fast-path later that doesn't call `poll_ready` again if we haven't been interrupted since the service was last discovered to be ready, but let's not engage in premature optimization here.

tower-balance/src/p2c/service.rs
olix0r: if poll_pending returns `Ready(())`, there are no currently-pending services (so no more work to do). If not ready, also no more work to do at the moment. I'll add a comment.

tower-balance/src/p2c/service.rs
olix0r: I'm going to update the docs to reflect a more relaxed contract where calls to `check_ready*`  are optional; though we should continue to use it from the Balancer...

tower-ready-cache/src/cache.rs
```diff
@@ -0,0 +1,368 @@
...
+/// * a set of _ready_ services that have previously polled ready.
+///
+/// As each `S` typed `Service` is added to the cache via `ReadyCache::push`, it
+/// added to the _pending set_. As `ReadyCache::poll_pending` is invoked, pending
+/// services are polled and added to the _ready set_.
+///
+/// The ready set can hold services for an abitrarily long time. During this
+/// time, the runtime may process events that invalidate that ready state (for
+/// instance, if a keepalive detects a lost connection). Therfore it is
+/// **required** that a ready service is checked via `ReadyCache::check_ready`
```
olix0r: Relax this statement

tower-balance/src/p2c/service.rs
```diff
@@ -253,79 +177,52 @@ where
...
-                if let Ok(Async::Ready(())) = self.poll_ready_index_or_evict(index) {
-                    return Ok(Async::Ready(()));
+            // before a request is dispatched to it (i.e. in the same task
+            // invocation). If, e.g., a failure detector has changed the state
+            // of the service, it may be evicted from the ready set so that
+            // another service can be selected.
+            if let Some(index) = self.ready_index.take() {
+                match self.services.check_ready_index(index) {
+                    Err(Failed(_, error)) => {
+                        // If the ready endpoitn fails
```
olix0r: endpoint

tower-ready-cache/src/cache.rs
```diff
@@ -20,13 +20,17 @@ use tower_service::Service;
 /// added to the _pending set_. As `ReadyCache::poll_pending` is invoked, pending
 /// services are polled and added to the _ready set_.
 ///
+/// `ReadyCache::call_ready` (or `ReadyCache::call_ready_index`) dispatch a
```
jonhoo: _dispatches_

tower-ready-cache/src/cache.rs
```diff
@@ -20,13 +20,17 @@ use tower_service::Service;
 /// added to the _pending set_. As `ReadyCache::poll_pending` is invoked, pending
```
jonhoo: missing "is"

tower-ready-cache/src/cache.rs
```diff
@@ -20,13 +20,17 @@ use tower_service::Service;
...
 /// services are polled and added to the _ready set_.
 ///
+/// `ReadyCache::call_ready` (or `ReadyCache::call_ready_index`) dispatch a
+/// request to the specified service, but panic if the specified service is not
```
jonhoo: _panics_

PR #304, opened 2019-07-18
WIP RFC Introduce tower-breaker

olix0r:
tower-breaker introduces a circuit-breaking middleware that instruments
load-based backpressure.

This is a work-in-progress and should be fleshed out more fully before
merging.

### DIFF
--- a/tower-breaker/src/lib.rs
+++ b/tower-breaker/src/lib.rs
@@ -7,7 +7,7 @@ use tower_load::Load;
 use tower_service as svc;
 
 /// Determmines readiness based on an `M`-typed load metric.
-pub trait Breaker<M: PartialOrd> {
+pub trait Breaker<M> {
     /// Checks to see whether the breaker is open, given a load metric.
     ///
     /// If the breaker is closed, NotReady is returned and the task ust be
@@ -62,7 +62,7 @@ pub mod delay {
     use std::time::Instant;
     use tokio_timer::Delay;
 
-    pub trait BreakUntil<M: PartialOrd> {
+    pub trait BreakUntil<M> {
         fn break_until(&mut self, load: M) -> Option<Instant>;
     }
 
@@ -71,7 +71,7 @@ pub mod delay {
         delay: Option<Delay>,
     }
 
-    impl<M: PartialOrd, B: BreakUntil<M>> super::Breaker<M> for Breaker<B> {
+    impl<M, B: BreakUntil<M>> super::Breaker<M> for Breaker<B> {
         fn poll_breaker(&mut self, load: M) -> Poll<(), error::Error> {
             // Even if there's already a delay, update the inner breaker with
             // the new load and reset any pre-existing delay.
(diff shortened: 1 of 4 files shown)


tower-breaker/src/lib.rs
```diff
@@ -0,0 +1,103 @@
...
+            // Even if there's already a delay, update the inner breaker with
+            // the new load and reset any pre-existing delay.
+            //
+            // The `delay` is stored so that the breaker task is notified after
+            // the delay expires.
+            self.delay = self.break_until.break_until(load).map(Delay::new);
+
+            if let Some(delay) = self.delay.as_mut() {
+                try_ready!(delay.poll());
+                self.delay = None;
```
hawkw: Can't we just `take` the delay before polling it? Even if it returns `NotReady`, the next call to `poll_breaker` will still replace it when we generate the next delay.

olix0r: Per the comment:

```
            // The `delay` is stored so that the breaker task is notified after
            // the delay expires.
```

I don't think we can trust the timer to notify the task if we drop the handle.

hawkw: Ah, yup, never mind me.

tower-breaker/src/lib.rs
```diff
@@ -0,0 +1,103 @@
...
+/// Determmines readiness based on an `M`-typed load metric.
+pub trait Breaker<M> {
+    type Error;
+
+    /// Checks to see whether the breaker is open, given a load metric.
+    ///
+    /// If the breaker is closed, NotReady is returned and the task ust be
+    /// notified when the breaker should be polled again with an updated load
+    /// metric.
+    fn poll_breaker(&mut self, load: M) -> Poll<(), Self::Error>;
```
carllerche: If the `load` value changes between calls to `poll_breaker`, then I assume the inner state needs to be updated to reflect the last value of `load` as the most recent? In the past, arguments to `poll_*` fns have been problematic because how to handle the argument changing between calls was not obvious.

tower-breaker/src/lib.rs
```diff
@@ -0,0 +1,103 @@
+//! A middleware that blocks readiness based on an inner service's load.
+
+pub mod error;
+
+use futures::{future, try_ready, Future, Poll};
+use tower_load::Load;
+use tower_service as svc;
+
+/// Determmines readiness based on an `M`-typed load metric.
```
gj:
```suggestion
/// Determines readiness based on an `M`-typed load metric.
```

tower-breaker/src/lib.rs
```diff
@@ -0,0 +1,103 @@
...
+pub trait Breaker<M> {
+    type Error;
+
+    /// Checks to see whether the breaker is open, given a load metric.
+    ///
+    /// If the breaker is closed, NotReady is returned and the task ust be
```
gj:
```suggestion
    /// If the breaker is closed, NotReady is returned and the task must be
```

olix0r:
We'll revive this later

LucioFranco:
@olix0r yeah, I'd be happy to take a stab at this once we get everything over to `std::future::Future`?

Issue #306, opened 2019-07-21
Expose subcrates default features in `tower`.


samscott89:
Tower pulls in tower-buffer  which has the `log` feature enabled by default. Would be good to re-expose this as a default feature in tower so it can be turned off if necessary. Presumably will be the same for other tower-* crates if/when tracing gets added to them.

samscott89:
Perhaps better to solve this in tracing - https://github.com/tokio-rs/tracing/issues/204

jonhoo:
This is fixed by #431, and specifically #432.

PR #307, opened 2019-07-21
Add feature flag for log support.

samscott89:
Just a quick potential fix to #306 (if desired).

### DIFF
--- a/tower/Cargo.toml
+++ b/tower/Cargo.toml
@@ -24,12 +24,13 @@ keywords = ["io", "async", "non-blocking", "futures", "service"]
 edition = "2018"
 
 [features]
-default = ["full"]
+default = ["std-log-support", "full"]
+std-log-support = ["tower-buffer/log"]
 full = []
 
 [dependencies]
 futures = "0.1.26"
-tower-buffer = "0.1.0"
+tower-buffer = { version = "0.1.0", default-features = false }
 tower-discover = "0.1.0"
 tower-layer = "0.1.0"
 tower-limit = "0.1.0"


LucioFranco:
@samscott89 would this still provide tracing support?

samscott89:
I believe so. AFAICT, the idea seems to be move everything to tracing, and then use the `tracing/log` feature to provide compatibility with `log`. (I'm going off of ~ this convo https://github.com/tower-rs/tower/pull/298#discussion_r302291421). Which is on by default since people using log would other see their logs disappear.

The problem is for people using tracing first, and redirecting log outputs to tracing, we now get double the logs. And also a ton of noise since `tracing/log` logs a bunch of extra stuff, like when spans are entered. Although that is a separate thing.

hawkw:
> The problem is for people using tracing first, and redirecting log outputs to tracing, we now get double the logs. And also a ton of noise since tracing/log logs a bunch of extra stuff, like when spans are entered. Although that is a separate thing.

@samscott89 I think there's potentially some strategies we could use to work around the problem of duplicate logs when both `tracing-log`'s `LogTracer` and the log feature flag are in use, though they might need some design --- there's a `tracing` ticket in there for sure.

samscott89:
> > The problem is for people using tracing first, and redirecting log outputs to tracing, we now get double the logs. And also a ton of noise since tracing/log logs a bunch of extra stuff, like when spans are entered. Although that is a separate thing.
> 
> @samscott89 I think there's potentially some strategies we could use to work around the problem of duplicate logs when both `tracing-log`'s `LogTracer` and the log feature flag are in use, though they might need some design --- there's a `tracing` ticket in there for sure.

Hmm, yeah the idea that everyone would expose a `log` feature does seem a bit unwieldy.

I submitted https://github.com/tokio-rs/tracing/issues/204 as a starting point

samscott89:
Okay, I'll close this PR then in favour of solving it in tracing.

PR #315, opened 2019-08-31
Update tower-buffer and tower-reconnect to std::future::Future

LucioFranco:
I did not do the tests yet, but I have tested both with hyper.


tower-buffer/src/future.rs
```diff
@@ -1,26 +1,26 @@
...
+use crate::{error::Error, message};
+use std::future::Future;
+use std::pin::Pin;
+use std::task::{Context, Poll};
 
 /// Future eventually completed with the response to the original request.
 pub struct ResponseFuture<T> {
     state: ResponseState<T>,
 }
+impl<T> Unpin for ResponseFuture<T> {}
```
jonhoo: So, the claim here is that we never move `state`, but I don't think that's true? Specifically, we do `self.state = ` in `poll` below. Why is that okay?

tower-buffer/src/future.rs
```diff
@@ -1,26 +1,26 @@
...
 
 enum ResponseState<T> {
     Failed(Option<Error>),
     Rx(message::Rx<T>),
     Poll(T),
 }
 
-impl<T> ResponseFuture<T>
+impl<T, A, B> ResponseFuture<T>
```
jonhoo: Can we rename these to `F, T, E`?

tower-buffer/src/future.rs
```diff
@@ -35,36 +35,36 @@ where
...
+    B: Into<Error>,
 {
-    type Item = T::Item;
-    type Error = Error;
-
-    fn poll(&mut self) -> Poll<Self::Item, Self::Error> {
-        use self::ResponseState::*;
+    type Output = Result<A, Error>;
 
+    fn poll(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
```
jonhoo: I don't think you need the `<'_>` here

LucioFranco: Say more? `Context` always needs a life time right?

tower-buffer/src/future.rs
LucioFranco: yup can do

tower-buffer/src/future.rs
```diff
@@ -35,36 +35,36 @@ where
...
-                    return Err(e.take().expect("polled after error"));
+                ResponseState::Failed(ref mut e) => {
+                    return Poll::Ready(Err(e.take().expect("polled after error")));
                 }
-                Rx(ref mut rx) => match rx.poll() {
-                    Ok(Async::Ready(Ok(f))) => fut = f,
-                    Ok(Async::Ready(Err(e))) => return Err(e.into()),
-                    Ok(Async::NotReady) => return Ok(Async::NotReady),
-                    Err(_) => return Err(Closed::new().into()),
+                ResponseState::Rx(ref mut rx) => match unsafe { Pin::new_unchecked(rx) }.poll(cx) {
```
jonhoo: So, this `unsafe` is okay because the only time we move `rx` is when we re-assign `self.state` below, and that's the drop point for `rx`, and that's okay? Would be good to place that justification in a comment here.

tower-buffer/src/future.rs
```diff
@@ -35,36 +35,36 @@ where
...
+                    Poll::Ready(Ok(Ok(f))) => fut = f,
+                    Poll::Ready(Ok(Err(e))) => return Poll::Ready(Err(e.into())),
+                    Poll::Ready(Err(e)) => return Poll::Ready(Err(e.into())),
+                    Poll::Pending => return Poll::Pending,
```
jonhoo: Any particular reason you don't use the `ready!` macro around `poll` above instead?

tower-buffer/src/future.rs
```diff
@@ -35,36 +35,36 @@ where
...
+                    // Err(_) => return Err(Closed::new().into()),
                 },
-                Poll(ref mut fut) => {
-                    return fut.poll().map_err(Into::into);
+                ResponseState::Poll(ref mut fut) => {
+                    return unsafe { Pin::new_unchecked(fut) }
```
jonhoo: I _think_ the argument for why this is safe is that once we are in `Poll`, we will never hit the `self.state = ` below, nor will we move `fut` or give out `&mut` to it until it is dropped. Should also be in a comment.

tower-buffer/src/future.rs
LucioFranco: That should drop the inner state instead of mem replacing it, right? So in theory that should be safe since we never hand out a &mut T nor a Pin.

tower-buffer/src/service.rs
```diff
@@ -81,9 +81,9 @@ where
     type Error = Error;
     type Future = ResponseFuture<T::Future>;
 
-    fn poll_ready(&mut self) -> Poll<(), Self::Error> {
+    fn poll_ready(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
```
jonhoo: Same here, I don't think you need the `<'_>`?

tower-buffer/src/future.rs
LucioFranco: Yup, thats what Im thinking, will add comments.

tower-buffer/src/worker.rs
```diff
@@ -28,6 +30,13 @@ where
     handle: Handle,
 }
 
+impl<T, R> Unpin for Worker<T, R>
```
jonhoo: What's the argument for why this is okay?

 - `current_message`: do we guarantee that once we create a `Pin<Message>`, we never move it again?
 - `rx`: is `mpsc::Receiver` unconditionally `Unpin`?
 - `service`: do we ever create a `Pin` to the `Service`?
 - `finish`, `failed`, and `handle` are fine, since we never give out `Pin` to them

tower-buffer/src/worker.rs
```diff
@@ -85,35 +94,38 @@ where
     ///
     /// If a `Message` is returned, the `bool` is true if this is the first time we received this
     /// message, and false otherwise (i.e., we tried to forward it to the backing service before).
-    fn poll_next_msg(&mut self) -> Poll<Option<(Message<Request, T::Future>, bool)>, ()> {
+    fn poll_next_msg(
+        &mut self,
+        cx: &mut Context<'_>,
```
jonhoo: Same: no `<'_>`

tower-buffer/src/worker.rs
jonhoo: I _think_ the argument for `current_message` is that we never give out a `Pin` to the message while it's sitting in `current_message`.

tower-buffer/src/worker.rs
```diff
@@ -141,6 +153,7 @@ where
         *inner = Some(error.clone());
         drop(inner);
 
+        println!("close");
```
jonhoo: Delete

tower-buffer/src/worker.rs
```diff
@@ -155,16 +168,15 @@ where
     T: Service<Request>,
     T::Error: Into<Error>,
 {
-    type Item = ();
-    type Error = ();
+    type Output = ();
 
-    fn poll(&mut self) -> Poll<(), ()> {
+    fn poll(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<()> {
```
jonhoo: No `<'_>`

tower-buffer/tests/buffer.rs
```diff
@@ -1,3 +1,5 @@
+#![cfg(feature = "broken")]
```
jonhoo: Not sure what this is doing here?

tower-layer/src/lib.rs
```diff
@@ -20,7 +20,7 @@
 ///
 /// Take request logging as an example:
 ///
-/// ```rust
+/// ```rust,ignore
```
jonhoo: We should probably fix at least the doctests before merging this. Ideally also the tests.

tower-reconnect/Cargo.toml
```diff
@@ -23,7 +23,6 @@ edition = "2018"
 publish = false
 
 [dependencies]
-log = "0.4.1"
-futures = "0.1.26"
-tower-service = "0.2.0"
-tower-util = "0.1.0"
+tracing = "0.1"
```
jonhoo: Moving to `tracing` should probably be done in a separate PR

tower-reconnect/src/future.rs
```diff
@@ -1,25 +1,32 @@
...
+    future::Future,
+    pin::Pin,
+    task::{Context, Poll},
+};
 
 pub struct ResponseFuture<F> {
     inner: F,
 }
 
+impl<F> Unpin for ResponseFuture<F> {}
```
jonhoo: The argument here is that we never move `inner` and never give out `&mut` to it?

tower-reconnect/src/future.rs
```diff
@@ -1,25 +1,32 @@
...
+    F: Future<Output = Result<T, E>>,
+    E: Into<Error>,
 {
-    type Item = F::Item;
-    type Error = Error;
+    type Output = Result<T, Error>;
 
-    fn poll(&mut self) -> Poll<Self::Item, Self::Error> {
-        self.inner.poll().map_err(Into::into)
+    fn poll(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
```
jonhoo: `<'_>`

tower-reconnect/src/future.rs
```diff
@@ -1,25 +1,32 @@
...
+        unsafe { Pin::new_unchecked(&mut self.inner) }
```
jonhoo: The argument here is pretty straightforward I guess -- we never do anything to self.inner other than `Pin` it and call `poll`, so we trivially uphold the contract. I think I'd prefer `map_unchecked_mut` here though, as it's a little less unsafe.

tower-reconnect/src/lib.rs
```diff
@@ -59,19 +64,19 @@ where
     type Error = Error;
     type Future = ResponseFuture<S::Future>;
 
-    fn poll_ready(&mut self) -> Poll<(), Self::Error> {
+    fn poll_ready(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
```
jonhoo: `<'_>`

tower-reconnect/src/lib.rs
```diff
@@ -51,6 +55,7 @@ where
 impl<M, Target, S, Request> Service<Request> for Reconnect<M, Target>
 where
     M: Service<Target, Response = S>,
+    M::Future: Unpin,
```
jonhoo: Why require `Unpin` here?

tower-buffer/src/future.rs
jonhoo: You don't need to name the lifetime if you are just taking a reference to it. If you just remove it, it'll still compile just fine.

ebkalderon:
Is this PR still being worked on?

PR #316, opened 2019-09-01
Update tower-test to std::future::Future

JohnDoneth:
It works but I'm not entirely happy with the amount of `Box::pin` and `.as_mut` combos. I'm assuming there's a more ergonomic way to go about `Pin`'ing types? Any suggestions or improvements will be gladly accepted.


tower-test/src/mock/future.rs
```diff
@@ -23,18 +28,17 @@ impl<T> ResponseFuture<T> {
...
 
-    fn poll(&mut self) -> Poll<Self::Item, Self::Error> {
+    fn poll(mut self: Pin<&mut Self>, cx: &mut Context) -> Poll<Self::Output> {
         match self.rx {
-            Some(ref mut rx) => match rx.poll() {
-                Ok(Async::Ready(Ok(v))) => Ok(v.into()),
-                Ok(Async::Ready(Err(e))) => Err(e),
-                Ok(Async::NotReady) => Ok(Async::NotReady),
-                Err(_) => Err(error::Closed::new().into()),
+            Some(ref mut rx) => match Box::pin(rx).as_mut().poll(cx) {
```
taiki-e: It can be written as follows by using [`pin_utils::unsafe_pinned!(rx: Option<Rx<T>>)`](https://docs.rs/pin-utils/0.1.0-alpha.4/pin_utils/macro.unsafe_pinned.html) and [`Option::as_pin_mut()`](https://doc.rust-lang.org/nightly/core/option/enum.Option.html#method.as_pin_mut):
```rust
    match self.rx().as_pin_mut() {
        Some(rx) => match rx.poll(cx) {
        // ...
    
```

JohnDoneth: Done!

tower-test/Cargo.toml
```diff
@@ -8,23 +8,22 @@ name = "tower-test"
...
 
 [dependencies]
-futures = "0.1.26"
-tokio-sync = "0.1.3"
-tower-service = "0.2.0"
-
-[dev-dependencies]
-tower = { version = "0.1.0", path = "../tower" }
+futures-util-preview = "0.3.0-alpha.18"
+futures-executor-preview = "0.3.0-alpha.18"
```
LucioFranco: We should drop this dependency and use https://docs.rs/tokio-test/0.2.0-alpha.4/tokio_test/ which contains mock tasks.

PR #317, opened 2019-09-02
Update tower-balance to std::future::Future

LucioFranco:
This is an update of tower balance and its siblings. No tests were updated waiting on tower-test but the functionality here should be the same. This also only updates the p2c section. We can update pool in a following PR.

### DIFF
--- a/tower-balance/src/p2c/service.rs
+++ b/tower-balance/src/p2c/service.rs
@@ -1,12 +1,16 @@
 use crate::error;
-use futures::{future, stream, try_ready, Async, Future, Poll, Stream};
+use futures_util::{stream, try_future, StreamExt, TryFutureExt};
 use indexmap::IndexMap;
+use pin_project::pin_project;
 use rand::{rngs::SmallRng, FromEntropy};
+use std::future::Future;
+use std::marker::PhantomData;
+use std::pin::Pin;
+use std::task::{Context, Poll};
 use tokio_sync::oneshot;
 use tower_discover::{Change, Discover};
 use tower_load::Load;
 use tower_service::Service;
-use tower_util::Ready;
 use tracing::{debug, trace};
 
 /// Distributes requests across inner services using the [Power of Two Choices][p2c].
@@ -38,14 +42,20 @@ pub struct Balance<D: Discover, Req> {
     rng: SmallRng,
 }
 
+impl<D: Discover, Req> Unpin for Balance<D, Req> {}
+
 /// A Future that becomes satisfied when an `S`-typed service is ready.
 ///
 /// May fail due to cancelation, i.e. if the service is removed from discovery.
+#[pin_project]
 #[derive(Debug)]
 struct UnreadyService<K, S, Req> {
+    // This is a false positive
     key: Option<K>,
+    #[pin]
     cancel: oneshot::Receiver<()>,
-    ready: tower_util::Ready<S, Req>,
+    #[pin]
+    ready: Ready<S, Req>,
 }
 
 enum Error<E> {
@@ -98,19 +108,21 @@ where
     /// Polls `discover` for updates, adding new items to `not_ready`.
     ///
     /// Removals may alter the order of either `ready` or `not_ready`.
-    fn poll_discover(&mut self) -> Poll<(), error::Discover> {
+    fn poll_discover(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), error::Discover>> {
         debug!("updating from discover");
         loop {
-            match try_ready!(self.discover.poll().map_err(|e| error::Discover(e.into()))) {
-                Change::Remove(key) => {
+            match self.discover.poll(cx) {
+                Poll::Ready(Ok(Change::Remove(key))) => {
                     trace!("remove");
                     self.evict(&key)
                 }
-                Change::Insert(key, svc) => {
+                Poll::Ready(Ok(Change::Insert(key, svc))) => {
                     trace!("insert");
                     self.evict(&key);
                     self.push_unready(key, svc);
                 }
+                Poll::Ready(Err(e)) => return Poll::Ready(Err(error::Discover(e.into()))),
+                Poll::Pending => return Poll::Pending,
             }
         }
     }
@@ -137,18 +149,20 @@ where
         }
     }
 
-    fn poll_unready(&mut self) {
+    fn poll_unready(&mut self, cx: &mut Context<'_>) {
         loop {
-            match self.unready_services.poll() {
-                Ok(Async::NotReady) | Ok(Async::Ready(None)) => return,
-                Ok(Async::Ready(Some((key, svc)))) => {
+            match self.unready_services.poll_next_unpin(cx) {
+                Poll::Ready(None) | Poll::Pending => return,
+                Poll::Ready(Some(Ok((key, svc)))) => {
                     trace!("endpoint ready");
                     let _cancel = self.cancelations.remove(&key);
                     debug_assert!(_cancel.is_some(), "missing cancelation");
                     self.ready_services.insert(key, svc);
                 }
-                Err((key, Error::Canceled)) => debug_assert!(!self.cancelations.contains_key(&key)),
-                Err((key, Error::Inner(e))) => {
+                Poll::Ready(Some(Err((key, Error::Canceled)))) => {
+                    debug_assert!(!self.cancelations.contains_key(&key))
+                }
+                Poll::Ready(Some(Err((key, Error::Inner(e))))) => {
                     let error = e.into();
                     debug!({ %error }, "dropping failed endpoint");
                     let _cancel = self.cancelations.swap_remove(&key);
@@ -207,31 +221,35 @@ where
         svc.load()
     }
 
-    fn poll_ready_index_or_evict(&mut self, index: usize) -> Poll<(), ()> {
+    fn poll_ready_index_or_evict(
+        &mut self,
+        cx: &mut Context<'_>,
+        index: usize,
+    ) -> Poll<Result<(), ()>> {
         let (_, svc) = self
             .ready_services
             .get_index_mut(index)
             .expect("invalid index");
 
-        match svc.poll_ready() {
-            Ok(Async::Ready(())) => Ok(Async::Ready(())),
-            Ok(Async::NotReady) => {
+        match svc.poll_ready(cx) {
+            Poll::Ready(Ok(())) => Poll::Ready(Ok(())),
+            Poll::Pending => {
                 // became unready; so move it back there.
                 let (key, svc) = self
                     .ready_services
                     .swap_remove_index(index)
                     .expect("invalid ready index");
                 self.push_unready(key, svc);
-                Ok(Async::NotReady)
+                Poll::Pending
             }
-            Err(e) => {
+            Poll::Ready(Err(e)) => {
                 // failed, so drop it.
                 let error = e.into();
                 debug!({ %error }, "evicting failed endpoint");
                 self.ready_services
                     .swap_remove_index(index)
                     .expect("invalid ready index");
-                Err(())
+                Poll::Ready(Err(()))
             }
         }
     }
@@ -248,7 +266,7 @@ where
 {
     type Response = <D::Service as Service<Req>>::Response;
     type Error = error::Error;
-    type Future = future::MapErr<
+    type Future = try_future::MapErr<
         <D::Service as Service<Req>>::Future,
         fn(<D::Service as Service<Req>>::Error) -> error::Error,
     >;
@@ -257,13 +275,14 @@ where
     ///
     /// When `Async::Ready` is returned, `ready_index` is set with a valid index
     /// into `ready` referring to a `Service` that is ready to disptach a request.
-    fn poll_ready(&mut self) -> Poll<(), Self::Error> {
+    fn poll_ready(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
         // First and foremost, process discovery updates. This removes or updates a
         // previously-selected `ready_index` if appropriate.
-        self.poll_discover()?;
+        let _ = self.poll_discover(cx)?;
 
-        // Drive new or busy services to readiness.
-        self.poll_unready();
+        //Drive new or busy services to readiness.
+        self.poll_unready(cx);
+
         trace!({ nready = self.ready_services.len(), nunready = self.unready_services.len() }, "poll_ready");
 
         loop {
@@ -276,8 +295,8 @@ where
                 trace!({ next.idx = index }, "preselected ready_index");
                 debug_assert!(index < self.ready_services.len());
 
-                if let Ok(Async::Ready(())) = self.poll_ready_index_or_evict(index) {
-                    return Ok(Async::Ready(()));
+                if let Poll::Ready(Ok(())) = self.poll_ready_index_or_evict(cx, index) {
+                    return Poll::Ready(Ok(()));
                 }
 
                 self.next_ready_index = None;
@@ -286,7 +305,7 @@ where
             self.next_ready_index = self.p2c_next_ready_index();
             if self.next_ready_index.is_none() {
                 debug_assert!(self.ready_services.is_empty());
-                return Ok(Async::NotReady);
+                return Poll::Pending;
             }
         }
     }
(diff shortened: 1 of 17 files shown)


tower-balance/src/p2c/service.rs
```diff
@@ -38,14 +42,20 @@ pub struct Balance<D: Discover, Req> {
     rng: SmallRng,
 }
 
+impl<D: Discover, Req> Unpin for Balance<D, Req> {}
```
jonhoo: Probably not okay?

LucioFranco: @jonhoo discover doesn't take pin so it must box so this is safe since there is no unsafe code.

tower-balance/src/p2c/service.rs
```diff
@@ -137,18 +151,20 @@ where
         }
     }
 
-    fn poll_unready(&mut self) {
+    fn poll_unready(&mut self, cx: &mut Context<'_>) {
```
jonhoo: Pretty sure this'll need to take `Pin<&mut Self>`

jonhoo: Oh, I guess since `Service` is all `&mut self` for the time being, it doesn't?

tower-balance/src/p2c/service.rs
```diff
@@ -257,13 +277,14 @@ where
...
     /// When `Async::Ready` is returned, `ready_index` is set with a valid index
     /// into `ready` referring to a `Service` that is ready to disptach a request.
-    fn poll_ready(&mut self) -> Poll<(), Self::Error> {
+    fn poll_ready(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
         // First and foremost, process discovery updates. This removes or updates a
         // previously-selected `ready_index` if appropriate.
-        self.poll_discover()?;
+        let _ = self.poll_discover(cx)?;
+
+        //Drive new or busy services to readiness.
```
```suggestion
        // Drive new or busy services to readiness.
```

tower-balance/src/p2c/service.rs
```diff
@@ -276,8 +297,8 @@ where
                 trace!({ next.idx = index }, "preselected ready_index");
                 debug_assert!(index < self.ready_services.len());
 
-                if let Ok(Async::Ready(())) = self.poll_ready_index_or_evict(index) {
-                    return Ok(Async::Ready(()));
+                if let Poll::Ready(Ok(())) = self.poll_ready_index_or_evict(cx, index) {
```
jonhoo: Use `ready!`?

tower-balance/src/p2c/service.rs
```diff
@@ -307,25 +328,62 @@ where
 }
 
 impl<K, S: Service<Req>, Req> Future for UnreadyService<K, S, Req> {
-    type Item = (K, S);
-    type Error = (K, Error<S::Error>);
+    type Output = Result<(K, S), (K, Error<S::Error>)>;
```
jonhoo: Should this now be
```rust
type Output = (K, Result<S, Error<S::Error>>)
```
?

tower-balance/src/p2c/service.rs
LucioFranco: Yup, and it seems to work out pretty well!

tower-balance/src/p2c/service.rs
```diff
@@ -307,25 +328,62 @@ where
...
+impl<S, Req> Ready<S, Req> {
+    pub fn new(inner: S) -> Self {
+        Self {
+            inner: Some(inner),
+            _pd: PhantomData,
+        }
+    }
+}
+
+impl<S, Req> Unpin for Ready<S, Req> {}
```
jonhoo: Probably not okay.

tower-discover/src/lib.rs
```diff
@@ -34,7 +35,10 @@ pub trait Discover {
     type Error;
 
     /// Yields the next discovery change set.
-    fn poll(&mut self) -> Poll<Change<Self::Key, Self::Service>, Self::Error>;
+    fn poll(
```
jonhoo: We should probably be making all of these poll methods in traits `Pin` sooner rather than later.

tower-discover/src/list.rs
```diff
@@ -14,6 +14,8 @@ where
     inner: Enumerate<T::IntoIter>,
 }
 
+impl<T: IntoIterator> Unpin for ServiceList<T> {}
```
jonhoo: Probably not okay.

tower-load/Cargo.toml
```diff
@@ -23,11 +23,11 @@ edition = "2018"
 publish = false
 
 [dependencies]
-futures = "0.1.26"
-log = "0.4.1"
-tokio-timer = "0.2.4"
-tower-service = "0.2.0"
-tower-discover = "0.1.0"
+tracing = "0.1"
```
jonhoo: Should moving to `tracing` be done in its own PR?

tower-load/src/constant.rs
```diff
@@ -54,14 +54,21 @@ impl<D: Discover, M: Copy> Discover for Constant<D, M> {
...
+            Poll::Ready(Err(e)) => return Poll::Ready(Err(e)),
+            Poll::Pending => return Poll::Pending,
         };
 
-        Ok(Async::Ready(change))
+        Poll::Ready(Ok(change))
     }
 }
+
+impl<D, M> Unpin for Constant<D, M> {}
```
jonhoo: Probably not okay.

tower-load/src/instrument.rs
```diff
@@ -59,18 +56,23 @@ where
     }
 }
 
-impl<F, I, H> Future for InstrumentFuture<F, I, H>
+impl<F, I, H, T, E> Future for InstrumentFuture<F, I, H>
 where
-    F: Future,
-    I: Instrument<H, F::Item>,
+    F: Future<Output = Result<T, E>>,
+    I: Instrument<H, T>,
```
jonhoo: Why have the bounds changed so dramatically here?

tower-load/src/instrument.rs
```diff
@@ -59,18 +56,23 @@ where
...
+    type Output = Result<I::Output, E>;
+
+    fn poll(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+        let me = self.project();
 
-    fn poll(&mut self) -> Poll<Self::Item, Self::Error> {
-        let rsp = try_ready!(self.future.poll());
-        let h = self.handle.take().expect("handle");
-        Ok(self.instrument.instrument(h, rsp).into())
+        let rsp = match me.future.poll(cx) {
```
jonhoo: Use `ready!`?

tower-load/src/peak_ewma.rs
```diff
@@ -108,10 +108,13 @@ where
...
 
-    fn poll(&mut self) -> Poll<Change<D::Key, Self::Service>, D::Error> {
-        let change = match try_ready!(self.discover.poll()) {
-            Change::Remove(k) => Change::Remove(k),
-            Change::Insert(k, svc) => {
+    fn poll(
+        &mut self,
+        cx: &mut Context<'_>,
+    ) -> Poll<Result<Change<D::Key, Self::Service>, D::Error>> {
+        let change = match self.discover.poll(cx) {
```
jonhoo: Use `ready!`?

jonhoo: I think you can also use `?` here in fact.

tower-balance/src/p2c/service.rs
LucioFranco: So since poll_ready just takes a `&mut self` it must be Unpin therefore any !Unpin future within it must be Boxed etc. This makes this Unpin safe. So from what I understand if you have _no_ unsafe projections then impl Unpin is 100% safe.

tower-load/src/peak_ewma.rs
```diff
@@ -120,12 +123,16 @@ where
...
+            Poll::Ready(Err(e)) => return Poll::Ready(Err(e)),
+            Poll::Pending => return Poll::Pending,
         };
 
-        Ok(Async::Ready(change))
+        Poll::Ready(Ok(change))
     }
 }
 
+impl<D, I> Unpin for PeakEwmaDiscover<D, I> {}
```
jonhoo: Probably not okay

tower-load/src/peak_ewma.rs
```diff
@@ -291,118 +298,118 @@ fn nanos(d: Duration) -> f64 {
...
-        assert_eq!(super::nanos(Duration::new(0, 0)), 0.0);
-        assert_eq!(super::nanos(Duration::new(0, 123)), 123.0);
-        assert_eq!(super::nanos(Duration::new(1, 23)), 1_000_000_023.0);
-        assert_eq!(
-            super::nanos(Duration::new(::std::u64::MAX, 999_999_999)),
-            18446744074709553000.0
-        );
-    }
-}
+// #[cfg(test)]
```
jonhoo: Can we do this with a `/*` `*/` instead to avoid the large diff?

tower-load/src/pending_requests.rs
```diff
@@ -109,18 +109,27 @@ where
...
+    fn poll(
+        &mut self,
+        cx: &mut Context<'_>,
+    ) -> Poll<Result<Change<D::Key, Self::Service>, D::Error>> {
         use self::Change::*;
 
-        let change = match try_ready!(self.discover.poll()) {
-            Insert(k, svc) => Insert(k, PendingRequests::new(svc, self.instrument.clone())),
-            Remove(k) => Remove(k),
+        let change = match self.discover.poll(cx) {
```
jonhoo: `ready!` + `?` ?

tower-discover/src/lib.rs
LucioFranco: I am actually leaning to keeping this without Pin. We thought through many api's and in reality, it makes much more sense right now to keep pin out. So far I have yet to run into issues without Pin.

When I first converted this crate, I had pin but quickly found it to be pretty useless.

tower-load/src/pending_requests.rs
```diff
@@ -109,18 +109,27 @@ where
...
+            Poll::Ready(Err(e)) => return Poll::Ready(Err(e)),
+            Poll::Pending => return Poll::Pending,
         };
 
-        Ok(Async::Ready(change))
+        Poll::Ready(Ok(change))
     }
 }
 
+impl<D, I> Unpin for PendingRequestsDiscover<D, I> {}
```
jonhoo: Probably not okay

tower-load/src/pending_requests.rs
```diff
@@ -129,72 +138,72 @@ impl RefCount {
...
-        assert_eq!(svc.load(), Count(2));
-
-        drop(i1);
-        assert_eq!(svc.load(), Count(1));
-
-        drop(i0);
-        assert_eq!(svc.load(), Count(0));
-    }
-}
+// #[cfg(test)]
```
jonhoo: `/* */` plz

tower-discover/src/list.rs
LucioFranco: Same here, there is no unsafe thus this is 100% safe.

tower-load/Cargo.toml
LucioFranco: Yeah, I was just doing a drop in replacement.

@hawkw any preference how we approach this?

tower-balance/src/p2c/service.rs
jonhoo: Hmm, but then why do you need this `impl` at all?

tower-load/src/constant.rs
LucioFranco: Safe since no unsafe is used.

tower-balance/src/p2c/service.rs
jonhoo: Right, but then why do you need this `impl` at all?

tower-load/src/instrument.rs
LucioFranco: I am trying to avoid using `TryFuture` etc so in this case this replicates it via adding additional generics. This shouldn't affect the end user since F should ALWAYS allow T, E to be infered.

tower-load/src/instrument.rs
LucioFranco: I have been trying to avoid pulling in futures-util only for a small macro 🤷‍♂

tower-load/src/peak_ewma.rs
LucioFranco: This is safe since we have no unsafe projections.

LucioFranco:
@jonhoo thanks for the review, Ill update this PR hopefully later today. Now that tower-test is in I think I can fix up the tests.

As for the `impl Unpin` I believe all of them are safe in this case. From what our discussion yesterday, impl Unpin is unsound iff you use unsafe pin projections. If you use pin-project it will stop you from manually implementing Unpin. So in this case since all Service and Discover take &mut self and no Pin<&mut Self> the impl Unpin is 100% safe.

I am also trying to avoid bringing in futures-util for now since that is a heavy dep just for a small macro :shrug:

tower-balance/src/p2c/service.rs
LucioFranco: So I need the impl to deref Ready -> &mut Ready so that I can poll the inner type. Since S is always Unpin there are two ways to do this:
- safely impl Unpin
- S: Unpin so that the compiler can auto generate the same as above. 

so by adding impl Unpin in this case I am avoiding adding + Unpin everywhere pretty much.

taiki-e:
@LucioFranco 
> If you use pin-project it will stop you from manually implementing Unpin.

pin-project provides a way to implement this manually.
https://docs.rs/pin-project-internal/0.4.0-alpha.7/pin_project_internal/attr.pin_project.html#safety

>   To enforce this, this attribute will automatically generate an Unpin implementation for you, which will require that all structurally pinned fields be Unpin **If you wish to provide an manual Unpin impl, you can do so via the UnsafeUnpin argument.**

See [`UnsafeUnpin` trait](https://docs.rs/pin-project/0.4.0-alpha.7/pin_project/trait.UnsafeUnpin.html) for more details.

tower-load/src/instrument.rs
jonhoo: `ready!` is in `futures-core`

LucioFranco:
@taiki-e up! What im saying, and feel free to correct me, is unless I use UnsafeUnpin. Using just pin_project and pin will provide a safe api because it stops you from manually implementing Unpin where it could be unsound.

jonhoo:
@LucioFranco why do you want to implement `Unpin` in the first place?

Also, see my note above about how `ready!` is in `futures-core` :)

tower-balance/src/p2c/service.rs
jonhoo: I don't follow -- why is `S` always `Unpin`?

LucioFranco:
@jonhoo so the issue is this:

```rust
struct MyFuture<T>(T);

// This is safe because there is no unsafe usage in poll.
// IF WE USED unsafe unchecked then this would be unsound.
impl<T> Unpin for MyFuture<T> {}

impl<T: Service> Future for MyFuture<T> {
	type Output = T::Output;

	fn poll(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
	// This requires that we can get a &mut Self
	// one way is to use DerefMut via P::Target: Unpin (aka Self: Unpin)
	// the other way is to use unsafe Pin::new_unchecked.
	//
	// So the question is how do we ensure that _we_ are Unpin.
	// We can let the compiler infer it via auto trait which pretty much
	// adds this bound `impl<T: Unpin> Unpin for Self`
	// This would require us to add a `T: Service + Unpin` everywhere
	// which would bubble up to ALL the generics.
	//
	// Since we can safely impl Unpin since there is no unsafe required
	// we can just blanket impl Unpin for MyFuture safely and avoid
	// propogating Unpin everywhere.
	let fut = &mut self.0;

	// Since poll_ready just takes &mut T and not Pin<&mut T>,
	// we don't need to do any projections. We just need to get
	// &mut Self so that we can project self.0.
	fut.poll_ready(cx)
}
}
```

Ok jotted down some comments. Its a bit all over the place. I am happy to clean it up if it would be a good reference.

tower-balance/src/p2c/service.rs
LucioFranco: Technically speaking it is not, but since we only need a &mut T and it is never pinned in the first place, it acts as if it were Unpin always since we already have a &mut T. The thing with Unpin is that it says that Pin<&mut T> is == to &mut T iff T: Unpin. So thus any &mut T is technically already Unpinned in theory.

jonhoo:
Hmm, so the argument goes something like this I think:
 - `T` can't have been pinned _before_ we constructed `MyFuture`, because then you wouldn't be able to construct `MyFuture` in the first place (or `T` is `Unpin`)
 - _We_ never construct a `Pin` to `T`, so we never enter into the pin contract
 - Therefore, moving `MyFuture` is always fine, since we know a `Pin` to `T` cannot exist, and therefore moving `T` is fine.

I think my worry here is that this relies on `Pin` never being added to `Service`. If it did, we would either need to box, or issue a breaking change that removes the unconditional `impl Unpin`. In addition, should we ever introduce a `Pin` in this code, we would need to remember to also remove the `Unpin`, which is relatively distant in the code. I worry that we might later forget that, and thereby introduce a contract-violation.

Are we really sure that `Service` will never need `Pin`? That seems dubious to me, since, for example using `service_fn` with an `async fn` seems totally reasonable. The future passed to `service_fn` would be `!Unpin`, which means that it would only be able to poll the future if the service is pinned.

taiki-e:
> ```rust
> // This is safe because there is no unsafe usage in poll.
> // IF WE USED unsafe unchecked then this would be unsound.
> impl<T> Unpin for MyFuture<T> {}
> ```

pin-project generates this automatically if no field requires pinning.

> ```rust
>      let fut = &mut self.0;
> 
> 	// Since poll_ready just takes &mut T and not Pin<&mut T>,
> 	// we don't need to do any projections. We just need to get
> 	// &mut Self so that we can project self.0.
> 	fut.poll_ready(cx)
> ```

So, In this case, we can call `self.0.poll_ready(cx)` directly.

taiki-e:
> So, In this case, we can call `self.0.poll_ready(cx)` directly.

Also, if we try to rewrite this as follows in the future, we will get a compile error.
```rust
Pin::new(&mut self.0).poll_ready(cx)
```

```text
the trait bound `T: core::marker::Unpin` is not satisfied

the trait `core::marker::Unpin` is not implemented for `T`
```

jonhoo:
> pin-project generates this automatically if no field requires pinning.

Ah, neat! Then we should _definitely_ be using `pin-project` here.

olix0r:
I want to call out that this code is in-flux on master (https://github.com/tower-rs/tower/pull/303) -- i got pulled away from that PR for the past few weeks but was planning to come back to it now that some related changes have been made in Linkerd... I guess the v0.3.x branch will have to pick up these changes after they land?

tower-balance/src/p2c/service.rs
LucioFranco: I believe we want to own the service in the future then insert it as ready so we want to keep it in the item type.

tower-load/Cargo.toml
hawkw: I'm fine with either. I think if we want to replace the unstructured log lines with more idiomatic structured `tracing` diagnostics, _that_ should be done separately; I don't really have a problem with just switching the import in this PR though.

With that said, to avoid changing behavior, we should probably enable `tracing`'s `log` feature, so that we don't break `log` records for users who depend on them. We should probably follow the same strategy as `tower-balance` where we enable that conditionally on a default-on `log` feature flag on this crate, so that users who don't want to emit `log` records can disable it:
https://github.com/tower-rs/tower/blob/793e2e8e94b5ca6c1d12171b3909d78505ab6667/tower-buffer/Cargo.toml#L24-L26

LucioFranco:
Closing in favor of @jonhoo 's PRs

PR #320, opened 2019-09-06
Update tower-discover to std::future

jonhoo:


tower-discover/src/list.rs
```diff
@@ -31,15 +34,19 @@ where
 impl<T, U> Discover for ServiceList<T>
 where
     T: IntoIterator<Item = U>,
+    Enumerate<T::IntoIter>: Unpin,
```
LucioFranco: Why is Unpin required here?

jonhoo: https://github.com/taiki-e/pin-project/issues/76

jonhoo: https://github.com/taiki-e/pin-project/issues/76

PR #321, opened 2019-09-06
Update tower-load to std::future

jonhoo:
Will rebase and mark as ready once #320 lands (which this depends on)

### DIFF
--- a/tower-load/src/pending_requests.rs
+++ b/tower-load/src/pending_requests.rs
@@ -116,14 +116,14 @@ where
     type Error = D::Error;
 
     /// Yields the next discovery change set.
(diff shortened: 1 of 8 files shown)


tower-load/src/pending_requests.rs
```diff
@@ -132,25 +143,28 @@ impl RefCount {
...
         }
 
         fn call(&mut self, (): ()) -> Self::Future {
             future::ok(())
         }
     }
 
-    #[test]
-    fn default() {
+    #[tokio::test]
```
LucioFranco: I want to say using `tokio_test::block_on` might be better here?

PR #323, opened 2019-09-06
Update tower-buffer to std::future

jonhoo:
Will rebase and mark as ready once I finish the changes and #322 lands (which this depends on)

### DIFF
--- a/tower-buffer/src/future.rs
+++ b/tower-buffer/src/future.rs
@@ -5,6 +5,7 @@ use crate::{
     message,
 };
 use futures_core::ready;
+use pin_project::{pin_project, project};
 use std::{
     future::Future,
     pin::Pin,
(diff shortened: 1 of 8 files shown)


tower-buffer/src/future.rs
```diff
@@ -35,36 +36,34 @@ where
...
-    type Error = Error;
+    type Output = Result<T, Error>;
 
-    fn poll(&mut self) -> Poll<Self::Item, Self::Error> {
-        use self::ResponseState::*;
+    fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+        // we only ever give out Pin to what's inside state when it is ResponseState::Poll
+        // we only give it out once state has been set to ResponseState::Poll
+        // once state has been set to ResponseState::Poll, we never move it again
+        let this = unsafe { self.get_unchecked_mut() };
```
LucioFranco: why this over pin-project?

jonhoo: Because it was quite annoying to get `pin-project` to work through the enum-wrapped-in-a-newtype, especially given that we need to mutate `self.state` to go from polling the `rx` to polling the returned future (which is a move that `pin-project` wouldn't allow).

taiki-e: It is not enabled by default, but pin-project supports these patterns :)
e.g.:
* https://github.com/taiki-e/tower/commit/2fd18e4032cd5ee0ed33bb18df0a993839b9a0f8 (based on this branch)
* https://github.com/hyperium/hyper/blob/master/src/proto/h2/server.rs#L256-L317

[docs](https://docs.rs/pin-project-internal/0.4.0-alpha.10/pin_project_internal/attr.pin_project.html#enums)

jonhoo: Ah, I wasn't aware of the ability to call `set` too -- that's what I was missing. Will update!

PR #326, opened 2019-09-09
Update tower-retry to std::future

jonhoo:


tower-retry/src/lib.rs
```diff
@@ -37,15 +37,15 @@ impl<P, S> Retry<P, S> {
 
 impl<P, S, Request> Service<Request> for Retry<P, S>
 where
-    P: Policy<Request, S::Response, S::Error> + Clone,
+    P: Policy<Request, S::Response, S::Error> + Clone + Unpin,
```
jonhoo: This bound is needed because `Service` implicitly requires `Self: Unpin`, due to the `poll_ready`/`call` conflict.

tower-retry/src/policy.rs
```diff
@@ -46,7 +44,7 @@ use futures::Future;
 /// ```
 pub trait Policy<Req, Res, E>: Sized {
     /// The `Future` type returned by `Policy::retry()`.
-    type Future: Future<Item = Self, Error = ()>;
+    type Future: Future<Output = Option<Self>>;
```
jonhoo: Should we just make this `Output = Self`? We currently handle the case when the future "fails" in `src/future.rs`, but I don't know if that's still relevant? /cc @carllerche

LucioFranco: Yeah, make this `Self` 👍

tower-retry/src/lib.rs
LucioFranco: Don't you want just the Future of this to be `Unpin`?

jonhoo: It was basically because `ResponseFuture::poll` (which only has `Pin<&mut Self>`) needed to call `Retry::poll_ready`, which takes `&mut self`. That only works (without `unsafe`) if we make `Retry: Unpin`. I worked around it by assuming that `Retry::poll_ready` is really just calling `poll_ready` on the service, and thus all it needs is to get `&mut S`, which it can do since `S` is never pinned. It ain't pretty, but it works.

tower-retry/src/future.rs
```diff
@@ -1,27 +1,39 @@
...
+use futures_core::ready;
+use pin_project::{pin_project, project};
+use std::future::Future;
+use std::pin::Pin;
+use std::task::{Context, Poll};
 use tower_service::Service;
 
+// NOTE: this is the trait generated for Ready::project() by pin-project.
+// We need it here to be able to go "through" Ready to &mut Service without adding Unpin bounds.
+use crate::__RetryProjectionTrait;
```
taiki-e: Filed https://github.com/taiki-e/pin-project/issues/80 to resolve this.

PR #330, opened 2019-09-09
Update tower-util and tower to std::future

jonhoo:
This bumps tower-util and tower to 0.3.0-alpha.1.

One noteworthy change here is that `CallAll` now also has a `take_service` method that can be used through a `Pin`. Not sure whether it is necessary, but it was handy for tests.


tower-test/Cargo.toml
```diff
@@ -27,4 +27,4 @@ futures-executor-preview = "0.3.0-alpha.18"
 tokio-test = "0.2.0-alpha.2"
 tokio-sync = "0.2.0-alpha.2"
 tower-service = "0.3.0-alpha.1"
-pin-utils = "0.1.0-alpha.4"
\ No newline at end of file
+pin-utils = "0.1.0-alpha.4"
```
LucioFranco: btw pin_mut is avail in futures-util

tower-test/src/lib.rs
```diff
@@ -6,3 +6,21 @@
 
 mod macros;
 pub mod mock;
+
+use futures_util::future::{ready, Ready};
+use std::task::{Context, Poll};
+use tower_service::Service;
+
+/// A type for use in doctests.
+pub struct MyService;
```
LucioFranco: feature flag this?

LucioFranco: or I guess doc-tests should use something like service_fn?

tower-util/tests/service_fn.rs
```diff
@@ -1,13 +1,10 @@
...
+use futures_util::future::ready;
 
-#[test]
-fn simple() {
-    block_on(async {
-        let mut add_one = service_fn(|req| future::ok::<_, ()>(req + 1));
-        let answer = add_one.call(1).await.unwrap();
-        assert_eq!(answer, 2);
-    });
+#[tokio::test]
```
LucioFranco: block_on should work here?

tower/tests/builder.rs
```diff
@@ -1,69 +1,52 @@
...
 use tower::builder::ServiceBuilder;
+use tower::util::ServiceExt;
 use tower_buffer::BufferLayer;
 use tower_limit::{concurrency::ConcurrencyLimitLayer, rate::RateLimitLayer};
 use tower_retry::{Policy, RetryLayer};
 use tower_service::*;
-use void::Void;
+use tower_test::{mock};
+
+#[tokio::test]
```
LucioFranco: block_on should work though not sure if its preferred.

tower-test/src/lib.rs
jonhoo: Ah, to be clear, this is useful in situations like this: https://github.com/tower-rs/tower/pull/330/files#diff-9631c0644350438ce2d92778f80033d3R90
Where you really just want to show that _some_ service is passed in. I guess we could replace it with `service_fn`, but in these examples all that matters is that there is _some_ user-provided service, so the less noise the better.

tower/tests/builder.rs
jonhoo: It's actually pretty annoying to convert this to a non `#[tokio::test]`, since buffer ends up spawning a worker, which gets awkward in mock context (and for some reason also with `block_on`).

tower-test/Cargo.toml
jonhoo: This was there from before, but I've replaced it with `pin-project` now.

tower-test/src/lib.rs
jonhoo: So I _think_ I'd prefer to keep it the way it is. Why would you like to have it under a feature flag?

LucioFranco: The other option would be to do docs like

```rust
# async fn thing(svc: impl Service<()>) {
svc.call(()).await.unwrap();
# }
```

but im fine either way

PR #331, opened 2019-09-10
Update tower-filter to std::future

jonhoo:
This bumps tower-filter to 0.3.0-alpha.1

### DIFF
--- a/tower-filter/src/future.rs
+++ b/tower-filter/src/future.rs
@@ -72,9 +72,8 @@ where
                         .expect("we either give it back or leave State::Check once we take");
 
                     // Poll predicate
-                    match this.check.as_mut().poll(cx) {
-                        Poll::Ready(r) => {
-                            r?;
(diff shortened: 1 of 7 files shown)


tower-filter/src/future.rs
```diff
@@ -1,84 +1,91 @@
...
+                        .expect("we either give it back or leave State::Check once we take");
+
                     // Poll predicate
-                    match self.check.poll()? {
-                        Async::Ready(_) => {
-                            let response = self.service.call(request);
-                            self.state = WaitResponse(response);
+                    match this.check.as_mut().poll(cx) {
+                        Poll::Ready(r) => {
+                            r?;
```
LucioFranco: I think ist more preferable here to match also on the result like
```rust
Poll::Ready(Ok(..)) =>
Poll::Ready(Err(e)) =>
```
I personally find this a bit easier to read than using the `r?;`

PR #332, opened 2019-09-10
Update tower-spawn-ready to std::future

jonhoo:
Will rebase and mark as ready once #330 lands (which this depends on)

### DIFF
--- a/tower-spawn-ready/src/make.rs
+++ b/tower-spawn-ready/src/make.rs
@@ -1,5 +1,11 @@
 use crate::SpawnReady;
-use futures::{try_ready, Future, Poll};
+use futures_core::ready;
+use pin_project::pin_project;
+use std::{
+    future::Future,
+    pin::Pin,
+    task::{Context, Poll},
+};
 use tokio_executor::DefaultExecutor;
 use tower_service::Service;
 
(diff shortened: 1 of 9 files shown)


jonhoo:
Bah, forgot to mark this as a draft. Oh well. I wonder why GitHub won't let me mark a PR as draft after the fact...

tower-spawn-ready/src/make.rs
```diff
@@ -11,8 +17,10 @@ pub struct MakeSpawnReady<S, E = DefaultExecutor> {
 }
 
 /// Builds a SpawnReady with the result of an inner Future.
+#[pin_project]
```
LucioFranco: we should prob just drop the `E` since all the executor traits are in flux anyways.

jonhoo:
Removed the generic executor parameter in the latest commit.

tower-spawn-ready/src/service.rs
```diff
@@ -26,34 +25,20 @@ enum Inner<T> {
...
     }
 }
 
-impl<T, E, Request> Service<Request> for SpawnReady<T, E>
+impl<T, Request> Service<Request> for SpawnReady<T>
 where
     T: Service<Request> + Send,
     T::Error: Into<Error>,
-    E: BackgroundReadyExecutor<T, Request>,
+    DefaultExecutor: BackgroundReadyExecutor<T, Request>,
```
LucioFranco: don't think you need this?

jonhoo: You do, otherwise the call to `spawn` doesn't work (or you'd have to `Box::pin` the future).

Issue #319, opened 2019-09-03
Pinning and Service


jonhoo:
Now that we are moving to `std::future` (#279), we have to make a decision as to how this affects the `Service` trait. The current `Service` trait does not mention `Pin`:

```rust
pub trait Service<Request> {
    fn poll_ready(&mut self) -> Poll<(), Self::Error>;
    fn call(&mut self, req: Request) -> Self::Future;
}
```

Consider what happens in the common case where `poll_ready` wants to internally poll another future. To do so, it must have a `Pin` to that future. Since we do not pass `Pin<&mut Self>` to `poll_ready`, this means that the `Service` must internally box such futures (or otherwise cause them to be `Unpin`) so that `Pin::new` works. It can't use `Pin::new_unchecked`, since it cannot guarantee that the service itself isn't moved.

Effectively, this means that we are requiring `Service` to be `Unpin` (note that if `Service` doesn't have any internal futures to poll, then it will trivially be `Unpin`). This adds some additional "hidden" costs to each service, as they effectively have to box (or otherwise heap-allocate) any internal futures. Ignoring those costs, however, currently, that restriction is implicit in the type signature. We have a few options here as to how to "fix" this.

### Option A: keep everything as-is

Implementors of `Service` will have to discover that they must be `Unpin`, or will write unsound unsafe code to access fields (because again, `Pin::new_unchecked` is _not_ okay in this setting).

### Option B: add `Unpin` bound to `Service`

```rust
pub trait Service<Request>: Unpin { /* ... */ }
```

This makes it explicit that implementors of `Service` must implement `Unpin`, but does not _actually_ make it more difficult to implement `Service`. Adding this bound does have the added benefit that _hopefully_ implementors won't start playing with unsafe `Pin::new_unchecked` stuff as they will be forced to make their type implement `Unpin`.

### Option C: B, and also make `poll_ready` take `Pin`

```rust
pub trait Service<Request>: Unpin {
    fn poll_ready(self: Pin<&mut Self>) -> Poll<(), Self::Error>;
    fn call(&mut self, req: Request) -> Self::Future;
}
```

[...]

PR #333, opened 2019-09-10
Update tower-reconnect to std::future

jonhoo:
This bumps tower-reconnect to 0.3.0-alpha.1.
It also makes the tower-make version consistent with the 0.3.0 move.

### DIFF
--- a/tower-reconnect/src/lib.rs
+++ b/tower-reconnect/src/lib.rs
@@ -1,15 +1,19 @@
-#![doc(html_root_url = "https://docs.rs/tower-load-shed/0.1.0")]
+#![doc(html_root_url = "https://docs.rs/tower-reconnect/0.3.0-alpha.1")]
 #![deny(rust_2018_idioms)]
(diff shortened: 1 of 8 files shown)


tower-reconnect/src/lib.rs
```diff
@@ -52,26 +56,27 @@ impl<M, Target, S, Request> Service<Request> for Reconnect<M, Target>
 where
     M: Service<Target, Response = S>,
     S: Service<Request>,
+    M::Future: Unpin,
```
jonhoo: This bound is pretty sad, but needs to be there since `poll_ready` needs to poll that future, and only has `&mut self`. We _can_ `Box::pin` that future if the bound makes us sad. See also #319.

LucioFranco: So I think in this case we can provide a `MapBoxService` that boxes all the outgoing futures and can take some Servce that produces a !Unpin future and pin it. So that this choice can be done by the User. So in this case I'm 👍 on Unpin here.

PR #334, opened 2019-09-10
Update tower-hedge to std::future

jonhoo:
Will rebase and mark as ready once #331 lands (which this depends on)


tower-hedge/src/latency.rs
```diff
@@ -42,15 +49,15 @@ where
 impl<S, R, Request> Service<Request> for Latency<R, S>
 where
     S: Service<Request>,
-    S::Error: Into<super::Error>,
+    super::Error: From<S::Error>,
```
LucioFranco: why this change?

tower-hedge/src/latency.rs
```diff
@@ -62,24 +69,20 @@ where
...
 }
 
-impl<R, F> Future for ResponseFuture<R, F>
+impl<R, F, T, E> Future for ResponseFuture<R, F>
 where
     R: Record,
-    F: Future,
-    F::Error: Into<super::Error>,
+    F: Future<Output = Result<T, E>>,
+    super::Error: From<E>,
```
LucioFranco: same here?

tower-hedge/src/rotating_histogram.rs
```diff
@@ -1,8 +1,8 @@
 extern crate tokio_timer;
 
 use hdrhistogram::Histogram;
+use log::trace;
```
LucioFranco: tracing?

jonhoo: I want to change to `tracing` in a separate PR.

tower-hedge/src/latency.rs
jonhoo: Because for some reason, if I keep it as `Into<super::Error>`, I get a tonne of errors saying that the error does not implement `Send` and `Sync`. I'm not entirely sure why inverting the bound fixes the issue?

jonhoo:
```
note: required because of the requirements on the impl of `std::convert::From<E>` for `std::boxed::Box<dyn std::error::Error + std::marker::Send + std::marker::Sync>`
```

jonhoo: Actually, maybe @seanmonstar has some insight to provide here?

carllerche: Can we just use `Into<Box<dyn Error>>` or whatever?

jonhoo: That's what `super::Error` is:
https://github.com/tower-rs/tower/blob/e32a5ecd0cd24f5fabd139310ba854a8583ca2c4/tower-hedge/src/lib.rs#L50

seanmonstar: You're seeing the error down in `Future::poll` because of the `?` operator, which uses `From` instead of `Into`.

jonhoo: Thanks!

PR #335, opened 2019-09-10
Update tower-balance to std::future

jonhoo:
Will rebase and mark as ready once #330 lands (which this depends on)

This also includes a small patch to `tower-discover` to delegate the `Discover` trait through `Pin<D>` where `D::Target: Discover` (so we can use it through, say, `Pin<Box<>>`), and a patch to `tower-load` to make `Constant: Debug`.

### DIFF
--- a/tower-balance/src/lib.rs
+++ b/tower-balance/src/lib.rs
@@ -1,10 +1,10 @@
 //! Load balancing middlewares.
 
-#![doc(html_root_url = "https://docs.rs/tower-balance/0.1.0")]
+#![doc(html_root_url = "https://docs.rs/tower-balance/0.3.0-alpha.1")]
 #![deny(missing_docs)]
 #![deny(rust_2018_idioms)]
 #![allow(elided_lifetimes_in_paths)]
-#![deny(warnings)]
+//#![deny(warnings)]
 
 pub mod error;
 pub mod p2c;
--- a/tower-balance/src/lib.rs
+++ b/tower-balance/src/lib.rs
@@ -4,7 +4,7 @@
 #![deny(missing_docs)]
 #![deny(rust_2018_idioms)]
 #![allow(elided_lifetimes_in_paths)]
-//#![deny(warnings)]
+#![deny(warnings)]
 
 pub mod error;
 pub mod p2c;
--- a/tower-balance/src/p2c/service.rs
+++ b/tower-balance/src/p2c/service.rs
@@ -27,8 +27,15 @@ use tracing::{debug, trace};
 /// > The maximum load variance between any two servers is bound by `ln(ln(n))` where
 /// > `n` is the number of servers in the cluster.
 ///
+/// Note that `Balance` requires that the `Discover` you use is `Unpin` in order to implement
+/// `Service`. This is because it needs to be accessed from `Service::poll_ready`, which takes
+/// `&mut self`. You can achieve this easily by wrapping your `Discover` in [`Box::pin`] before you
+/// construct the `Balance` instance. For more details, see [#319].
+///
 /// [finagle]: https://twitter.github.io/finagle/guide/Clients.html#power-of-two-choices-p2c-least-loaded
 /// [p2c]: http://www.eecs.harvard.edu/~michaelm/postscripts/handbook2001.pdf
+/// [`Box::pin`]: https://doc.rust-lang.org/std/boxed/struct.Box.html#method.pin
+/// [#319]: https://github.com/tower-rs/tower/issues/319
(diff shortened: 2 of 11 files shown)


tower-balance/src/p2c/service.rs
```diff
@@ -88,7 +98,7 @@ where
 
 impl<D, Req> Balance<D, Req>
 where
-    D: Discover,
+    D: Discover + Unpin,
```
jonhoo: This makes me really sad, but it is required since we need to call `Discover::poll_discover` (which takes `Pin<&mut Self>`) from `Service::poll_ready` (which takes `&mut Self`).

jonhoo: See also #319

tower-balance/src/lib.rs
```diff
@@ -1,10 +1,10 @@
 //! Load balancing middlewares.
 
-#![doc(html_root_url = "https://docs.rs/tower-balance/0.1.0")]
+#![doc(html_root_url = "https://docs.rs/tower-balance/0.3.0-alpha.1")]
 #![deny(missing_docs)]
 #![deny(rust_2018_idioms)]
 #![allow(elided_lifetimes_in_paths)]
-#![deny(warnings)]
+//#![deny(warnings)]
```
LucioFranco: ?

tower-balance/src/p2c/service.rs
LucioFranco: Yeah, this is fine 👍 can we make sure we have docs on why this is required maybe, just point people to that issue and what they can do to get Unpin (aka box the discover)?

tower-balance/src/pool/mod.rs
```diff
@@ -267,7 +277,7 @@ impl Builder {
         };
 
         Pool {
-            balance: Balance::from_entropy(d),
+            balance: Balance::from_entropy(Box::pin(d)),
```
LucioFranco: This is unfortunate but this box should not happen that often :(

tower-balance/src/pool/mod.rs
```diff
@@ -316,48 +327,51 @@ where
     MS::Error: Into<error::Error>,
     Target: Clone,
 {
-    type Response = <Balance<PoolDiscoverer<MS, Target, Req>, Req> as Service<Req>>::Response;
-    type Error = <Balance<PoolDiscoverer<MS, Target, Req>, Req> as Service<Req>>::Error;
-    type Future = <Balance<PoolDiscoverer<MS, Target, Req>, Req> as Service<Req>>::Future;
+    type Response =
+        <Balance<Pin<Box<PoolDiscoverer<MS, Target, Req>>>, Req> as Service<Req>>::Response;
```
LucioFranco: a type alias might help here

tower-balance/src/pool/test.rs
```diff
@@ -1,285 +1,200 @@
...
-    with_task(|| {
-        assert_fut_ready!(fut2, "bar");
+    task::mock(|cx| {
+        // start the pool
+        let (mock, handle) =
+            mock::pair::<(), load::Constant<mock::Mock<(), &'static str>, usize>>();
+        pin_mut!(handle);
+
+        let mut pool = Builder::new()
+            .urgency(1.0) // so _any_ NotReady will add a service
```
LucioFranco:
```suggestion
            .urgency(1.0) // so _any_ Pending will add a service
```

tower-discover/src/lib.rs
```diff
@@ -43,6 +44,24 @@ pub trait Discover {
     ) -> Poll<Result<Change<Self::Key, Self::Service>, Self::Error>>;
 }
 
+// delegate through Pin
+impl<P> Discover for Pin<P>
```
LucioFranco: I believe we need to add a &mut T and Box<T> impls as well. This is what Future and Stream do.

tower-balance/src/lib.rs
jonhoo: Doing large refactors when warnings also cause errors makes the process a pain. Just failed to change it back :sweat_smile:

jonhoo: Separately, I _really_ dislike this `deny(warnings)` business, since it causes semver-compatible changes of dependencies to be breaking changes for us. For example, `IndexMap` has deprecated some methods, and that now causes `tower-balance` to no longer compile, even though our `Cargo.toml` hasn't changed.

tower-balance/src/pool/mod.rs
jonhoo: :sob:

PR #338, opened 2019-09-11
Update pin-project to 0.4.0-alpha.11

taiki-e:
There is a breaking change: https://github.com/taiki-e/pin-project/pull/86


LucioFranco:
@seanmonstar last thing I remember @jonhoo said that -alpha.x will not bump to the next by default. So technically its pinned without `=`?

seanmonstar:
Nope, I just tried in a new directory, settings `futures-core-preview = "0.3.0-alpha.16"`, and it downloaded `alpha.18`.

LucioFranco:
Then sounds like we need to update all the deps in all the tower crates then.

PR #340, opened 2019-09-11
Make Ready only take Service by reference

jonhoo:
Rather than consuming `self` and returning `(Self, _)`. This did mean
that a few crates that depended on `Ready` to own the `Service` and
provide it once it was ready had to change to call `poll_ready`
directly. Which in turn meant adding in some PhantomData<Request> so
that the impl blocks wouldn't be under-constrainted. Take, for example:

```rust
impl<K, S: Service<Req>, Req> Future for UnreadyService<K, S>
```

would fail to compile with

```
error[E0207]: the type parameter `Req` is not constrained by the impl trait, self type, or predicates
```

### DIFF
--- a/tower-spawn-ready/src/future.rs
+++ b/tower-spawn-ready/src/future.rs
@@ -3,6 +3,7 @@
 use crate::error::Error;
 use futures_core::ready;
 use pin_project::pin_project;
+use std::marker::PhantomData;
 use std::{
     future::Future,
     pin::Pin,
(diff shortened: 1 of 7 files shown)


tower-spawn-ready/src/future.rs
```diff
@@ -75,7 +72,9 @@ where
             return Poll::Ready(());
         }
 
-        let result = ready!(this.ready.poll(cx));
+        let result = ready!(this.service.as_mut().expect("illegal state").poll_ready(cx))
```
LucioFranco: this is uh very unfortunate..........

tower-util/src/ready.rs
```diff
@@ -13,28 +13,28 @@ use tower_service::Service;
 ///
 /// `Ready` values are produced by `ServiceExt::ready`.
 #[pin_project]
-pub struct Ready<T, Request> {
-    inner: Option<T>,
+pub struct Ready<'a, T, Request> {
+    inner: Option<&'a mut T>,
```
LucioFranco: does this need to be an option?

tower-spawn-ready/src/future.rs
jonhoo: That `expect` (like the ones below) is for `poll` being called again after `Poll::Ready` has been returned. If it is hit, the other `expect`s under would also be hit.

PR #342, opened 2019-09-13
change poll_next to poll_recv

xmclark:
This fixes #341 and also may fix the [build error](https://docs.rs/crate/tower/0.3.0-alpha.1/builds/184929) on docs.rs.

The error suggests using `poll_recv` instead, so I tried it and the minimum example compiles now.

PR #343, opened 2019-09-13
Prepare `tower-buffer` 0.3.0-alpha.1a release

### LINKED ISSUE #341: Compiler error build using 0.3.0-alpha.1
Trying to run a binary app with dependency on `build 0.3.0-alpha.1` on windows with `cargo 1.39.0-nightly (22f7dd049 2019-08-27)` has a compiler error:
```
error[E0599]: no method named `poll_next` found for type `std::pin::Pin<&mut tokio_sync::mpsc::bounded::Receiver<message::Message<Request, <T as tower_service::Service<Request>>::Future>>>` in the current s
cope
   --> C:\Users\Mac\.cargo\registry\src\github.com-1ecc6299db9ec823\tower-buffer-0.3.0-alpha.1\src\worker.rs:118:65
    |
118 |         while let Some(mut msg) = ready!(Pin::new(&mut self.rx).poll_next(cx)) {
    |                                                                 ^^^^^^^^^ help: there is a method with a similar name: `poll_recv`

error: aborting due to previous error
```

Built with this Cargo.toml:
```
[package]
name = "test-tower"
version = "0.1.0"
authors = ["Mackenzie Clark <mackenzie.a.z.c@gmail.com>"]
edition = "2018"

[dependencies]
tower = "0.3.0-alpha.1"
```

An empty main function:
```
fn main() {
}
```

LucioFranco:
Fixes #341 since for some reason `0.3.0-alpha.1` for buffer got released with a broken version.

### DIFF
--- a/tower-balance/Cargo.toml
+++ b/tower-balance/Cargo.toml
@@ -48,7 +48,7 @@ quickcheck = { version = "0.6", default-features = false }
 tokio = "=0.2.0-alpha.4"
 tokio-executor = "=0.2.0-alpha.4"
 tokio-test = "=0.2.0-alpha.4"
-tower-buffer = { version = "=0.3.0-alpha.1", path = "../tower-buffer" }
+tower-buffer = { version = "=0.3.0-alpha.1a", path = "../tower-buffer" }
 tower-limit = { version = "=0.3.0-alpha.1", path = "../tower-limit" }
 tower-test = { version = "=0.3.0-alpha.1", path = "../tower-test" }
-tower = { version = "=0.3.0-alpha.1", path = "../tower" }
+tower = { version = "=0.3.0-alpha.1a", path = "../tower" }
--- a/tower-buffer/CHANGELOG.md
+++ b/tower-buffer/CHANGELOG.md
@@ -1,3 +1,7 @@
+# 0.3.0-alpha.1a (September 13, 2019)
+
+- Fix `poll_next` not exisitng.
+
 # 0.3.0-alpha.1 (September 11, 2019)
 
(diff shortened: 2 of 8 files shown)


jonhoo:
Do we think this will actually fix the issue?

LucioFranco:
@jonhoo https://github.com/tower-rs/tower/pull/342 fixes it, though I don't know why this was not happening locally at all...but I've tested that PR and that change also compiles. I'm gonna have to assume this has to do with some weird feature flag magic.

LucioFranco:
so I think this one is good to go now that #342 is merged.

tower-balance/Cargo.toml
```diff
@@ -48,7 +48,7 @@ quickcheck = { version = "0.6", default-features = false }
 tokio = "=0.2.0-alpha.4"
 tokio-executor = "=0.2.0-alpha.4"
 tokio-test = "=0.2.0-alpha.4"
-tower-buffer = { version = "=0.3.0-alpha.1", path = "../tower-buffer" }
+tower-buffer = { version = "=0.3.0-alpha.1a", path = "../tower-buffer" }
```
carllerche: My vote is to just call this `-alpha.2`.

LucioFranco: Only reason I want to do `1a` is because I want people to know this will work with all `alpha.1` if we go to `alpha.2` then I need to release all the crates again or we will have a mix of 2 and 1.

LucioFranco: This isn't a breaking change either it is just a patch :shrug:

PR #345, opened 2019-09-13
Prepare `tower-buffer` 0.3.0-alpha.1b release

LucioFranco:
Since I messed up the last one!

### DIFF
--- a/tower-buffer/src/worker.rs
+++ b/tower-buffer/src/worker.rs
@@ -2,7 +2,7 @@ use crate::{
     error::{Closed, Error, ServiceError},
     message::Message,
 };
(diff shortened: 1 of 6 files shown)


tower-buffer/src/worker.rs
```diff
@@ -2,7 +2,7 @@ use crate::{
     error::{Closed, Error, ServiceError},
     message::Message,
 };
-use futures_core::{ready, stream::Stream};
```
jonhoo: Can we _please_ turn off `#![deny(warnings)]`?

LucioFranco: Yeah, I think all of them deny only in test cfg.

PR #347, opened 2019-09-17
Fix up broken dependencies and deprecated methods

lukesteensen:
I ran into a few issues getting everything to compile on master. The fixes were:

1. Follow the reorganization of `tracing-fmt` into `tracing_subscriber`
2. Update all crates to the same version of `'tokio-sync`
3. Stop calling deprecated method on `IndexMap` (should be a change in name only).

### DIFF
--- a/tower-balance/Cargo.toml
+++ b/tower-balance/Cargo.toml
@@ -41,7 +41,7 @@ tower-util = "0.1.0"
 slab = "0.4"
 
 [dev-dependencies]
-tracing-fmt = { git = "https://github.com/tokio-rs/tracing.git" }
+tracing-subscriber = { git = "https://github.com/tokio-rs/tracing.git" }
 hdrhistogram = "6.0"
 quickcheck = { version = "0.6", default-features = false }
 tokio = "0.1.7"
--- a/tower-balance/Cargo.toml
+++ b/tower-balance/Cargo.toml
@@ -41,7 +41,7 @@ tower-util = "0.1.0"
 slab = "0.4"
 
 [dev-dependencies]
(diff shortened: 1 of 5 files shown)


tower-balance/Cargo.toml
```diff
@@ -41,7 +41,7 @@ tower-util = "0.1.0"
 slab = "0.4"
 
 [dev-dependencies]
-tracing-fmt = { git = "https://github.com/tokio-rs/tracing.git" }
+tracing-subscriber = { git = "https://github.com/tokio-rs/tracing.git" }
```
LucioFranco:
```suggestion
tracing-subscriber = "0.1.3"
```

tower-balance/src/p2c/service.rs
```diff
@@ -132,7 +132,7 @@ where
                 .next_ready_index
                 .and_then(|i| Self::repair_index(i, idx, self.ready_services.len()));
             debug_assert!(!self.cancelations.contains_key(key));
-        } else if let Some(cancel) = self.cancelations.remove(key) {
+        } else if let Some(cancel) = self.cancelations.swap_remove(key) {
```
LucioFranco: does this have the same affect? Why this change?

lukesteensen: `remove` is deprecated, so our `#![deny(warnings)]` gives a compilation failure. According to the docs, `remove` was the same as `swap_remove`. I imagine they're just clarifying the naming.

LucioFranco: Yup just checked 👍

PR #352, opened 2019-09-23
Consistently apply deny/warn rules

jonhoo:
This makes all tower subcrates have the following lints as warn (rather than allow): `missing_docs`, `rust_2018_idioms`, and `missing_debug_implementations`. In addition, it consistently applies `deny(warning)` *only* under `#[cfg(test)]` so that deprecations and macro changes in minor version bumps in dependencies will never cause `tower` crates to stop compiling.

Note that `tower-reconnect` has the `missing_docs` lint disabled for now since it contained _no_ documentation previously. Also note that this patch does not add documentation to the various `new` methods, as they are considered self-explanatory. They are instead marked as`#[allow(missing_docs)]`.

### DIFF
--- a/tower-balance/src/lib.rs
+++ b/tower-balance/src/lib.rs
@@ -1,9 +1,12 @@
 //! Load balancing middlewares.
 
 #![doc(html_root_url = "https://docs.rs/tower-balance/0.3.0-alpha.1")]
-#![warn(missing_docs)]
-#![warn(rust_2018_idioms)]
-#![warn(missing_debug_implementations)]
+#![warn(
+    missing_debug_implementations,
+    missing_docs,
+    rust_2018_idioms,
+    unreachable_pub
+)]
 #![cfg_attr(test, deny(warnings))]
 #![allow(elided_lifetimes_in_paths)]
 
--- a/tower-balance/src/lib.rs
+++ b/tower-balance/src/lib.rs
@@ -7,7 +7,6 @@
     rust_2018_idioms,
     unreachable_pub
 )]
-#![cfg_attr(test, deny(warnings))]
 #![allow(elided_lifetimes_in_paths)]
 
 pub mod error;
--- a/tower-balance/src/lib.rs
+++ b/tower-balance/src/lib.rs
@@ -1,10 +1,11 @@
 //! Load balancing middlewares.
 
 #![doc(html_root_url = "https://docs.rs/tower-balance/0.3.0-alpha.1")]
-#![deny(missing_docs)]
-#![deny(rust_2018_idioms)]
+#![warn(missing_docs)]
+#![warn(rust_2018_idioms)]
+#![warn(missing_debug_implementations)]
+#![cfg_attr(test, deny(warnings))]
 #![allow(elided_lifetimes_in_paths)]
-#![deny(warnings)]
 
 pub mod error;
 pub mod p2c;
--- a/tower-balance/src/pool/mod.rs
+++ b/tower-balance/src/pool/mod.rs
@@ -20,6 +20,7 @@ use futures_core::ready;
 use pin_project::pin_project;
 use slab::Slab;
 use std::{
+    fmt,
     future::Future,
     pin::Pin,
     task::{Context, Poll},
@@ -42,9 +43,9 @@ enum Level {
     High,
 }
 
-#[pin_project]
 /// A wrapper around `MakeService` that discovers a new service when load is high, and removes a
 /// service when load is low. See [`Pool`].
+#[pin_project]
 pub struct PoolDiscoverer<MS, Target, Request>
 where
     MS: MakeService<Target, Request>,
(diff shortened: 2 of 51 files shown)


LucioFranco:
@jonhoo would you rather deny on test or on CI? I feel like CI may be better?

tower-balance/src/lib.rs
```diff
@@ -1,10 +1,11 @@
 //! Load balancing middlewares.
 
 #![doc(html_root_url = "https://docs.rs/tower-balance/0.3.0-alpha.1")]
-#![deny(missing_docs)]
-#![deny(rust_2018_idioms)]
+#![warn(missing_docs)]
```
LucioFranco: minor nit: can we combine this to one line?

tower-balance/src/pool/mod.rs
```diff
@@ -61,6 +62,24 @@ where
     limit: Option<usize>,
 }
 
+impl<MS, Target, Request> fmt::Debug for PoolDiscoverer<MS, Target, Request>
+where
+    MS: MakeService<Target, Request> + fmt::Debug,
+    MS::Future: fmt::Debug,
```
LucioFranco: why does this need to be debug?

tower-buffer/src/layer.rs
```diff
@@ -12,6 +12,7 @@ pub struct BufferLayer<Request, E = DefaultExecutor> {
 }
 
 impl<Request> BufferLayer<Request, DefaultExecutor> {
+    #[allow(missing_docs)]
```
LucioFranco: Why are we allowing this?

tower-balance/src/lib.rs
jonhoo: I personally prefer having them one per line to make it explicit which ones we're using. Scanning a line horizontally is a bit more painful. If you feel strongly I can merge them though.

tower-balance/src/pool/mod.rs
jonhoo: Because `PoolDiscover` internally holds a `MS::Future`, which is printed out for `Debug`.

tower-balance/src/lib.rs
LucioFranco: I don't really have a strong preference just seems to be the common pattern in tokio https://github.com/tokio-rs/tokio/blob/master/tokio/src/lib.rs#L2

Either way up to you!

tower-buffer/src/layer.rs
jonhoo: Because having docs on every `new` method that says "Creates a new Foo" seemed unnecessary, and I don't want to _also_ write expansive docs for all the methods in this patch :)

tower-timeout/src/lib.rs
```diff
@@ -1,7 +1,9 @@
 #![doc(html_root_url = "https://docs.rs/tower-timeout/0.3.0-alpha.1")]
-#![deny(missing_debug_implementations, missing_docs, rust_2018_idioms)]
-#![allow(elided_lifetimes_in_paths)]
+#![warn(missing_docs)]
+#![warn(rust_2018_idioms)]
+#![warn(missing_debug_implementations)]
 #![cfg_attr(test, deny(warnings))]
+#![allow(elided_lifetimes_in_paths)]
```
LucioFranco: I think this may be removed since this bug got fixed in rustc? (maybe)

tower/src/lib.rs
```diff
@@ -2,9 +2,11 @@
 // Allows refining features in the future without breaking backwards
 // compatibility
 #![cfg(feature = "full")]
-#![deny(missing_docs, missing_debug_implementations, rust_2018_idioms)]
-#![allow(elided_lifetimes_in_paths)]
+#![warn(missing_docs)]
+#![warn(rust_2018_idioms)]
+#![warn(missing_debug_implementations)]
 #![cfg_attr(test, deny(warnings))]
+#![allow(elided_lifetimes_in_paths)]
```
LucioFranco: same here possibly

tower-buffer/src/layer.rs
LucioFranco: works for me 😄

tower-balance/src/pool/mod.rs
LucioFranco: That seems incorrect though, I would assume that it wouldn't need to print the future? Not sure though.

jonhoo: So, the alternative here is https://github.com/tower-rs/tower/pull/352/commits/8c8dddf5fa575019ff33970742734f272ef6a751. Do you prefer it that way?

LucioFranco: Yeah, I think that looks great! Thanks <3

jonhoo:
Do we even have a ci feature? I'm happy for deny to be on `test`, since it means that the errors also appear locally during development.

LucioFranco:
@jonhoo we can do it via a env var. I actually think I prefer CI to deny since you may want to test code but may not have resolved all the warnings like docs.

PR #353, opened 2019-09-23
Re-enable all CI

jonhoo:


ci/azure-deploy-docs.yml
```diff
@@ -13,7 +13,7 @@ jobs:
   - template: azure-install-rust.yml
     parameters:
       platform: ${{parameters.name}}
-      rust_version: stable
+      rust: nightly
```
LucioFranco: Pin to 9-10?

jonhoo: Not sure what you mean by 9-10, but I've pinned to 2019-09-24.

jonhoo: Oh, :man_shrugging:, I don't think that's necessary. Do you still want us to pin?

LucioFranco: nah its fine 🚢 it

PR #357, opened 2019-10-01
Remove tokio alpha.6 patches

LucioFranco:


LucioFranco:
Thanks @taiki-e <3

taiki-e:
Oh, I think ci/patch.toml also has a patch.

LucioFranco:
@taiki-e good catch!

PR #362, opened 2019-10-29
Expect the poll_acquire error, not return

zyctree:
the error `AcquireError` poll_acquire returned only after semaphore closed,
which will never happen,

after this change, the wrapper of `Error` in tower-limit can be removed,
and it can use `S::Error` directly.


jonhoo:
So the argument here is that we never call `Semaphore::close`, and therefore we can never hit the error case there? If that is indeed the case, this looks good to me!

zyctree:
Right, just added more commit, the `Error` and `Future` in `Service` can be transparent

PR #365, opened 2019-11-16
add service layer to map from one request type to another

JeanMertz:
This came from a use-case discussed in https://github.com/timberio/vector/issues/1127#issuecomment-548507027.


tower-map/Cargo.toml
```diff
@@ -0,0 +1,32 @@
+[package]
+name = "tower-map"
```
LucioFranco: it probably makes sense to add this to `tower-util`, what do you think?

LucioFranco:
@JeanMertz hey! I am going to close this as I am going to migrate the master branch to use std::future::Future. Feel free to reopen against the proper branch.

PR #373, opened 2019-12-01
Update timeout, tower-test and reconnect

LucioFranco:
`tower-test` needs a better refactor to follow what we do in `tokio/tests`.

### DIFF
--- a/tower-test/Cargo.toml
+++ b/tower-test/Cargo.toml
@@ -8,7 +8,7 @@ name = "tower-test"
 #   - README.md
 # - Update CHANGELOG.md.
 # - Create "v0.1.x" git tag.
-version = "0.3.0-alpha.2"
+version = "0.3.0"
 authors = ["Tower Maintainers <team@tower-rs.com>"]
 license = "MIT"
 readme = "README.md"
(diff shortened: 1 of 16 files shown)


tower-test/Cargo.toml
```diff
@@ -22,9 +22,8 @@ categories = ["asynchronous", "network-programming"]
...
 [dependencies]
-futures-util-preview = "=0.3.0-alpha.19"
-futures-executor-preview = "=0.3.0-alpha.19"
-tokio-test = "=0.2.0-alpha.6"
-tokio-sync = "=0.2.0-alpha.6"
-tower-service = { version = "=0.3.0-alpha.2", path = "../tower-service" }
+futures-util = "0.3"
+tokio = { version = "0.2", features = ["sync"]}
+tokio-test = "0.2"
+tower-service = { version = "0.3", path = "../tower-service" }
```
jonhoo: I thought the idea was to only have `path` dependencies if you depend on unreleased changes?

tower-timeout/Cargo.toml
```diff
@@ -7,22 +7,22 @@ name = "tower-timeout"
...
-version = "0.3.0-alpha.2"
+# - Create "v0.3.x" git tag.
+version = "0.3.0"
 authors = ["Tower Maintainers <team@tower-rs.com>"]
 license = "MIT"
 readme = "README.md"
 repository = "https://github.com/tower-rs/tower"
 homepage = "https://github.com/tower-rs/tower"
-documentation = "https://docs.rs/tower-timeout/0.3.0-alpha.2"
+documentation = "https://docs.rs/tower-timeout/0.3.0"
```
jonhoo: We shouldn't need to set these any more. `docs.rs` already takes care of that.

jonhoo:
I wonder if `tower` should also go the route of `tokio` and bundle all these subcrates into one crate with features?

tower-test/Cargo.toml
LucioFranco: Not sure it works for now, if we run into issues we can remove 🤷‍♂

jonhoo: I think it causes some pain with people depending on git versions of the `tower-` crates, but I forget exactly how.

LucioFranco:
I am gonna merge this so follow up work can happen.

PR #375, opened 2019-12-04
Update tower-limit and prepare for release

LucioFranco:


tower-limit/Cargo.toml
```diff
@@ -8,28 +8,28 @@ name = "tower-limit"
...
 [dependencies]
-futures-core-preview = "=0.3.0-alpha.19"
-tower-service = { version = "=0.3.0-alpha.2", path = "../tower-service" }
-tower-layer = { version = "=0.3.0-alpha.2", path = "../tower-layer" }
-tokio-sync = "=0.2.0-alpha.6"
-tokio-timer = "=0.3.0-alpha.6"
+futures-core = "0.3"
+tower-service = { version = "0.3", path = "../tower-service" }
+tower-layer = { version = "0.3", path = "../tower-layer" }
+tokio-sync = "0.2.0-alpha.6"
```
jonhoo: Shouldn't this be just using the `sync` feature of `tokio`?

LucioFranco: @jonhoo sync currently does not have semaphore exposed. Carl has said he wants to rewrite it before we expose it :man_shrugging:

LucioFranco: cc @carllerche

tower-test/src/mock/mod.rs
```diff
@@ -2,35 +2,39 @@
...
+/// Spawn a layer onto a mock service.
+pub fn spawn_layer<T, U, L>(layer: L) -> (Spawn<L::Service>, Handle<T, U>)
 where
-    F: FnMut(&mut Context, &mut Pin<&mut Mock<T, U>>, &mut Pin<&mut Handle<T, U>>),
+    L: Layer<Mock<T, U>>,
 {
-    tokio_test::task::spawn(()).enter(|cx, _| {
-        let (mock, handle) = pair();
+    let (inner, handle) = pair();
+    let svc = layer.layer(inner);
```
jonhoo: Reading this just made me realize that the `layer` method really should be called `before`. Would have read so much nicer. Oh well :man_shrugging:

tower-limit/Cargo.toml
jonhoo: Ah! Hmm... Do we want a 0.3 that depends on an `alpha`? I guess maybe it doesn't matter too much since it's a relatively small dependency, and isn't publicly exposed?

tower-test/src/mock/mod.rs
LucioFranco: We can still change :P

tower-limit/Cargo.toml
LucioFranco: Yeah, its not public, and we can swap it out for futures-intrusive if we need. So I am punting :D

PR #377, opened 2019-12-04
Update buffer and prepare for release

LucioFranco:
Big change here is dropping executor support and defaulting layer to use `tokio::spawn` but also add a new `Buffer::pair` that will return the `Buffer` and `Worker` pair.

### DIFF
--- a/tower-buffer/src/service.rs
+++ b/tower-buffer/src/service.rs
@@ -1,14 +1,13 @@
 use crate::{
-    error::{Error, SpawnError},
+    error::Error,
     future::ResponseFuture,
     message::Message,
-    worker::{Handle, Worker, WorkerExecutor},
+    worker::{Handle, Worker},
 };
 
 use futures_core::ready;
 use std::task::{Context, Poll};
-use tokio_executor::DefaultExecutor;
-use tokio_sync::{mpsc, oneshot};
+use tokio::sync::{mpsc, oneshot};
 use tower_service::Service;
 
 /// Adds a buffer in front of an inner service.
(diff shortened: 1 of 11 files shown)


tower-buffer/src/service.rs
```diff
@@ -42,35 +41,30 @@ where
...
-    ///
-    /// `executor` is used to spawn a new `Worker` task that is dedicated to
-    /// draining the buffer and dispatching the requests to the internal
-    /// service.
+    /// Creates a new `Buffer` wrapping `service` but returns the background worker.
     ///
-    /// `bound` gives the maximal number of requests that can be queued for the service before
-    /// backpressure is applied to callers.
-    pub fn with_executor<E>(service: T, bound: usize, executor: &mut E) -> Self
+    /// This is useful if you do not want to spawn directly onto the tokio runtime
```
hawkw: nit: 
```suggestion
    /// This is useful if you do not want to spawn directly onto the `tokio` runtime
```

tower-buffer/src/service.rs
```diff
@@ -42,35 +41,30 @@ where
         T::Error: Send + Sync,
         Request: Send + 'static,
     {
-        Self::with_executor(service, bound, &mut DefaultExecutor::current())
+        let (tx, rx) = mpsc::channel(bound);
+        let (handle, worker) = Worker::new(service, rx);
+        tokio::spawn(worker);
+        Buffer { tx, handle }
```
jonhoo: Should this just call `pair` directly and then `spawn`?

LucioFranco: Yes probably :) we could use a bit more clean up before we release again.

PR #385, opened 2019-12-11
tower-util: reduce dependencies, make call-all optional

seanmonstar:


azure-pipelines.yml
```diff
@@ -18,7 +18,7 @@ jobs:
       # - tower-discover
```
LucioFranco: do we want to enable all of these again oops :D

Issue #244, opened 2019-04-09
Reconnect should not go back to idle if the MakeService::poll_ready errors


seanmonstar:
We've decided that if `poll_ready` returns an error, that means the service is dead. Reconnect currently will try again if its inner `MakeService::poll_ready` errors: https://github.com/tower-rs/tower/blob/0f58e50c3d86588c977316bf66a4ebeafde2a8e6/tower-reconnect/src/lib.rs#L93

seanmonstar:
Sorry, I was slightly wrong. The issue is that if the `MakeService::Future` polls into an error (fails to connect), `Reconnect` will return that error in its `poll_ready`, but still be ready to try again.

LucioFranco:
@seanmonstar I would imagine this error should be returned via the call, not via poll_ready?

seanmonstar:
Instead of blocking 0.1, we should just remove the `tower-reconnect` re-export from `tower`.

PR #386, opened 2019-12-11
reconnect: Fix maker error to return in future

### LINKED ISSUE #383: Service never attempts to reconnect
When running the helloworld client/server from [tonic](https://github.com/hyperium/tonic/tree/master/tonic-examples/src/helloworld) (with a modified client to run in a loop) I found that if a client ever has its connection broken to the server (due to me killing the server for some amount of time) that even if the server is brought back up and is healthy that the client is never able to reconnect and it appears this is due to the client never attempting to retry to establish a connection.

I would expect that the client should attempt to reestablish the connection if it ever has its connection with the server broken.

cc: @LucioFranco

LucioFranco:
Fixes #383

These changes reconnect to return the make service failure in the future instead of in poll_ready. This change allows us to _reuse_ the reconnect service on failed reconnects. This pushes the error down to the caller of `call` since technically the `make_service` is ready. This makes it so you can keep calling call until the service reconnects. 

This service is a bit odd, because it doesn't really fit in with the other types of services. This is because its an adapter to go from `MakeService` to `Service`. Thus, the `poll_ready` contract here doesn't mean the same as a regular `Service` because it has to bundle connecting.

This question brings up a big issue with how buffer is implemented, because if `poll_ready` returns the error then we must throw away the service, and buffer will just keep returning the failed connection error, even if we can successfully reestablish.

### DIFF
--- a/tower-reconnect/src/future.rs
+++ b/tower-reconnect/src/future.rs
@@ -1,5 +1,5 @@
 use crate::Error;
-use pin_project::pin_project;
+use pin_project::{pin_project, project};
 use std::{
     future::Future,
     pin::Pin,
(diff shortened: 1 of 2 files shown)


seanmonstar:
Hm, this is a significant behavior change, I'll need to think through its repercussions. This also seems related to #244.

LucioFranco:
Yeah, 100% related, funny how I came to the same conclusion again naturally https://github.com/tower-rs/tower/issues/244#issuecomment-481734842

@seanmonstar so I guess if we think about it this way, a service should be thrown away if _poll_ready_ returns an error, in this case its not returning an error thus our inner service is valid. Therefore, it makes sense to return a make_service error on call since we got that error from call and not from poll_ready.

LucioFranco:
@seanmonstar friendly ping

LucioFranco:
I am going to reopen this since we moved things around.

PR #388, opened 2019-12-11
limit: Vendor `tokio::sync::Semaphore`

LucioFranco:
The idea here is to just use a vendored private version of `Semaphore` to unblock us and to also remove any `alpha` deps.


LucioFranco:
@seanmonstar its straight up copied from tokio and has a commit sha, technically, its already been approved but this would unblock us and remove extra deps. Last I heard carl wanted to clean up the impl before exposing it publicly. This was the best solution I could think of.

PR #394, opened 2019-12-16
make and util: enable optional features for docs.rs

seanmonstar:


LucioFranco:
@seanmonstar can you bump the version to `0.3.1` for both and then you can release, 0.3 just went out.

jonhoo:
I'll close this following #432, though we should do the same thing for all the features that introduced!

PR #396, opened 2019-12-17
Add Future-like combinators

hlb8122:
**Motivation**
It is often the case that one wishes to take a service and modify it's output or input in some way without changing the behavior of `poll_ready()`. For example, turning a service returning `Result<(Parts, Body), _>` into a service returning `Result<Request<Body>, _>` or modifying the error type.

Currently, there is no ergonomic way to apply such changes without a fair amount of boilerplate.

**Proposal**
Extend `ServiceExt` trait to add various methods mirroring the combinators from `TryFutureExt` and various services and layers to accommodate this.

**Requirements**
Because combinators could be heavily chained together it's probably wise to trim all the fat off here and if possible make it zero-cost when compared to simply applying the combinators inside of the inner service's call.

**Result**
Having such methods should allow a more expressive algebra of tools to construct and combine services.

**Current Implementation**
At the time of writing I've introduced four combinators:
```rust
    fn map_ok<F, Response>(self, f: F) -> MapOk<Self, F>
    where
        Self: Sized,
        F: FnOnce(Self::Response) -> Response + Clone,
    {
        MapOk::new(self, f)
    }

    fn map_err<F, Error>(self, f: F) -> MapErr<Self, F>
    where
        Self: Sized,
        F: FnOnce(Self::Error) -> Error + Clone,
    {
        MapErr::new(self, f)
    }

    fn with<F, NewRequest>(self, f: F) -> With<Self, F>
    where
        Self: Sized,
        F: FnOnce(NewRequest) -> Request + Clone,
    {
        With::new(self, f)
    }

    fn try_with<F, NewRequest>(self, f: F) -> With<Self, F>
    where
        Self: Sized,
        F: FnOnce(NewRequest) -> Result<Request, Self::Error> + Clone,
    {
        TryWith::new(self, f)
    }
```
and an associated `Service` and `Layer` for each.

**TODO**

- [x] Which trait constraints are minimally required on the closures to ensure an ergonomic API without any excess overhead?
- [x] Documentation
- [ ] More combinators such as and_then


tower-util/src/combinators/map_err.rs
```diff
@@ -0,0 +1,107 @@
...
+        fn poll_ready(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
+            Poll::Ready(Ok(()))
+        }
+
+        fn call(&mut self, request: String) -> Self::Future {
+            futures_util::future::ready(Ok(request))
+        }
+    }
+
+    #[test]
```
hlb8122: This should be removed later. 

The purpose of these tests are to demonstrate how the API behaves.

tower-util/src/combinators/map_ok.rs
```diff
@@ -0,0 +1,109 @@
...
+
+    fn layer(&self, inner: S) -> Self::Service {
+        MapOk {
+            f: self.f.clone(),
+            inner,
+        }
+    }
+}
+
+#[cfg(test)]
```
hlb8122: https://github.com/tower-rs/tower/pull/396/files#r358888802

tower-util/src/combinators/with.rs
```diff
@@ -0,0 +1,107 @@
...
+
+    fn layer(&self, inner: S) -> Self::Service {
+        With {
+            f: self.f.clone(),
+            inner,
+        }
+    }
+}
+
+#[cfg(test)]
```
hlb8122: https://github.com/tower-rs/tower/pull/396/files#r358888802

tower-util/src/lib.rs
```diff
@@ -76,6 +78,30 @@ pub trait ServiceExt<Request>: tower_service::Service<Request> {
...
+
+    fn map_err<F, Error>(self, f: F) -> MapErr<Self, F>
+    where
+        Self: Sized,
+        F: FnOnce(Self::Error) -> Error + Clone,
+    {
+        MapErr::new(self, f)
+    }
+
+    fn with<F, NewRequest>(self, f: F) -> With<Self, F>
```
hlb8122: The name `with` here is pulled from mirrored from here https://docs.rs/futures/0.3.1/futures/sink/trait.SinkExt.html#method.with

> Composes a function in front of the sink.
> 
> This adapter produces a new sink that passes each value through the given function f before sending it to self.

tower-util/src/combinators/map_err.rs
davidbarsky: Could these tests be made doc comments?

hlb8122: I've just added doc tests for `map_ok`, `map_err`, and `with`. Would you like more on the underlying services/layers too?

tower-util/src/combinators/with.rs
```diff
@@ -0,0 +1,70 @@
...
+    f: F,
+}
+
+impl<S, F> With<S, F> {
+    pub fn new(inner: S, f: F) -> Self {
+        With { inner, f }
+    }
+}
+
+impl<S, F, NewRequest, OldRequest> Service<NewRequest> for With<S, F>
```
hlb8122: Currently, `With` service only support infallible mapping of the request to the new request - that is, `NewRequest -> OldRequest`.

This could be extended to, or complimented with a `TryWith`(?) service where `NewRequest -> Result<OldRequest, Error>` where `Error` is either `Into<ServiceError>` or `Error=ServiceError`.

tower-util/src/combinators/map_err.rs
davidbarsky: Thanks for adding them! No, I think this is fine for now.

(Tower lacks good documentation, so we've got to start somewhere...)

LucioFranco:
This looks like a fantastic start! I would like to review this once we get the 0.3 release out since we can always add more items in follow up patches.

hlb8122:
Has there been any decisions made with respect to this? I noticed `map_request` has been implemented in https://github.com/tower-rs/tower/pull/435.

tower-util/src/combinators/map_err.rs
```diff
@@ -0,0 +1,72 @@
...
+    fn call(&mut self, request: Request) -> Self::Future {
+        self.inner.call(request).map_err(self.f.clone())
+    }
+}
+
+#[derive(Debug)]
+pub struct MapErrLayer<F, Request, Error> {
+    f: F,
+    request: PhantomData<Request>,
+    error: PhantomData<Error>,
```
hawkw, line 42: nit: a few notes on the use of `PhantomData`:

1. let's use `PhantomData<fn(T)>` rather than `PhantomData<T>`, so that the type is not considered to contain a field of type `T`. that way, if `T` is not `Send` or `Sync`, the containing type can still implement those traits (since it doesn't actually contain a `T`).
2. since these fields are unused, the compiler will warn if we don't prefix them with `_` --- let's do that.
3. i'd prefer to collapse these into one `PhantomData` field that contains both the `Request` and `Error` types --- multiple `PhantomData`s just adds noise IMO.

so, in sum, i'd prefer something like this:
```suggestion
    _p: PhantomData<fn(Request, Error)>,
```

tower-util/src/combinators/map_err.rs
```diff
@@ -0,0 +1,72 @@
...
+
+impl<F, Request, Error> MapErrLayer<F, Request, Error>
+where
+    F: Clone,
+{
+    pub fn new(f: F) -> Self {
+        MapErrLayer {
+            f,
+            request: PhantomData::default(),
+            error: PhantomData::default(),
```
hawkw, line 53: nit: this can just be
```suggestion
            request: PhantomData,
            error: PhantomData,
```

tower-util/src/combinators/map_err.rs
```diff
@@ -0,0 +1,72 @@
...
+    pub fn new(inner: S, f: F) -> Self {
+        MapErr { f, inner }
+    }
+}
+
+impl<S, F, Request, Error> Service<Request> for MapErr<S, F>
+where
+    S: Service<Request>,
+    F: FnOnce(S::Error) -> Error,
+    F: Clone,
```
hawkw, line 23: Using `FnOnce + Clone` here seemed a little weird at first, but then I realized that the requirement that the function be moved into the `MapErr` future and cloned in `Layer::layer` kind of drives this. So if we're going to have to require `Clone`, it makes sense to use `FnOnce`.

tower-util/src/combinators/with.rs
```diff
@@ -0,0 +1,70 @@
+use std::marker::PhantomData;
+use std::task::{Context, Poll};
+use tower_layer::Layer;
+use tower_service::Service;
+
+pub struct With<S, F> {
```
hawkw, line 6: `tower-http` contains a [`tower-request-modifier` crate](https://github.com/tower-rs/tower-http/tree/master/tower-request-modifier) that does a bunch of HTTP-specific request mutations...I wonder if that could be reimplemented on top of this API, eventually? It would be nice to keep that in mind and make sure this is possible.

tower-util/src/combinators/with.rs
hawkw: I don't think there's an easy way to have this service implement both fallible and infallible mapping, so adding a separate `TryWith` sounds good to me.

tower-util/src/combinators/with.rs
```diff
@@ -0,0 +1,70 @@
...
+impl<S, F> With<S, F> {
+    pub fn new(inner: S, f: F) -> Self {
+        With { inner, f }
+    }
+}
+
+impl<S, F, NewRequest, OldRequest> Service<NewRequest> for With<S, F>
+where
+    S: Service<OldRequest>,
+    F: FnOnce(NewRequest) -> OldRequest,
```
hawkw, line 20: Nit: This naming feels a little backwards to me. Since this service is called _before_ the inner service, it seems like this service's `Request` type is the "old" request, and the returned request is the "new" request. I could be convinced otherwise.

tower-util/src/combinators/with.rs
```diff
@@ -0,0 +1,70 @@
...
+    fn call(&mut self, request: NewRequest) -> S::Future {
+        self.inner.call((self.f.clone())(request))
+    }
+}
+
+#[derive(Debug)]
+pub struct WithLayer<F, OldRequest, NewRequest> {
+    f: F,
+    old_request: PhantomData<OldRequest>,
+    new_request: PhantomData<NewRequest>,
```
hawkw, line 40: nit/TIOLI: I think we wouldn't need any `PhantomData` here if we just changed this to 
```suggestion
pub struct WithLayer<F, OldRequest, NewRequest>
where
   F: FnOnce(NewRequest) -> OldRequest,
{
    f: F,
```
but then we would need to put this bound on all impls for this type...

hawkw, line 40: nit: alternatively, if we are going to keep these `PhantomData`s, we should apply the same changes I suggested above.

tower-util/src/combinators/with.rs
```diff
@@ -0,0 +1,70 @@
...
+
+impl<F, OldRequest, NewRequest> WithLayer<F, OldRequest, NewRequest>
+where
+    F: Clone,
+{
+    pub fn new(f: F) -> Self {
+        WithLayer {
+            f,
+            old_request: PhantomData::default(),
+            new_request: PhantomData::default(),
```
hawkw, line 51: nit:
```suggestion
            old_request: PhantomData,
            new_request: PhantomData,
```

tower-util/src/lib.rs
```diff
@@ -76,6 +78,161 @@ pub trait ServiceExt<Request>: tower_service::Service<Request> {
     {
         CallAll::new(self, reqs)
     }
+
+    /// # Example
```
hawkw, line 82: Can we add a summary of the method, as well as an example?

tower-util/src/lib.rs
```diff
@@ -76,6 +78,161 @@ pub trait ServiceExt<Request>: tower_service::Service<Request> {
...
+    /// ```
+    fn map_ok<F, Response>(self, f: F) -> MapOk<Self, F>
+    where
+        Self: Sized,
+        F: FnOnce(Self::Response) -> Response + Clone,
+    {
+        MapOk::new(self, f)
+    }
+
+    /// # Example
```
hawkw, line 135: As above, would be nice to have a description of the method here.

tower-util/src/lib.rs
```diff
@@ -76,6 +78,161 @@ pub trait ServiceExt<Request>: tower_service::Service<Request> {
...
+    /// ```
+    fn map_err<F, Error>(self, f: F) -> MapErr<Self, F>
+    where
+        Self: Sized,
+        F: FnOnce(Self::Error) -> Error + Clone,
+    {
+        MapErr::new(self, f)
+    }
+
+    /// # Example
```
hawkw, line 188: And again, a one-two sentence summary would be good.

tower-util/src/lib.rs
hawkw: I'm a little on the fence about naming...I agree that consistency with the `futures` crate is nice, but I don't find `with` to be super descriptive. To me, it sounds more like `layer`-like composition of two `Service`s. I _might_ consider erring on the side of consistency with the rest of _our_ API and calling this `MapRequest` or `MapReq`.

@LucioFranco what do you think?

hlb8122: Not strictly a defense of `with`, but if we were to rename this `map_req` then https://github.com/tower-rs/tower/pull/396#discussion_r446291692 would imply the existence of a `try_map_req` (or something like that?), which is could be more of a mouthful than `try_with`.

tower-util/src/lib.rs
hlb8122: For both of these I've almost copy + pasted the summaries from the futures crate.

tower-util/src/combinators/with.rs
hlb8122: I'll give https://github.com/tower-rs/tower-http/tree/master/tower-request-modifier a fork and try it out when I get some spare time.

tower-util/src/combinators/with.rs
hlb8122: I see, my rationale here was that `S` is your "old" service and `With<S, F>` is your "new" service and hence `NewRequest` is the request for the "new" service.

tower-util/src/combinators/with.rs
hlb8122: I remember noticing this at the time but opted for the consistency of having the same layout of the phantom data on all these combinators and fewer impls.

tower-util/src/combinators/with.rs
hlb8122: It's added with `type Future = Either<S::Service, Ready<Result<S::Response, S::Error>>>`.

tower-util/src/combinators/with.rs
```diff
@@ -0,0 +1,70 @@
...
+    pub fn new(inner: S, f: F) -> Self {
+        With { inner, f }
+    }
+}
+
+impl<S, F, NewRequest, OldRequest> Service<NewRequest> for With<S, F>
+where
+    S: Service<OldRequest>,
+    F: FnOnce(NewRequest) -> OldRequest,
+    F: Clone,
```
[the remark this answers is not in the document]
hlb8122: My thought process here was that, every `F` needs to be `Clone` as a result of `layer(&self, inner)` and then the weakest bound on `F` is `FnOnce`.

tower/src/util/combinators/map_err.rs
```diff
@@ -0,0 +1,64 @@
+use futures_util::{future::MapErr as MapErrFut, TryFutureExt};
+use std::marker::PhantomData;
+use std::task::{Context, Poll};
+use tower_layer::Layer;
+use tower_service::Service;
+
+#[derive(Debug)]
+pub struct MapErr<S, F> {
```
hawkw, line 8: It would be nice to have a brief doc comment on this struct --- it could just be 
```suggestion
/// Service returned by the [`map_err`] combinator.
///
/// [`map_err`]: ../../trait.ServiceExt.html#method.map_err
pub struct MapErr<S, F> {
```

tower/src/util/combinators/map_err.rs
```diff
@@ -0,0 +1,64 @@
...
+    inner: S,
+    f: F,
+}
+
+impl<S, F> MapErr<S, F> {
+    pub fn new(inner: S, f: F) -> Self {
```
hawkw, line 14:
```suggestion
    /// Returns a new `MapErr` service.
    pub fn new(inner: S, f: F) -> Self {
```

tower/src/util/combinators/map_err.rs
```diff
@@ -0,0 +1,64 @@
...
+        self.inner.poll_ready(cx).map_err(self.f.clone())
+    }
+
+    fn call(&mut self, request: Request) -> Self::Future {
+        self.inner.call(request).map_err(self.f.clone())
+    }
+}
+
+#[derive(Debug)]
+pub struct MapErrLayer<F, Request, Error> {
```
hawkw, line 39: Similarly:
```suggestion
/// A [`Layer`] that produces a [`MapErr`] service.
///
/// [`Layer`]: ../../../trait.Layer.html
/// [`MapErr`]: struct.MapErr.html
#[derive(Debug)]
pub struct MapErrLayer<F, Request, Error> {
```

tower/src/util/combinators/map_err.rs
```diff
@@ -0,0 +1,64 @@
...
+    f: F,
+    _p: PhantomData<fn(Request, Error)>,
+}
+
+impl<F, Request, Error> MapErrLayer<F, Request, Error> {
+    pub fn new(f: F) -> Self {
```
hawkw, line 45:
```suggestion
    /// Returns a new `MapErr` layer.
    pub fn new(f: F) -> Self {
```

tower/src/util/combinators/map_ok.rs
```diff
@@ -0,0 +1,64 @@
+use futures_util::{future::MapOk as MapOkFut, TryFutureExt};
+use std::marker::PhantomData;
+use std::task::{Context, Poll};
+use tower_layer::Layer;
+use tower_service::Service;
+
+#[derive(Debug)]
+pub struct MapOk<S, F> {
```
hawkw, line 8: Similarly, let's add some quick doc comments here so that CI stops whining about it.

tower/src/util/combinators/try_with.rs
```diff
@@ -0,0 +1,67 @@
+use futures_util::future::{ready, Either, Ready};
+use std::marker::PhantomData;
+use std::task::{Context, Poll};
+use tower_layer::Layer;
+use tower_service::Service;
+
+#[derive(Debug)]
+pub struct TryWith<S, F> {
```
hawkw, line 8: And let's add doc comments here as well.

tower/src/util/combinators/with.rs
```diff
@@ -0,0 +1,63 @@
+use std::marker::PhantomData;
+use std::task::{Context, Poll};
+use tower_layer::Layer;
+use tower_service::Service;
+
+#[derive(Debug)]
+pub struct With<S, F> {
```
hawkw, line 7: ...and here (even though CI isn't warning about it for some reason? are the types in this file not being exported publicly?)

tower/src/util/mod.rs
```diff
@@ -82,6 +85,240 @@ pub trait ServiceExt<Request>: tower_service::Service<Request> {
...
     }
+
+    /// Maps this service's response value to a different value. This does not
+    /// alter the behaviour of the [`poll_ready`](trait.Service.html#tymethod.poll_ready)
+    /// method.
+    ///
+    /// This method can be used to change the [`Response`](trait.Service.html#associatedtype.Response)
+    /// type of the service into a different type. It is similar to the [`Result::map_err`]
+    /// method. You can use this method to chain along a computation once the
+    /// services response has been resolved.
```
hawkw, line 96: These docs are lovely, thank you! <3

tower/src/util/mod.rs
```diff
@@ -82,6 +85,240 @@ pub trait ServiceExt<Request>: tower_service::Service<Request> {
...
+    /// fn main() {
+    ///     // A service returning Result<_, Error>
+    ///     let service = DatabaseService::new("127.0.0.1:8080");
+    ///
+    ///     // Map the error into a new error
+    ///     let mut new_service = service.map_err(|err| err.code);
+    ///
+    ///     async {
+    ///         let id = 13;
+    ///         let code = new_service.call(id).await.unwrap_err();
```
hawkw, line 199: It looks like this doctest [doesn't compile](https://github.com/tower-rs/tower/pull/396/checks?check_run_id=816022386#step:4:494) because of issues with type inference in async blocks/functions. I think if we add an explicit
```suggestion
    ///         let id: u32 = 13;
    ///         let code = new_service.call(id).await.unwrap_err();
```
type annotation it will be okay.

tower/src/util/mod.rs
```diff
@@ -82,6 +85,240 @@ pub trait ServiceExt<Request>: tower_service::Service<Request> {
...
+    /// #
+    /// fn main() {
+    ///     // A service taking an u32 as a request
+    ///     let service = DatabaseService::new("127.0.0.1:8080");
+    ///
+    ///     // Map the request into a new request
+    ///     let mut new_service = service.with(|id_str: &str| id_str.parse().unwrap());
+    ///
+    ///     async {
+    ///         let id = "13";
```
hawkw, line 252: similarly, i think we need a `u32` type annotation. also, it looks like this is a string literal?
```suggestion
    ///         let id: u32 = 13;
```

tower/src/util/mod.rs
```diff
@@ -82,6 +85,240 @@ pub trait ServiceExt<Request>: tower_service::Service<Request> {
...
+    /// #
+    /// fn main() {
+    ///     // A service taking an u32 as a request
+    ///     let service = DatabaseService::new("127.0.0.1:8080");
+    ///
+    ///     // Map the request into a new request fallibly
+    ///     let mut new_service = service.with(|id_str: &str| id_str.parse().map_err(DbError::Parse));
+    ///
+    ///     async {
+    ///         let id = "13";
```
hawkw, line 310:
```suggestion
    ///         let id = "13";
```
```suggestion
    ///         let id: u32 = 13;
```

tower/src/util/mod.rs
```diff
@@ -82,6 +85,240 @@ pub trait ServiceExt<Request>: tower_service::Service<Request> {
...
+    ///     // A service returning Result<_, Error>
+    ///     let service = DatabaseService::new("127.0.0.1:8080");
+    ///
+    ///     // Map the error into a new error
+    ///     let mut new_service = service.map_err(|err| err.code);
+    ///
+    ///     async {
+    ///         let id = 13;
+    ///         let code = new_service.call(id).await.unwrap_err();
+    ///     };
```
hawkw, line 200: It might be nice to show the async block actually being used in the example...I don't want us to accidentally give readers the impression that typing `async { ... }` will actually run their code on its own. 

Maybe we could hide the `fn main()` and just show an async function instead as the example, so we don't need to worry about that?

hlb8122:
Quick note about the documentation: I've used [intra-doc links](https://github.com/rust-lang/rfcs/blob/master/text/1946-intra-rustdoc-links.md) here which are only supported on nightly (and the docs.rs compiler).

The reason for this is that the of `#[doc(inline)]` on `ServiceExt` in `lib.rs` breaks the relative links.

The same problem exists for the `CallAll` link in the documentation of `ServiceExt::call_all` method from the `/doc/tower/trait.ServiceExt.html` page (rather than `/doc/tower/util/trait.ServiceExt.html`)

tower-util/src/combinators/with.rs
hlb8122: https://github.com/tower-rs/tower-http/pull/31

I departed from the previous API by a fair bit but I think it turned out ok.

tower/src/util/combinators/map_err.rs
```diff
@@ -0,0 +1,70 @@
...
+    /// Creates a new [`MapErr`] service.
+    pub fn new(inner: S, f: F) -> Self {
+        MapErr { f, inner }
+    }
+}
+
+impl<S, F, Request, Error> Service<Request> for MapErr<S, F>
+where
+    S: Service<Request>,
+    F: FnOnce(S::Error) -> Error + Clone,
```
LucioFranco, line 26: Why clone? Why not `Fn(S::Error) -> Error`? I believe `Fn` is semantically similar to `FnOnce + Clone` while being less restrictive.

tower/src/util/combinators/map_err.rs
```diff
@@ -0,0 +1,70 @@
...
+{
+    type Response = S::Response;
+    type Error = Error;
+    type Future = MapErrFut<S::Future, F>;
```
LucioFranco, line 30: I've been using `MapErrFuture` instead of just `Fut`. See the map combinator.

tower/src/util/combinators/map_err.rs
```diff
@@ -0,0 +1,70 @@
...
+    /// Creates a new [`MapErrLayer`].
+    pub fn new(f: F) -> Self {
+        MapErrLayer { f, _p: PhantomData }
+    }
+}
+
+impl<S, F, Request, Error> Layer<S> for MapErrLayer<F, Request, Error>
+where
+    S: Service<Request>,
+    F: FnOnce(S::Error) -> Error + Clone,
```
LucioFranco, line 60: This is similar to what I did here https://github.com/tower-rs/tower/blob/master/tower/src/util/map/response.rs#L83

The reason I went with just `F` clone here is to remove the need to have the generic for the Request/Error type in the layer implementation. This removes the need for the phantom data. I would say we should probably follow how that implementation does it.

tower/src/util/combinators/map_ok.rs
```diff
@@ -0,0 +1,70 @@
...
+    /// Creates a new `MapOk` service.
+    pub fn new(inner: S, f: F) -> Self {
+        MapOk { f, inner }
+    }
+}
+
+impl<S, F, Request, Response> Service<Request> for MapOk<S, F>
+where
+    S: Service<Request>,
+    F: FnOnce(S::Response) -> Response + Clone,
```
LucioFranco, line 26: Same as above.

tower/src/util/combinators/map_ok.rs
```diff
@@ -0,0 +1,70 @@
...
+{
+    type Response = Response;
+    type Error = S::Error;
+    type Future = MapOkFut<S::Future, F>;
```
LucioFranco, line 30: Same as above.

tower/src/util/combinators/map_ok.rs
```diff
@@ -0,0 +1,70 @@
...
+    fn call(&mut self, request: Request) -> Self::Future {
+        self.inner.call(request).map_ok(self.f.clone())
+    }
+}
+
+/// A [`Layer`] that produces a [`MapOk`] service.
+///
+/// [`Layer`]: tower_layer::Layer
+#[derive(Debug)]
+pub struct MapOkLayer<F, Request, Response> {
```
LucioFranco, line 45: This should follow how `Map` does it.

tower/src/util/combinators/mod.rs
```diff
@@ -0,0 +1,9 @@
+mod map_err;
```
LucioFranco, line 1: Not sure if I like the module `combinators` I'd rather these be in the root no need for an extra module imo.

tower/src/util/combinators/try_with.rs
```diff
@@ -0,0 +1,74 @@
...
+    /// Creates a new [`TryWith`] service.
+    pub fn new(inner: S, f: F) -> Self {
+        TryWith { inner, f }
+    }
+}
+
+impl<S, F, NewRequest, OldRequest> Service<NewRequest> for TryWith<S, F>
+where
+    S: Service<OldRequest>,
+    F: FnOnce(NewRequest) -> Result<OldRequest, S::Error> + Clone,
```
LucioFranco, line 26: Same as above

tower/src/util/combinators/try_with.rs
```diff
@@ -0,0 +1,74 @@
...
+}
+
+impl<S, F> TryWith<S, F> {
+    /// Creates a new [`TryWith`] service.
+    pub fn new(inner: S, f: F) -> Self {
+        TryWith { inner, f }
+    }
+}
+
+impl<S, F, NewRequest, OldRequest> Service<NewRequest> for TryWith<S, F>
```
LucioFranco, line 23: We use `R1` and `R2` as the generics when we have two requests types.

tower/src/util/combinators/try_with.rs
```diff
@@ -0,0 +1,74 @@
+use futures_util::future::{ready, Either, Ready};
+use std::marker::PhantomData;
+use std::task::{Context, Poll};
+use tower_layer::Layer;
+use tower_service::Service;
+
+/// Service returned by the [`try_with`] combinator.
```
LucioFranco, line 7: can we get a bit more in the docs here? The name isn't clear enough that we probably want to add some more docs.

tower/src/util/combinators/try_with.rs
```diff
@@ -0,0 +1,74 @@
...
+    pub fn new(f: F) -> Self {
+        TryWithLayer { f, _p: PhantomData }
+    }
+}
+
+impl<S, F, OldRequest, NewRequest> Layer<S> for TryWithLayer<F, OldRequest, NewRequest>
+where
+    S: Service<OldRequest>,
+    F: FnOnce(NewRequest) -> OldRequest,
+    F: Clone,
```
LucioFranco, line 64: If we just used `Fn` we can drop that bound and just keep `Clone`. See the MapLayer implementation.

tower/src/util/combinators/with.rs
```diff
@@ -0,0 +1,69 @@
+use std::marker::PhantomData;
+use std::task::{Context, Poll};
+use tower_layer::Layer;
+use tower_service::Service;
+
+/// Service returned by the [`with`] combinator.
```
LucioFranco, line 6: Same as above.

tower/src/util/combinators/with.rs
```diff
@@ -0,0 +1,69 @@
...
+}
+
+impl<S, F> With<S, F> {
+    /// Creates a new [`With`] service.
+    pub fn new(inner: S, f: F) -> Self {
+        With { inner, f }
+    }
+}
+
+impl<S, F, NewRequest, OldRequest> Service<NewRequest> for With<S, F>
```
LucioFranco, line 22: `R1` and `R2`, same as above.

tower/src/util/combinators/try_with.rs
LucioFranco: We can also link to the docs in the extension trait.

tower/src/util/combinators/map_err.rs
hawkw: @LucioFranco I don't think that's the case. `Fn` can be called with an `&` reference. However, since the function is called _after_ the inner service's future, the function needs to be moved into the future that the `MapErr` service returns. Therefore, it cannot be borrowed without constraining the future's lifetime --- so it must be `Clone`d.  And if it _has_ to be `Clone`, it may as well be `FnOnce`, since each clone will only be called a single time.

If we wanted to use `Fn`, we would need to wrap it in an `Arc` internally, and clone the `Arc` into each future. That would also work, but in some cases (such as function pointers), the `Arc` isn't needed, since little-f `fn` can be freely cloned already.

tower/src/util/combinators/try_with.rs
hawkw: This should link to the combinator that creates it, which is the authoritative source for the docs. It's also what users will probably see first. Since the combinator docs are linked, I don't think the type really needs any additional documentation --- it would just add the burden of having to maintain two separate versions of essentially the same documentation, keep them in sync, and so on. 

This is basically what std does for iterator combinators etc.

tower/src/util/combinators/map_err.rs
LucioFranco: @hawkw is right, we should then change the bounds on the `map` combinators as well.

LucioFranco:
Thanks @hlb8122 a ton!

tower/src/util/map_err.rs
```diff
@@ -0,0 +1,67 @@
...
+    fn poll_ready(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
+        self.inner.poll_ready(cx).map_err(self.f.clone())
+    }
+
+    fn call(&mut self, request: Request) -> Self::Future {
+        self.inner.call(request).map_err(self.f.clone())
+    }
+}
+
+/// A [`Layer`] that produces a [`MapErr`] service.
```
hawkw, line 40: does the `MapErr` link need a reference?

tower/src/util/map_err.rs
```diff
@@ -0,0 +1,67 @@
...
+///
+/// [`Layer`]: tower_layer::Layer
+#[derive(Debug)]
+pub struct MapErrLayer<F> {
+    f: F,
+}
+
+impl<F> MapErrLayer<F> {
+    /// Creates a new [`MapErrLayer`].
```
hawkw, line 49: again, does this link work as-is, or does it need a reference?

IMO, a constructor doesn't _really_ need a link to the constructed type, but it is nice to have.

tower/src/util/map_ok.rs
```diff
@@ -0,0 +1,67 @@
...
+    fn poll_ready(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
+        self.inner.poll_ready(cx)
+    }
+
+    fn call(&mut self, request: Request) -> Self::Future {
+        self.inner.call(request).map_ok(self.f.clone())
+    }
+}
+
+/// A [`Layer`] that produces a [`MapOk`] service.
```
hawkw, line 40: Does the link to `MapOk` work, or does it need a reference?

tower/src/util/map_ok.rs
```diff
@@ -0,0 +1,67 @@
...
+///
+/// [`Layer`]: tower_layer::Layer
+#[derive(Debug)]
+pub struct MapOkLayer<F> {
+    f: F,
+}
+
+impl<F> MapOkLayer<F> {
+    /// Creates a new [`MapOkLayer`] layer.
```
hawkw, line 49: Does this need a reference?

tower/src/util/mod.rs
```diff
@@ -82,6 +90,244 @@ pub trait ServiceExt<Request>: tower_service::Service<Request> {
     {
         CallAll::new(self, reqs)
     }
+
+    /// Maps this service's response value to a different value. This does not
+    /// alter the behaviour of the [`poll_ready`] method.
+    ///
+    /// This method can be used to change the [`Response`] type of the service
+    /// into a different type. It is similar to the [`Result::map`]
```
hawkw, line 98: Does `Result::map` need a reference?

tower/src/util/mod.rs
```diff
@@ -82,6 +90,244 @@ pub trait ServiceExt<Request>: tower_service::Service<Request> {
...
+        F: FnOnce(Self::Response) -> Response + Clone,
+    {
+        MapOk::new(self, f)
+    }
+
+    /// Maps this services's error value to a different value. This does not
+    /// alter the behaviour of the [`poll_ready`] method.
+    ///
+    /// This method can be used to change the [`Error`] type of the service
+    /// into a different type. It is similar to the [`Result::map_err`] method.
```
hawkw, line 162: does `Result::map_err` need a reference?

tower/src/util/mod.rs
```diff
@@ -82,6 +90,244 @@ pub trait ServiceExt<Request>: tower_service::Service<Request> {
...
+        Self: Sized,
+        F: FnOnce(Self::Error) -> Error + Clone,
+    {
+        MapErr::new(self, f)
+    }
+
+    /// Composes a function *in front of* the service.
+    ///
+    /// This adapter produces a new service that passes each value through the
+    /// given function `f` before sending it to `self`.
```
hawkw, line 223: Perhaps this should reference the `futures::SinkExt::with` method as a comparison? Not a blocker.

tower/src/util/mod.rs
```diff
@@ -82,6 +90,244 @@ pub trait ServiceExt<Request>: tower_service::Service<Request> {
...
+        F: FnOnce(NewRequest) -> Request + Clone,
+    {
+        With::new(self, f)
+    }
+
+    /// Composes a fallible function *in front of* the service.
+    ///
+    /// This adapter produces a new service that passes each value through the
+    /// given function `f` before sending it to `self`.
+    ///
```
hawkw, line 278: Again, maybe worth linking `futures::SinkExt::try_with` here? Not a big deal.

tower/src/util/try_with.rs
```diff
@@ -0,0 +1,70 @@
...
+///
+/// [`try_with`]: crate::util::ServiceExt::try_with
+#[derive(Debug)]
+pub struct TryWith<S, F> {
+    inner: S,
+    f: F,
+}
+
+impl<S, F> TryWith<S, F> {
+    /// Creates a new [`TryWith`] service.
```
hawkw, line 16: Does this need a ref?

tower-util/src/lib.rs
hawkw: mmk, I'm fine with `with` if everyone else is!

tower/src/util/try_with.rs
```diff
@@ -0,0 +1,70 @@
...
+
+    fn call(&mut self, request: R1) -> Self::Future {
+        match (self.f.clone())(request) {
+            Ok(ok) => Either::Left(self.inner.call(ok)),
+            Err(err) => Either::Right(ready(Err(err))),
+        }
+    }
+}
+
+/// A [`Layer`] that produces a [`TryWith`] service.
```
hawkw, line 43: Does `TryWith`  need a ref?

tower/src/util/with.rs
```diff
@@ -0,0 +1,66 @@
...
+///
+/// [`with`]: crate::util::ServiceExt::with
+#[derive(Debug)]
+pub struct With<S, F> {
+    inner: S,
+    f: F,
+}
+
+impl<S, F> With<S, F> {
+    /// Creates a new [`With`] service.
```
hawkw, line 15: ...you know the drill by now...

tower/src/util/with.rs
```diff
@@ -0,0 +1,66 @@
...
+    fn poll_ready(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), S::Error>> {
+        self.inner.poll_ready(cx)
+    }
+
+    fn call(&mut self, request: R1) -> S::Future {
+        self.inner.call((self.f.clone())(request))
+    }
+}
+
+/// A [`Layer`] that produces a [`With`] service.
```
hawkw, line 39: ...and again.

tower/src/util/with.rs
```diff
@@ -0,0 +1,66 @@
...
+///
+/// [`Layer`]: tower_layer::Layer
+#[derive(Debug)]
+pub struct WithLayer<F> {
+    f: F,
+}
+
+impl<F> WithLayer<F> {
+    /// Creates a new [`WithLayer`].
```
hawkw, line 48: this one too :)

tower/src/util/try_with.rs
hlb8122: ["Docs.rs automatically builds crates' documentation released on crates.io using the nightly release of the Rust compiler".](https://docs.rs/about)

I've been testing it under `cargo +nightly doc --all-features --open` and everything seems to behave as intended.

hawkw: Okay, if these render correctly on docs.rs, that seems right to me --- I didn't realize that. Disregard the rest of my review! :)

hlb8122: They seem to work in my experience. The one "gotcha" that's caught me out is that when you're working in a workspace locally and referencing those crates it works fine on `cargo +nightly doc`, but when you upload them individually to crates.io you'll lose your nice links via https://github.com/rust-lang/rust/issues/65983

PR #398, opened 2019-12-19
balance: Upgrade rand to 0.7

LucioFranco:


seanmonstar:
I was super confused at first why "buffer" was even using random, but it's a typo for "balance" XD

LucioFranco:
@seanmonstar I had not had my coffee yet :D

PR #405, opened 2019-12-22
Default poll_ready implementation

hlb8122:
**Rationale**
Often one wants to quickly whip up a basic service. Many basic services will always be ready and the more complex polling dynamics will arise from layers wrapping them. 

This makes `Poll::Ready(Ok(())` a canonical and reasonable choice for a default implementation.

**Result**
Less boilerplate.

### DIFF
--- a/tower-service/src/lib.rs
+++ b/tower-service/src/lib.rs
@@ -219,6 +219,8 @@ pub trait Service<Request> {
     /// Once `poll_ready` returns `Poll::Ready(Ok(()))`, a request may be dispatched to the
     /// service using `call`. Until a request is dispatched, repeated calls to
     /// `poll_ready` must return either `Poll::Ready(Ok(()))` or `Poll::Ready(Err(_))`.
+    /// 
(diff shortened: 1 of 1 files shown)


LucioFranco:
I don't think we want to do this because we want to make sure people make the explicit choice to implement poll_ready.

LucioFranco:
might be better to open an issue to further disucss this

jonhoo:
I am against this change, as leaving a default impl there makes it too likely for people to forget to forward it when wrapping a service, which would have bad consequences. On the other hand, if someone truly wants a non-forwarding `poll_ready`, then it's super easy to write. I'd be in favor of closing this :)

jonhoo:
I am against this change, as leaving a default impl there makes it too likely for people to forget to forward it when wrapping a service, which would have bad consequences. On the other hand, if someone truly wants a non-forwarding `poll_ready`, then it's super easy to write. I'd be in favor of closing this :)

LucioFranco:
Agreed

Issue #403, opened 2019-12-19
Use github actions and cargo deny


LucioFranco:

rizsotto:
Hey Lucio, can you write down what you want to be done here? I would be glad to volunteer for this in the next year.

LucioFranco:
@rizsotto hi! basically just need to reproduce our azure build with github ci, there are no hard requirements :) Let me know if you need anything

PR #407, opened 2019-12-28
Use GitHub actions

rizsotto:
The #403 issue wanted to port the existing azure CI to GitHub actions.

The major differences are:

- Does compile it all it once (and not by crate),
- therefore it does not do the patching of the `Cargo.toml` with the local dependencies. (Is this okay? Or you guys want the per crate approach?)
- Does not update the `gh-pages` branch with the docs. (Needs to clarify: What would be the trigger for that?)
- It runs on 3 different OS and against the stable and nightly compilers.
- It has a release workflow for publish the packages to `crate.io`. (Is this desired or should be deleted for now?)
- It has a step to run `clippy`. (It breaks the build, so I commented out for now. Is this desired or should be deleted for now?)


davidbarsky:
Thanks so much for doing this and sorry for the delay! Not speaking authoritatively on behalf of the Tower project, but:

> therefore it does not do the patching of the Cargo.toml with the local dependencies. (Is this okay? Or you guys want the per crate approach?)

I _think_ this is okay.

> Does not update the gh-pages branch with the docs. (Needs to clarify: What would be the trigger for that?)

I think it'd be this trigger: https://help.github.com/en/actions/automating-your-workflow-with-github-actions/workflow-syntax-for-github-actions#onpushpull_requestbranchestags.

> It has a release workflow for publish the packages to crate.io. (Is this desired or should be deleted for now?)

This is _very_ much desired. However, for a workspace like this, `cargo publish` only applies to a single package, not a workspace. Something like https://github.com/sunng87/cargo-release might need to used instead, but that comes with its own set of workflow gotchas.

> It has a step to run clippy. (It breaks the build, so I commented out for now. Is this desired or should be deleted for now?)

Let's keep this, but note to fix clippy in a followup.

rizsotto:
Thanks David for your feedback. You've answered all my questions.

* I fixed the publish workflow (to publish the crates separately).
* Added the GitHub page publish workflow too.

.github/workflows/release.yml
```diff
@@ -0,0 +1,24 @@
+name: Create Release
```
LucioFranco: I think for now we want to not do this, since the publish process is a bit complicated right now since we have many sub dependencies and things that depend on other crates being released before others.

LucioFranco:
I think we should try and do the same as we do with azure when it comes to compiling crates so that we can ensure we catch specific errors that happen when compiling one crate instead of the entire workspace.

rizsotto:
Will try to do the per crate compiling. But there are 19 crates and 3 OSs and 2 compilers, which will make 114 jobs, plus the format job... Is there a way to reduce the number of jobs?

For curiosity, what are the errors that the per crate compiling catches?

LucioFranco:
@rizsotto we can do this, three jobs that compiles the whole workspace on each OS, then each crate gets their own job. That way its only 22 jobs doing checks and test. I don't remember exactly but when I was updating all the crates recently it caught an error that I didn't get locally. I think the idea is to disable all other crates in the workspace and only let that crate compile I think? I don't remember exactly, if you don't find a good solution then we can always stick with this and improve it moving forward.

rizsotto:
@LucioFranco I think I made the way you've described.

davidbarsky:
@rizsotto thanks, this looks great! I'm happy to merge on this and fix any issues that might crop up after the fact?

.github/workflows/build.yml
```diff
@@ -0,0 +1,135 @@
+name: continuous integration
+
+on: [push, pull_request]
```
LucioFranco: I think we want this to be like https://github.com/hyperium/tonic/blob/master/.github/workflows/CI.yml#L3

.github/workflows/build.yml
```diff
@@ -0,0 +1,135 @@
...
+        uses: actions/checkout@v1
+      - name: Install rust toolchain
+        uses: actions-rs/toolchain@v1
+        with:
+          toolchain: ${{ matrix.rust }}
+          override: true
+      - name: Build
+        uses: actions-rs/cargo@v1
+        with:
+          command: build
```
LucioFranco: this can be a `check` instead of `build` which should speed up CI time.

.github/workflows/build.yml
```diff
@@ -0,0 +1,135 @@
...
+          args: --verbose
+
+  build_per_crate:
+
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        crate:
+          - tower
```
LucioFranco: thanks for doing this 👍

.github/workflows/build.yml
```diff
@@ -0,0 +1,135 @@
...
+          # Patch all crates
+          cat ci/patch.toml >> Cargo.toml
+          # Print `Cargo.toml` for debugging
+          echo "~~~~ Cargo.toml ~~~~"
+          cat Cargo.toml
+          echo "~~~~~~~~~~~~~~~~~~~~"
+      - name: Build
+        uses: actions-rs/cargo@v1
+        with:
+          command: build
```
LucioFranco: prob can use check again here.

.github/workflows/build.yml
```diff
@@ -0,0 +1,135 @@
...
+          override: true
+          components: rustfmt
+      - name: Run rustfmt
+        uses: actions-rs/cargo@v1
+        with:
+          command: fmt
+          args: --all -- --check
+
+# This is failing, because it finds errors...
+#  clippy_check:
```
LucioFranco: Yeah, we should prob run clippy fix and then start using this 👍

.github/workflows/build.yml
rizsotto: That's pretty much the same as the current, but you lock out other branches than master.

I think the whole point for a CI server to protect the master branch. Which you can only do, if you check those branches which will be merged to master. So, I would recommend to keep it like this. What do you think?

.github/workflows/build.yml
rizsotto: fixed

.github/workflows/build.yml
rizsotto: fixed

rizsotto:
I've tested the `build.yml`, but the `publish.yml` is not yet tested. (Did try, but don't know how to verify it.)

PR #414, opened 2020-02-11
Add basic "batteries-included" `retry::Policy`s.

hdevalence:
These are lifted out of the test and example code.  Two policies are provided:

* `RetryLimit`, with a bounded number of retry attempts;
* `RetryError`, with unbounded retry attempts.

Both policies require `Req: Clone` as it's not possible to write a generic
retry policy without being able to clone requests.

The docs should be updated with a little blurb that explains when they should
not be used.  Also, the `Policy` docs had an `Attempts` example corresponding
to `RetryLimit` (formerly `Limit` in the tests code); I removed it because
`RetryLimit` exists and can be view-sourced.

### DIFF
--- a/tower-retry/src/policies.rs
+++ b/tower-retry/src/policies.rs
@@ -1,0 +1,61 @@
+use super::Policy;
+use futures_util::future;
+
+/// A very basic retry policy with a limited number of retry attempts.
+///
+/// FIXME: explain why not to use this.
+#[derive(Clone, Debug)]
+pub struct RetryLimit {
+    remaining_tries: usize,
+}
+
+impl RetryLimit {
+    /// Create a policy with the given number of retry attempts.
+    pub fn new(retry_attempts: usize) -> Self {
+        RetryLimit {
+            remaining_tries: retry_attempts,
+        }
+    }
+}
+
+impl<Req: Clone, Res, E> Policy<Req, Res, E> for RetryLimit {
+    type Future = future::Ready<Self>;
+    fn retry(&self, _: &Req, result: Result<&Res, &E>) -> Option<Self::Future> {
+        if result.is_err() {
+            if self.remaining_tries > 0 {
+                Some(future::ready(RetryLimit {
+                    remaining_tries: self.remaining_tries - 1,
+                }))
+            } else {
+                None
+            }
+        } else {
+            None
+        }
+    }
+
+    fn clone_request(&self, req: &Req) -> Option<Req> {
+        Some(req.clone())
+    }
+}
+
+/// A very basic retry policy that always retries failed requests.
+///
+/// FIXME: explain why not to use this.
+#[derive(Clone, Debug)]
+pub struct RetryErrors;
+
+impl<Req: Clone, Res, E> Policy<Req, Res, E> for RetryErrors {
+    type Future = future::Ready<Self>;
+    fn retry(&self, _: &Req, result: Result<&Res, &E>) -> Option<Self::Future> {
+        if result.is_err() {
+            Some(future::ready(RetryErrors))
+        } else {
+            None
+        }
+    }
+
+    fn clone_request(&self, req: &Req) -> Option<Req> {
+        Some(req.clone())
+    }
+}
--- a/tower-retry/Cargo.toml
+++ b/tower-retry/Cargo.toml
@@ -27,9 +27,9 @@ tower-layer = "0.3"
 tokio = { version = "0.2", features = ["time"] }
 pin-project = "0.4"
 futures-core = { version = "0.3", default-features = false }
+futures-util = { version = "0.3", default-features = false }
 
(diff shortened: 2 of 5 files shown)


tower-retry/src/policies.rs
```diff
@@ -0,0 +1,61 @@
...
+        RetryLimit {
+            remaining_tries: retry_attempts,
+        }
+    }
+}
+
+impl<Req: Clone, Res, E> Policy<Req, Res, E> for RetryLimit {
+    type Future = future::Ready<Self>;
+    fn retry(&self, _: &Req, result: Result<&Res, &E>) -> Option<Self::Future> {
+        if result.is_err() {
```
LucioFranco: I wonder if there is a way to traitize this? For example, http retries you may want to retry 500s but not 400s. In this case, if I were to apply this to hyper I would only be able to retry system errors not http errors. What do you think about providing some method to allow users to specify what they want to retry?

hdevalence: One possibility would be to construct the `RetryLimit` with a closure `F: FnMut(&E) -> bool`; then callers who wanted to retry all errors could construct it as, e.g.,
```
let policy = RetryLimit::new(2, |_| true);
```
and callers who want to filter retries can stick whatever logic they want in the closure.  (Does that seem like the right bound for the closure?)

LucioFranco: Though since this only returns an error that wouldn't support retrying on http::Responses?

I think we could use some trait _and_ a TraitFn adapter for a closure, what do you think?

hdevalence: Hmm, I'm not sure what kind of trait you're thinking of. I think I'm missing something: if the user-supplied logic has access to the whole response and error, how would that be different from just implementing `Policy::retry` directly?

LucioFranco: This is closer to the related vector retrylogic thing I sent on discord, but the idea is that as a user I can just provide what i'd like to retry on instead of having to implement a retry policy myself.

LucioFranco: I wouldn't copy this 1-1 but https://github.com/timberio/vector/blob/master/src/sinks/util/retries.rs#L25

seanmonstar: > I wonder if there is a way to traitize this? For example, http retries you may want to retry 500s but not 400s.

For example, finagle uses a `ResponseClassifier` in a similar way.

hawkw:
@hdevalence any interest in wrapping this up, or should we just close the PR?

hdevalence:
I'd be happy to wrap this up, but I don't really know what the "good" design would be -- the current PR is just what worked for my needs.  But, if there was clarity about what changes would be good, I'd be happy to make them, and I think it would be convenient if something like this was included in Tower.

hawkw:
> I'd be happy to wrap this up, but I don't really know what the "good" design would be -- the current PR is just what worked for my needs. But, if there was clarity about what changes would be good, I'd be happy to make them, and I think it would be convenient if something like this was included in Tower.

I wonder if @luciofranco has any ideas about this.

LucioFranco:
@rcoh maybe you can chime in here a bit?

rcoh:
I don't necessarily want to impose our retry needs on others, but as a data point here are things we need:
- Classifier interface that has access to the full `Result<T, E>` (and not just `E`)
- The ability for the retry service to modify the responses, eg. set metadata about the number of retries required, set a custom error if you ran out of retries, etc. I sketched a possible design for this in #546 

Our retry policy behavior (around number of retries, backoff length, etc.) is far too complex to be well served by something shared, unfortunately, so I don't think we have much need for batteries included policies.

hdevalence:
Those things are all possible using the existing retry trait, right? And, if they're not, that's an issue with the trait itself, not with any default policies, correct?

My goal with this PR was just to include some basic implementations that are sufficient for simple retry logic — of course, if there are more complex use cases, nothing prevents a user from implementing the trait themselves, but I'm not sure that that should mean that it's a bad idea to include some simpler default behavior.

davidpdrsn (commented):
My 2 cents: I think this is worth merging (once the `FIXME`s are fixed) as simple retry policy people can use if they want.

I also think some kind of response classifier is useful and we've actually been designing something over in tower-http https://github.com/tower-rs/tower-http/blob/classify-response/tower-http/src/classify.rs. The traits themselves aren't protocol specific so they could in theory live in tower. I think thats a separate discussion however.

davidpdrsn:
@hdevalence sorry this has taken so long to land.

My opinion from 2 months still lands. If we fix the small docs things I would be okay with merging this. I think the discussion about whether or not to change Retry in a breaking way is a separate discussion.

Do you wanna drive this home or should I take over?

hdevalence:
@davidpdrsn I think I've lost context on this one, so feel free to push it over the finish line.

Issue #415, opened 2020-02-13
Panicked at 'missing cancelation' in tower-ready-cache


jonhoo:
I just had my code hit [this `.expect`](https://github.com/tower-rs/tower/blob/47c3a145605b7258ea280c4c64ee7488fdabf4c4/tower-ready-cache/src/cache.rs#L239) in `tower-ready-cache`. It's using `tower-balance::pool` (and thus also `tower-balance::p2c`). Not sure what caused it, but figured maybe @olix0r can tease it out from the backtrace?
```
thread 'tokio-runtime-worker' panicked at 'missing cancelation', /home/ubuntu/.cargo/registry/src/github.com-1ecc6299db9ec823/tower-ready-cache-0.3.0/src/cache.rs:236:37
stack backtrace:
   0: backtrace::backtrace::libunwind::trace
             at /cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.40/src/backtrace/libunwind.rs:88
   1: backtrace::backtrace::trace_unsynchronized
             at /cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.40/src/backtrace/mod.rs:66
   2: std::sys_common::backtrace::_print_fmt
             at src/libstd/sys_common/backtrace.rs:77
   3: <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt
             at src/libstd/sys_common/backtrace.rs:59
   4: core::fmt::write
             at src/libcore/fmt/mod.rs:1052
   5: std::io::Write::write_fmt
             at src/libstd/io/mod.rs:1428
   6: std::sys_common::backtrace::_print
             at src/libstd/sys_common/backtrace.rs:62
   7: std::sys_common::backtrace::print
             at src/libstd/sys_common/backtrace.rs:49
   8: std::panicking::default_hook::{{closure}}
             at src/libstd/panicking.rs:204
   9: std::panicking::default_hook
             at src/libstd/panicking.rs:224
  10: std::panicking::rust_panic_with_hook
             at src/libstd/panicking.rs:470
  11: rust_begin_unwind
             at src/libstd/panicking.rs:378
  12: core::panicking::panic_fmt
             at src/libcore/panicking.rs:85
  13: core::option::expect_failed
             at src/libcore/option.rs:1203
  14: core::option::Option<T>::expect
             at /rustc/fc23a81831d5b41510d3261c20c34dd8d32f0f31/src/libcore/option.rs:347
[...]

PR #416, opened 2020-02-20
balance: Add a stress test for p2c

jonhoo:
The hope for this was to reproduce #415 (which it does not sadly), but
at least it adds a test for p2c!

### DIFF
--- a/tower-balance/tests/p2c.rs
+++ b/tower-balance/tests/p2c.rs
@@ -161,7 +161,5 @@ fn stress() {
                 assert_eq!(nready, 0, "got pending when a service is ready");
             }
         }
-
-        // assert_eq!(cache.len(), services.len());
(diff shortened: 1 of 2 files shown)


tower-balance/tests/p2c.rs
```diff
@@ -0,0 +1,167 @@
...
+            Poll::Ready(Ok(())) => {
+                assert_ne!(nready, 0, "got ready when no service is ready");
+            }
+            Poll::Ready(_) => unreachable!("discover stream has not failed"),
+            Poll::Pending => {
+                assert_eq!(nready, 0, "got pending when a service is ready");
+            }
+        }
+
+        // assert_eq!(cache.len(), services.len());
```
LucioFranco: ?

PR #425, opened 2020-03-03
Add Buffer::new note on how to set bound

akshayknarayan:
Related to https://github.com/tower-rs/tower/issues/408. 
While the above issue is resolved, tower-buffer should document its behavior and offer guidance on setting the channel size.

cc @jonhoo

### DIFF
--- a/tower-buffer/src/service.rs
+++ b/tower-buffer/src/service.rs
@@ -36,7 +36,7 @@ where
     /// must be called while on the Tokio runtime.
     ///
     /// # Note on setting `bound`
-    /// `Buffer`'s implementation of `poll_ready` will, if it will return `Poll::Ready`, reserve a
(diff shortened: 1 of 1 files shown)


akshayknarayan:
CI failure seems to be unrelated to diff

tower-buffer/src/service.rs
```diff
@@ -34,6 +34,12 @@ where
     ///
     /// The default Tokio executor is used to run the given service, which means that this method
     /// must be called while on the Tokio runtime.
+    ///
+    /// # Note on setting `bound`
+    /// `Buffer`'s implementation of `poll_ready` will, if it will return `Poll::Ready`, reserve a
```
jonhoo: Let's phrase this in the present tense:
> When `Buffer`'s implementation of `poll_ready` returns `Poll::Ready`, it reserves a

PR #426, opened 2020-03-10
New sub-crate: tower-steer

akshayknarayan:
In some scenarios, you have a list of services and want to send requests to one of them, and get the corresponding response. In the example that motivated me to write this, I had a sharded key-value store, and needed to route requests to the service corresponding to the right shard.

To make this easier, tower-router provides functionality which keeps track of which services have pending requests, waits for those services to be ready, issues calls against them, and returns the response.

This can be thought of as an unbounded (to avoid head-of-line blocking) tower-buffer with extra details to allow routing requests to specific services.

@jonhoo


jonhoo:
I'll avoid line-by-line reviewing for the time being, and focus on the high-level points:

First, I think this is a cool service, and we should almost definitely have something like this in tower (cc @LucioFranco).

Second, as we discussed, I think `ServiceRouter::poll_ready` should poll _all_ the services. If the user _wants_ unlimited buffering, they should wrap the `Service` in an unbounded tower-buffer (which doesn't exist yet :sweat_smile:), but we should not force this on the user given that it is not a _requirement_ of router that it buffers. This would also, I believe, remove the need for the whole oneshot-arc-mutex-spawn song and dance the code currently has to do.

Also some slightly smaller points:

 - `Router` should probably also be given a reference to the slice of services it is choosing among.
 - Should there be a way to add or remove services from a `Router`?
 - I wonder if we could use something like `tower-ready-cache` to let us more efficiently `poll_ready` all the services.

And for the interested reader: the reason we **need to poll all services** is that we do not know at the time of the `poll_ready` which service the request subsequently sent to `call` will be for. We have two options. Either, we unconditionally accept the next element, and only then `poll_ready` on the relevant service, but then we are forced to buffer potentially an unlimited number of elements in case they are all for an unready service. Or, we get head-of-line blocking because _every_ request requires _all_ services to be ready. I suppose _in theory_ there is a third option of trying to "guess" which service the next request will be for, but that still deteriorates to the "buffer indefinitely" in the worst case.

tower-router/src/lib.rs
```diff
@@ -0,0 +1,398 @@
...
+/// the correct `Service`. It does *not* `.await` these `Future`s - this is left to the caller.
+/// This way, the caller is able to `.await` several returned `Future`s at once, instead of
+/// blocking on the completion of one at a time.
+///
+/// It will `poll_ready` only the services which have pending requests. Otherwise, we could
+/// end up repeatedly calling `poll_ready` on a service for which there are no requests.
+#[pin_project::pin_project]
+#[derive(Debug)]
+pub struct ServiceGroup<S, Req> {
+    cls: Vec<S>,
```
LucioFranco: This means each service must be the same type, correct? That is a pretty big limitation.

tower-router/src/lib.rs
```diff
@@ -0,0 +1,398 @@
...
+/// 2. Waits for that `Service` to be ready (using [`ServiceGroup`]).
+/// 3. Calls the `Service` with the request, and waits for the response.
+/// 4. Resolves to the response of the call.
+///
+/// Unlike `tower-buffer`, `ServiceRouter` does not require `S` to `impl Clone`, but like
+/// `tower-buffer` it uses a [`oneshot`] to manage requests.
+#[derive(Debug)]
+pub struct ServiceRouter<S, Req, T: 'static, F> {
+    router: F,
+    inner: Arc<Mutex<ServiceGroup<OneshotService<S>, (oneshot::Sender<T>, Req)>>>,
```
LucioFranco: Why the arc mutex?

tower-router/src/lib.rs
jonhoo: For posterity: Lucio and I chatted about this elsewhere and decided that it probably makes sense to have two different "routers": one, like this one, which chooses among services that have the same request type, and one that chooses which service to forward to depending on the request type. This one should probably be called something like `tower-choose` or `tower-shard`, while the other is (perhaps) a more natural fit for the name `tower-router`.

tower-router/src/lib.rs
jonhoo: It'll go away. I chatted to @akshayknarayan offline.

akshayknarayan:
I've renamed the implementation to `tower-steer` and deleted most of the code (which was there mostly for the oneshot-mutex stuff).

tower-steer/src/lib.rs
```diff
@@ -0,0 +1,153 @@
...
+/// An example use case is a sharded service.
+/// It accepts new requests, then:
+/// 1. Determines, via the provided [`Picker`], which `Service` the request coresponds to.
+/// 2. Waits (in `poll_ready`) for *all* services to be ready.
+/// 3. Calls the correct `Service` with the request, and returns a future corresponding to the
+///    call.
+#[derive(Debug)]
+pub struct Steer<S, F, Req> {
+    router: F,
+    // tuple of is_ready, service
```
jonhoo: ?

tower-steer/src/lib.rs
```diff
@@ -0,0 +1,153 @@
...
+    cls: Vec<S>,
```
jonhoo:
```suggestion
    services: Vec<S>,
```

tower-steer/src/lib.rs
```diff
@@ -0,0 +1,153 @@
...
+    ready: Vec<bool>,
```
jonhoo: This should probably be a `Vec<usize>`, otherwise `poll_ready` always has to iterate over all `N` services even if only 1 is still pending.

tower-steer/src/lib.rs
```diff
@@ -0,0 +1,153 @@
...
+    _phantom: std::marker::PhantomData<Req>,
+}
+
+impl<S, F, Req> Steer<S, F, Req>
+where
+    S: Service<Req, Error = StdError>,
+    S::Future: 'static,
```
jonhoo: Why do we need this bound?

tower-steer/src/lib.rs
```diff
@@ -0,0 +1,153 @@
...
+    router: F,
+    // tuple of is_ready, service
+    cls: Vec<S>,
+    ready: Vec<bool>,
+    _phantom: std::marker::PhantomData<Req>,
+}
+
+impl<S, F, Req> Steer<S, F, Req>
+where
+    S: Service<Req, Error = StdError>,
```
jonhoo: I _think_ we only need this bound on the `impl Service` block below?

tower-steer/src/lib.rs
```diff
@@ -0,0 +1,153 @@
...
+        Self {
+            router,
+            cls,
+            ready,
+            _phantom: Default::default(),
+        }
+    }
+}
+
+impl<S, Req, T, F> Service<Req> for Steer<S, F, Req>
```
jonhoo: I don't think you need to be generic over `T` here. Instead, just remove `Response = T` in the `Service` bound, and then use `S::Response` in place of `T` everywhere else.

tower-steer/src/lib.rs
```diff
@@ -0,0 +1,153 @@
...
+where
+    S: Service<Req, Response = T, Error = StdError>,
+    S::Future: 'static,
+    F: Picker<S, Req>,
+{
+    type Response = T;
+    type Error = StdError;
+    type Future = Pin<Box<dyn Future<Output = Result<T, StdError>>>>;
```
jonhoo: I think this can now just be `S::Future`?

tower-steer/src/lib.rs
```diff
@@ -0,0 +1,153 @@
...
+}
+
+impl<S, Req, T, F> Service<Req> for Steer<S, F, Req>
+where
+    S: Service<Req, Response = T, Error = StdError>,
+    S::Future: 'static,
+    F: Picker<S, Req>,
+{
+    type Response = T;
+    type Error = StdError;
```
jonhoo: I think this can now just be `S::Error`?

tower-steer/src/lib.rs
```diff
@@ -0,0 +1,153 @@
...
+    type Future = Pin<Box<dyn Future<Output = Result<T, StdError>>>>;
+
+    fn poll_ready(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
+        use futures_util::ready;
+        // must wait for *all* services to be ready.
+        // this will cause head-of-line blocking unless the underlying services are always ready.
```
jonhoo: May be useful to include this in the docs somewhere?

tower-steer/src/lib.rs
akshayknarayan: This will make it higher cost to have the `assert!(ready)` in `call` (would need to scan the `Vec<usize>` for `idx`), but maybe that is ok - it can be removed

tower-steer/src/lib.rs
akshayknarayan: it is: https://github.com/tower-rs/tower/pull/426/files/1cde2511a0fdee7e56a6f9ee4327edb74fcb1bd2#diff-91249aa82a3e4b3ca98888fc272df30dR86

maybe it can be made more prominent

tower-steer/src/lib.rs
jonhoo: I'm not sure I follow what you say here? This seems like it would be strictly _cheaper_, since you never need to walk a list?

tower-steer/src/lib.rs
```diff
@@ -0,0 +1,141 @@
...
+        use futures_util::ready;
+
+        // must wait for *all* services to be ready.
+        // this will cause head-of-line blocking unless the underlying services are always ready.
+        if self.not_ready.is_empty() {
+            Poll::Ready(Ok(()))
+        } else {
+            ready!(self.services[self.not_ready[0]].poll_ready(cx))?;
+            self.not_ready.pop_front();
+            self.poll_ready(cx)
```
jonhoo: This should probably be a `loop` rather than recursive, so we avoid overflowing stacks when there are many unready services. If we had guaranteed tail recursion of course, that'd be a different thing, but we do not (yet) sadly.

tower-steer/src/lib.rs
jonhoo: Ah, sorry, no, I was referring specifically to the advice about using an unbounded channel in front of each service / making sure it always returns `poll_ready`.

tower-steer/src/lib.rs
```diff
@@ -0,0 +1,141 @@
...
+        }
+    }
+
+    fn call(&mut self, req: Req) -> Self::Future {
+        let idx = self.router.pick(&req, &self.services[..]);
+        let cl = &mut self.services[idx];
+        let fut = cl.call(req);
+        // mark not ready
+        self.not_ready.push_back(idx);
```
jonhoo: I would probably just place this above the `.call` above, and then avoid the `fut` intermediate.

tower-steer/src/lib.rs
akshayknarayan: I'm specifically referring to the `assert!(ready)` - if you want to verify that a specific service is actually ready (which is not actually necessary because the implementation of poll_ready checks that all services are ready), you need to walk the ready list.
Since this check is unnecessary anyway, the latest commit removes it.

tower-steer/src/lib.rs
akshayknarayan: yeah, this is added to docs in the latest commit

tower-steer/src/lib.rs
akshayknarayan: :(

tower-steer/src/lib.rs
jonhoo: Ahh, I see what you mean. Yeah, that should be fine. You could just as well at this point `assert!(self.not_ready.is_empty())`.

tower-steer/Cargo.toml
```diff
@@ -0,0 +1,16 @@
...
+authors = ["Akshay Narayan <akshayn@mit.edu>"]
+edition = "2018"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+tower-service = "0.3"
+futures-util = "0.3"
+pin-project = "0.4"
+tokio = { version = "0.2", features = ["sync", "rt-core"]}
```
jonhoo: I don't think we depend on `tokio` any more now?

tower-steer/Cargo.toml
```diff
@@ -0,0 +1,16 @@
...
+version = "0.1.0"
+authors = ["Akshay Narayan <akshayn@mit.edu>"]
+edition = "2018"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+tower-service = "0.3"
+futures-util = "0.3"
+pin-project = "0.4"
```
jonhoo: Probably now unnecessary?

tower-steer/Cargo.toml
```diff
@@ -0,0 +1,16 @@
...
+name = "tower-steer"
+version = "0.1.0"
+authors = ["Akshay Narayan <akshayn@mit.edu>"]
+edition = "2018"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+tower-service = "0.3"
+futures-util = "0.3"
```
jonhoo: Probably now unnecessary?

tower-steer/Cargo.toml
```diff
@@ -0,0 +1,14 @@
+[package]
```
jonhoo: We probably want to tidy this up a bit and make it more like the `Cargo.toml` of the various other `tower-` subcrates. A simple `README.md` would also be good to add.

hawkw:
I suspect @olix0r will have some insights into this, though he's currently out of the office...

tower-steer/src/lib.rs
```diff
@@ -0,0 +1,142 @@
...
+//!     let reqs = vec!["A", "b", "C", "d"];
+//!     let reqs: Vec<String> = reqs.into_iter().map(String::from).collect();
+//!     for r in reqs {
+//!         poll_fn(|cx| s.poll_ready(cx)).await.unwrap();
+//!         s.call(r).await;
+//!     }
+//! }
+//! ```
+#![deny(missing_docs)]
+#![warn(unreachable_pub)]
```
LucioFranco: minor nit: can we combine these?

tower-steer/src/lib.rs
```diff
@@ -0,0 +1,142 @@
...
+/// It accepts new requests, then:
+/// 1. Determines, via the provided [`Picker`], which `Service` the request coresponds to.
+/// 2. Waits (in `poll_ready`) for *all* services to be ready.
+/// 3. Calls the correct `Service` with the request, and returns a future corresponding to the
+///    call.
+///
+/// Note that `Steer` must wait for all services to be ready since it can't know ahead of time
+/// which `Service` the next message will arrive for, and is unwilling to buffer items
+/// indefinitely. This will cause head-of-line blocking unless paired with a `Service` that does
+/// buffer items indefinitely, and thus always returns `Poll::Ready`.
```
LucioFranco: would be good to mention `tower-buffer` explicitly here.

tower-steer/src/lib.rs
```diff
@@ -0,0 +1,142 @@
...
+                    return Poll::Pending;
+                }
+
+                self.not_ready.pop_front();
+            }
+        }
+    }
+
+    fn call(&mut self, req: Req) -> Self::Future {
+        assert!(self.not_ready.is_empty());
```
LucioFranco: assert message would be good here

tower-steer/src/lib.rs
akshayknarayan: `tower-buffer` doesn't (currently) support buffering indefinitely, no? As a result it might be confusing to mention it.

LucioFranco: Ah you're right! Though it does solve part of this problem since you can set a high enough size that poll_ready will never apply back pressure.

akshayknarayan: sure, I can add a note mentioning that example.

jonhoo:
Haha, okay, so, given that #432 just landed, @akshayknarayan can you re-organize to match the new layout? Sorry for the trouble! Should be a matter of:

 - Add `steer` feature to `Cargo.toml`
 - `git mv tower-steer/src tower/src/steer`
 - `git mv tower/src/steer/lib.rs tower/src/steer/mod.rs`
 - `git rm tower-steer/Cargo.toml tower-steer/README.md`
 - Add
   ```rust
   #[cfg(feature = "steer")]
   pub mod steer;
   ```
   to `tower/src/lib.rs`

jonhoo:
The docs should also be updated to not talk about "crate" any more. And the various `#!` items should probably be removed/lifted to `tower/src/lib.rs`.

Also, I just noticed. There aren't any tests :thinking:

jonhoo:
Beautiful, thank you!

PR #427, opened 2020-03-20
util: Add ReadyAnd to do what Ready should do

jonhoo:
`ServiceExt::ready` says that it produces "A future yielding the service
when it is ready to accept a request." This is not true; it does _not_
yield the service when it is ready, it yields unit. This makes it
impossible to chain service ready with service call, which is sad.

This PR adds `ready_and`, which does what `ready` promised. It also
deprecates `ready` with the intention that we remove `ready` in a future
version, and make the strictly more general `ready_and` take its place.
We can't do it now since it's not a backwards-compatible change even
though it _probably_ wouldn't break any code.

The PR also updates the docs so that they reflect the observed behavior.

### DIFF
--- a/tower-util/src/ready.rs
+++ b/tower-util/src/ready.rs
@@ -8,23 +8,23 @@ use std::{
 };
 use tower_service::Service;
 
-/// A future that yields a `Service` when it is ready to accept a request.
+/// A future that yields the service when it is ready to accept a request.
 ///
-/// `ReadyAnd` values are produced by `ServiceExt::ready_and`.
-pub struct ReadyAnd<'a, T, Request> {
-    inner: Option<&'a mut T>,
+/// `ReadyOneshot` values are produced by `ServiceExt::ready_oneshot`.
+pub struct ReadyOneshot<T, Request> {
+    inner: Option<T>,
     _p: PhantomData<fn() -> Request>,
 }
 
 // Safety: This is safe because `Services`'s are always `Unpin`.
-impl<'a, T, Request> Unpin for ReadyAnd<'a, T, Request> {}
+impl<T, Request> Unpin for ReadyOneshot<T, Request> {}
 
-impl<'a, T, Request> ReadyAnd<'a, T, Request>
+impl<T, Request> ReadyOneshot<T, Request>
 where
     T: Service<Request>,
 {
     #[allow(missing_docs)]
-    pub fn new(service: &'a mut T) -> Self {
+    pub fn new(service: T) -> Self {
         Self {
             inner: Some(service),
             _p: PhantomData,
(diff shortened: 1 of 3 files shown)


tower-util/src/ready.rs
```diff
@@ -8,27 +8,114 @@ use std::{
...
 }
 
 // Safety: This is safe because `Services`'s are always `Unpin`.
+impl<T, Request> Unpin for ReadyOneshot<T, Request> {}
+
+impl<T, Request> ReadyOneshot<T, Request>
+where
+    T: Service<Request>,
+{
+    #[allow(missing_docs)]
```
LucioFranco: should we just add docs now :)

tower-util/src/ready.rs
```diff
@@ -8,27 +8,114 @@ use std::{
...
 use tower_service::Service;
 
-/// Future yielding a `Service` once the service is ready to process a request
+/// A future that yields the service when it is ready to accept a request.
 ///
-/// `Ready` values are produced by `ServiceExt::ready`.
-pub struct Ready<'a, T, Request> {
-    inner: &'a mut T,
+/// `ReadyOneshot` values are produced by `ServiceExt::ready_oneshot`.
+pub struct ReadyOneshot<T, Request> {
```
LucioFranco: Do we want a manual clone impl if T is clone?

tower-util/src/ready.rs
```diff
@@ -8,27 +8,114 @@ use std::{
...
+        f.debug_struct("ReadyOneshot")
+            .field("inner", &self.inner)
+            .finish()
+    }
+}
+
+/// A future that yields a mutable reference to the service when it is ready to accept a request.
+///
+/// `ReadyAnd` values are produced by `ServiceExt::ready_and`.
+pub struct ReadyAnd<'a, T, Request>(ReadyOneshot<&'a mut T, Request>);
```
LucioFranco: Would `ReadyMut` make sense here?

jonhoo: Really I want this to be called just `Ready`, but for the time being that's not really possible. In the interrim though, I find `ReadyMut` a weird name (I know tokio has used it in the past) because it's not really important that it holds something that is `mut`. It's more important that it holds something that is `&` (as opposed to owned), but `ReadyRef` would also be kind of weird (since it _doesn't_ contain "mut"). I went with `ReadyAnd` because it has a nice cadence to it: "await this service being ready and ...". It also then matches the method name on `ServiceExt`, which I think reads better as `ready_and` than as `ready_mut`.

tower-util/src/ready.rs
jonhoo: My first instinct says no — it should be rare that you'd want to clone a `ReadyOneshot` as opposed to cloning the underlying `Service`. If you did, that's _probably_ not what you meant to do.

tower-util/src/ready.rs
jonhoo: I always found documentation on `new` super weird since it becomes so trivial (for simple cases like this). I guess I could add something, but my inclination is to leave it :sweat_smile:

tower-util/src/ready.rs
LucioFranco: @jonhoo why don't we do a breaking change and bump version number? We could also consider combining crates to make this process easier.

jonhoo: I figured we might want to do that eventually, but didn't want to block this on a major version bump. I'm guessing there are some other changes we'd want to make for that too, so that would likely take longer?

LucioFranco: Yeah, lets just ship this as is then and we can do the nice stuff later.

PR #434, opened 2020-04-03
Test out Service::disarm for all tower services

jonhoo:
See #408 and https://github.com/rust-lang/futures-rs/issues/2109.

### DIFF
--- a/tower-service/src/lib.rs
+++ b/tower-service/src/lib.rs
@@ -79,6 +79,8 @@ use std::task::{Context, Poll};
 ///         // Return the response as an immediate future
 ///         Box::pin(fut)
 ///     }
+///
+///     fn disarm(&mut self) {}
 /// }
 /// ```
 ///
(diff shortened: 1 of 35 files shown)


tower-service/src/lib.rs
```diff
@@ -234,6 +236,79 @@ pub trait Service<Request> {
...
+    /// ```
+    ///
+    /// If many such forwarders exist, and they all forward into a single (cloned) `BufferService`,
+    /// then any number of the forwarders may be waiting for `rx.next` at the same time. While they
+    /// do, they are effectively each reducing the buffer's capacity by 1. If enough of these
+    /// forwarders are idle, forwarders whose `rx` _do_ have elements will be unable to find a spot
+    /// for them through `poll_ready`, and the system will deadlock.
+    ///
+    /// `disarm` solves this problem by allowing you to give up the reserved slot if you find that
+    /// you have to block. We can then fix the code above by writing:
```
LucioFranco, line 282: I don't really like this term block here, I think its overloading the word, we may want to find a better word here.

tower/src/ready_cache/cache.rs
```diff
@@ -367,6 +367,35 @@ where
...
+    /// Returns the number of services disarmed this way.
+    ///
+    /// Note that `ReadyCache` eagerly polls services (one call to `poll_ready` may ready several
+    /// services), so a call to disarm may disarm a large number of underlying services.
+    pub fn disarm(&mut self) -> usize {
+        let were_ready = self.ready.len();
+
+        // Take self.ready so we can call &mut self methods below
+        let mut ready = std::mem::take(&mut self.ready);
+        for (key, (mut svc, cancel)) in ready.drain(..) {
```
LucioFranco, line 382: ouch :(

tower/src/reconnect/mod.rs
```diff
@@ -146,6 +146,15 @@ where
         let fut = service.call(request);
         ResponseFuture::new(fut)
     }
+
+    fn disarm(&mut self) {
+        // we must be in State::Connected if poll_ready succeeded
+        if let State::Connected(ref mut service) = self.state {
+            service.disarm()
```
LucioFranco, line 153: technically should be able to disarm the make service?

tower-service/src/lib.rs
```diff
@@ -234,6 +236,79 @@ pub trait Service<Request> {
     /// Implementations are permitted to panic if `call` is invoked without
     /// obtaining `Poll::Ready(Ok(()))` from `poll_ready`.
     fn call(&mut self, req: Request) -> Self::Future;
+
+    /// Undo a successful call to `poll_ready`.
```
LucioFranco, line 240: Maybe I missed this but would be good to mention that most uses cases need to just forward the inner disarm.

tower/src/ready_cache/cache.rs
jonhoo: Yeah... It's not _expensive_ (like, it won't allocate for example), but it does mean that you might un-ready a bunch of services.

tower/src/reconnect/mod.rs
jonhoo: I don't think so? `disarm` should only ever be called after `poll_ready` completes successfully, and if I'm reading the code right, that always implies that the state is `State::Connected`. Which in turn implies that we called `poll_ready` on the make service _and_ called `make_service` (which consumes that readiness).

LucioFranco: that is fair!

PR #436, opened 2020-04-14
Refactor github actions

LucioFranco:


.github/workflows/CI.yml
```diff
@@ -7,161 +7,125 @@ on:
...
+    - uses: actions/checkout@master
+    - uses: actions-rs/toolchain@v1
+      with:
+        toolchain: ${{ matrix.rust }}
+        profile: minimal
+    - name: Check
+      uses: actions-rs/cargo@v1
+      with:
+        command: check
+        args: --all --bins --examples --tests --benches --all-features
```
jonhoo, line 26: Why does this explicitly enumerate instead of using, say, `--all-targets`?

.github/workflows/CI.yml
```diff
@@ -7,161 +7,125 @@ on:
...
-
     strategy:
       matrix:
-        crate:
-          - tower
-          - tower-layer
-          - tower-service
-          - tower-test
-
+        rust: [stable, beta, nightly, 1.39.0]
```
jonhoo, line 80: Is there a way for us to allow failures on nightly?

.github/workflows/CI.yml
```diff
@@ -7,161 +7,125 @@ on:
...
+    - uses: actions-rs/toolchain@v1
+      with:
+        toolchain: stable
+        components: rustfmt
+        profile: minimal
+    - name: rustfmt
+      uses: actions-rs/cargo@v1
+      with:
+        command: fmt
+        args: --all -- --check
```
jonhoo, line 108: Should we also run a clippy check?

.github/workflows/CI.yml
```diff
@@ -7,161 +7,125 @@ on:
...
-#          toolchain: nightly
-#          override: true
-#          components: clippy
-#      - name: Run clippy
-#        uses: actions-rs/clippy-check@v1
-#        with:
-#          token: ${{ secrets.GITHUB_TOKEN }}
-#          args: --all-targets --all-features
+    - uses: actions/checkout@v1
+    - uses: EmbarkStudios/cargo-deny-action@v0
```
jonhoo, line 131: newline

tower-service/Cargo.toml
```diff
@@ -24,4 +24,4 @@ edition = "2018"
 [dependencies]
 
 [dev-dependencies]
-http = "0.1"
+http = "0.2"
```
jonhoo, line 27: This seems unrelated?

tower/Cargo.toml
```diff
@@ -60,7 +60,7 @@ tokio = { version = "0.2", optional = true }
 [dev-dependencies]
 futures-util = { version = "0.3", default-features = false, features = ["alloc", "async-await"] }
 hdrhistogram = "6.0"
-quickcheck = { version = "0.6", default-features = false }
+quickcheck = { version = "0.9", default-features = false }
```
jonhoo, line 63: This seems unrelated?

deny.toml
```diff
@@ -0,0 +1,13 @@
+[licenses]
+unlicensed = "deny"
+# We want really high confidence when inferring licenses from text
+confidence-threshold = 0.92
+allow = [
```
jonhoo, line 5: Why this particular list?

tower-service/Cargo.toml
LucioFranco: Cargo deny was complaining, duplicate bytes versions.

tower/Cargo.toml
LucioFranco: Same with the above, duplicate versions of rand.

deny.toml
LucioFranco: I copied mostly from tonic but should include a list of licenses we accept.

LucioFranco: This can be expanded though.

jonhoo: If it helps, here is flurry's [`deny.toml`](https://github.com/jonhoo/flurry/blob/master/deny.toml).

.github/workflows/CI.yml
LucioFranco: There is but we have a lot of warnings so I am going to try and fix that in another PR.

.github/workflows/CI.yml
LucioFranco: uhhhhhhh not sure!

.github/workflows/CI.yml
LucioFranco: not totally sure will replace that

.github/workflows/CI.yml
LucioFranco: doesn't seem possible

PR #438, opened 2020-04-15
limit: Fix rate limit not reseting state correctly

LucioFranco:
We were running into an issue in `Vector` where the rate limit remaining count never got reset and we always hit the branch that sets the limited state. This was because we were double over writing the state. In the first if statement we would `self.state = State::Ready { .. }` but then we would go to the next if statement and if the remaining was larger than 1 we would reset the state back to the original status with one more remaining subtracted. This would then override the fact that we were resetting the state in the previous if statement. This would then cause us to only reset the state in the poll_ready where we now would have to check the timer. This result of this is that we were probably rate limiting more than expected and logging the trace log much more than expected.

cc @lukesteensen 

### DIFF
--- a/tower-limit/src/rate/service.rs
+++ b/tower-limit/src/rate/service.rs
@@ -1,5 +1,5 @@
 use super::{error::Error, future::ResponseFuture, Rate};
-use futures::{try_ready, Future, Poll};
+use futures::{Async, Future, Poll};
 use tokio_timer::{clock, Delay};
 use tower_service::Service;
 
@@ -65,7 +65,9 @@ where
         match self.state {
             State::Ready { .. } => return self.inner.poll_ready().map_err(Into::into),
             State::Limited(ref mut sleep) => {
-                try_ready!(sleep.poll());
+                if let Async::NotReady = sleep.poll()? {
+                    tracing::trace!("rate limited exceeded; sleeping.");
+                }
             }
         }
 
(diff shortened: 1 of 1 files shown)


LucioFranco:
All tests pass locally I am not sure why we aren't seeing tests in here.

tower-limit/src/rate/service.rs
```diff
@@ -65,7 +65,9 @@ where
         match self.state {
             State::Ready { .. } => return self.inner.poll_ready().map_err(Into::into),
             State::Limited(ref mut sleep) => {
-                try_ready!(sleep.poll());
+                if let Async::NotReady = sleep.poll()? {
+                    tracing::trace!("rate limited exceeded; sleeping.");
```
LucioFranco, line 69: @lukesteensen I like this better, what do you think?

LucioFranco:
I am also unsure how to actually test this behavior, technically speaking the only issue here is extra logging. The actual behavior was correct.

tower-limit/src/rate/service.rs
lukesteensen: A little late on this review, but I think this actually broke the whole thing :smile: 

Before, `try_ready!` would return `NotReady` if we still needed to sleep. Now we're just checking if we should still be sleeping, logging that we should, and continuing on and marking ourselves `Ready`.

PR #440, opened 2020-04-15
limit: Prepare `0.1.2` release

LucioFranco:


tower-limit/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+# 0.1.2 (April 15, 2020)
+
+- Fixed state not getting reset correctly.
```
hawkw, line 3: nit: could we change this to actually describe the bug that was fixed? this is not super clear what the user impact is, IMO

PR #443, opened 2020-04-17
Change Discover to be a sealed trait

jonhoo:
`Discover` was _really_ just a `TryStream<Item = Change>`, so this
change makes that much clearer. Specifically, users are intended to use
`Discover` only in bounds, whereas implementors should implement
`Stream` with the appropriate `Item` type. `Discover` then comes with a
blanket implementation for anything that implements `TryStream`
appropriately. This obviates the need for the `discover::stream` module.


jonhoo:
I'd like to get @olix0r's take on this too, since he's the one (as far as I know) who's been most involved with `Discover`.

olix0r:
The intent for `Discover`, as differentiated from Stream, which exposes optionality. 

All that said, I don't think it really matters. It can be up to the consumer what to do when the stream ends...

jonhoo:
Yeah, I think it makes more sense for the consumer to be able to distinguish between `Poll::Ready(None)` and `Poll::Pending`. They may decide to treat the two the same, or they may decide that one can be handled more efficiently by not polling again in the future.

PR #445, opened 2020-04-17
Tidy up tower::load

jonhoo:
This also renames the `Instrument` trait, and related types, to better
reflect what they do. Specifically, the trait is now called
`TrackCompletion`, and `NoInstrument` is called `CompleteOnResponse`.

This also brings back balance example and makes it compile.


tower/src/load/completion.rs
```diff
@@ -0,0 +1,91 @@
...
+    type Output;
+
+    /// Attaches a `H`-typed handle to a `V`-typed value.
+    fn track_completion(&self, handle: H, value: V) -> Self::Output;
+}
+
+/// A `TrackCompletion` implementation that considers the request completed when the response
+/// future is resolved.
+#[derive(Clone, Copy, Debug)]
+pub struct CompleteOnResponse;
```
LucioFranco, line 42: impl default and a `_p: ()` for backwards compat?

jonhoo: The naming change already means it's not backwards compatible :p
Why is it you want a private field? To avoid people being able to match on it? Could we use `#[non_exhaustive]` instead?

LucioFranco: Yeah, just seems like a common pattern that we do in general 🤷

jonhoo: `#[non_exhaustive]` _would_ bump msrv to 1.40, but I think that's generally the preferred way to do this?

hawkw: FWIW, `_p: ()` does more than _just_ preventing matching --- it also prevents user code from writing out a struct literal and requires the use of a constructor to get an instance of the type. 

In general, I think this can be a good practice for structs with no fields if there is the potential of them growing a field in the future, since adding the field would break any existing struct literals. If we can guarantee with some confidence that we won't need to ever add fields to this, it may not be necessary.

hawkw: Also (I'm not _sure_ about this OTTOMH) I _think_ that `#[non_exhaustive]` is only meaningful when applied to enum types. I think it specifically effects how _variants_ are matched. I don't know if it does anything here.

jonhoo: `#[non_exhaustive]` does all of the things you listed above, and it works for anything "struct-like" (including unit structs) :)

I _do_ think that there may be a decent argument for us never adding fields to `CompleteOnResponse` though, in which case the annotation wouldn't be needed.

hawkw: > `#[non_exhaustive]` does all of the things you listed above, and it works for anything "struct-like" (including unit structs)

TIL! Never mind!

PR #449, opened 2020-04-24
Spring cleaning for tower::balance

jonhoo:
Noteworthy changes:

 - All constructors now follow the same pattern: `new` uses OS entropy,
   `from_rng` takes a `R: Rng` and seeds the randomness from there.
   `from_rng` is fallible, since randomness generators can be fallible.
 - `BalanceLayer` was renamed to `MakeBalanceLayer`, since it is not
   _really_ a `BalanceLayer`. The name of `BalanceMake` was also
   "normalized" to `MakeBalance`.

Another observation: the `Debug` bound on `Load::Metric` in
`p2c::Balance`, while not particularly onerous, generates really
confusing errors if you forget it include it. And crucially, the error
never points at `Debug` (should we file a compiler issue?), so I pretty
much had to guess my way to that being wrong in the doc example.

It would probably be useful to add a documentation example to
`MakeBalanceLayer` or `MakeBalance` (I suspect just one of them is fine,
since they're basically the same). Since I've never used it, and find it
hard to think of uses for it, it might be good if someone with more
experience with it wrote one.

### DIFF
--- a/tower/src/balance/mod.rs
+++ b/tower/src/balance/mod.rs
@@ -1,4 +1,57 @@
-//! Load balancing middlewares.
+//! Middleware that allows balancing load among multiple services.
+//!
+//! In larger systems, multiple endpoints are often available for a given service. As load
+//! increases, you want to ensure that that load is spread evenly across the available services.
+//! Otherwise, clients could see spikes in latency if their request goes to a particularly loaded
+//! service, even when spare capacity is available to handle that request elsewhere.
+//!
+//! This module provides two pieces of middleware that helps with this type of load balancing:
+//!
+//! First, [`p2c`] implements the "[Power of Two Random Choices]" algorithm, a simple but robust
+//! technique for spreading load across services with only inexact load measurements. Use this if
+//! the set of available services is not within your control, and you simply want to spread load
+//! among that set of services.
+//!
+//!   [Power of Two Random Choices]: http://www.eecs.harvard.edu/~michaelm/postscripts/handbook2001.pdf
+//!
+//! Second, [`pool`] implements a dynamically sized pool of services. It estimates the overall
+//! current load by tracking successful and unsuccessful calls to `poll_ready`, and uses an
+//! exponentially weighted moving average to add (using [`tower::make_service::MakeService`]) or
+//! remove (by dropping) services in response to increases or decreases in load. Use this if you
+//! are able to dynamically add more service endpoints to the system to handle added load.
+//!
+//! # Examples
+//!
+//! ```rust
+//! # #[cfg(feature = "util")]
+//! # #[cfg(feature = "load")]
+//! # fn warnings_are_errors() {
+//! use tower::balance::p2c::Balance;
+//! use tower::load::Load;
+//! use tower::{Service, ServiceExt};
+//! use futures_util::pin_mut;
+//! # use futures_core::Stream;
+//! # use futures_util::StreamExt;
+//!
+//! async fn spread<Req, S: Service<Req> + Load>(svc1: S, svc2: S, reqs: impl Stream<Item = Req>)
+//! where
+//!     S::Error: Into<tower::BoxError>,
+//! # // this bound is pretty unfortunate, and the compiler does _not_ help
+//!     S::Metric: std::fmt::Debug,
+//! {
+//!     // Spread load evenly across the two services
+//!     let p2c = Balance::new(tower::discover::ServiceList::new(vec![svc1, svc2]));
+//!
+//!     // Issue all the requests that come in.
+//!     // Some will go to svc1, some will go to svc2.
+//!     pin_mut!(reqs);
+//!     let mut responses = p2c.call_all(reqs);
+//!     while let Some(rsp) = responses.next().await {
+//!         // ...
+//!     }
+//! }
+//! # }
+//! ```
 
 pub mod error;
 pub mod p2c;
--- a/tower/src/balance/mod.rs
+++ b/tower/src/balance/mod.rs
@@ -12,7 +12,7 @@
 //! the set of available services is not within your control, and you simply want to spread load
 //! among that set of services.
 //!
-//!   [Power of Two Random Choices]: http://www.eecs.harvard.edu/~michaelm/postscripts/handbook2001.pdf
+//! [Power of Two Random Choices]: http://www.eecs.harvard.edu/~michaelm/postscripts/handbook2001.pdf
 //!
 //! Second, [`pool`] implements a dynamically sized pool of services. It estimates the overall
 //! current load by tracking successful and unsuccessful calls to `poll_ready`, and uses an
--- a/tower/src/balance/p2c/layer.rs
+++ b/tower/src/balance/p2c/layer.rs
@@ -8,10 +8,10 @@ use tower_layer::Layer;
 ///
 /// This construction may seem a little odd at first glance. This is not a layer that takes
 /// requests and produces responses in the traditional sense. Instead, it is more like
-/// [`MakeService`](tower::make_service::MakeService) in that it takes service _descriptors_ and
-/// produces _services_. Since [`Balance`] spreads requests across a _set_ of services, the inner
-/// service should produce a [`Discover`], not just a single [`Service`], given a service
-/// descriptor.
+/// [`MakeService`](tower::make_service::MakeService) in that it takes service _descriptors_ (see
+/// `Target` on `MakeService`) and produces _services_. Since [`Balance`] spreads requests across a
+/// _set_ of services, the inner service should produce a [`Discover`], not just a single
+/// [`Service`], given a service descriptor.
 ///
 /// See the [module-level documentation](..) for details on load balancing.
 #[derive(Clone)]
--- a/tower/src/balance/p2c/layer.rs
+++ b/tower/src/balance/p2c/layer.rs
@@ -1,17 +1,27 @@
-use super::BalanceMake;
+use super::MakeBalance;
 use rand::{rngs::SmallRng, Rng, SeedableRng};
 use std::{fmt, marker::PhantomData};
 use tower_layer::Layer;
 
-/// Efficiently distributes requests across an arbitrary number of services
+/// Contruct load balancers ([`Balance`]) over dynamic service sets ([`Discover`]) produced by the
+/// "inner" service in response to requests coming from the "outer" service.
+///
+/// This construction may seem a little odd at first glance. This is not a layer that takes
+/// requests and produces responses in the traditional sense. Instead, it is more like
+/// [`MakeService`](tower::make_service::MakeService) in that it takes service _descriptors_ and
+/// produces _services_. Since [`Balance`] spreads requests across a _set_ of services, the inner
+/// service should produce a [`Discover`], not just a single [`Service`], given a service
+/// descriptor.
+///
+/// See the [module-level documentation](..) for details on load balancing.
 #[derive(Clone)]
-pub struct BalanceLayer<D, Req> {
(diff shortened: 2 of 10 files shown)


tower/src/balance/p2c/make.rs
```diff
@@ -12,15 +12,22 @@ use std::{
 };
 use tower_service::Service;
 
-/// Makes `Balancer`s given an inner service that makes `Discover`s.
+/// Contruct load balancers over dynamic service sets produced by a wrapped "inner" service.
```
LucioFranco, line 15:
```suggestion
/// Construct load balancers over dynamic service sets produced by a wrapped "inner" service.
```

tower/src/balance/p2c/layer.rs
```diff
@@ -1,25 +1,35 @@
...
-/// Efficiently distributes requests across an arbitrary number of services
+/// Construct load balancers ([`Balance`]) over dynamic service sets ([`Discover`]) produced by the
+/// "inner" service in response to requests coming from the "outer" service.
+///
+/// This construction may seem a little odd at first glance. This is not a layer that takes
+/// requests and produces responses in the traditional sense. Instead, it is more like
+/// [`MakeService`](tower::make_service::MakeService) in that it takes service _descriptors_ and
+/// produces _services_. Since [`Balance`] spreads requests across a _set_ of services, the inner
+/// service should produce a [`Discover`], not just a single [`Service`], given a service
+/// descriptor.
```
hawkw, line 14: in Linkerd we tend to refer to "service descriptors" (and, in general, all `Request` types for a `MakeService`) as "targets". Might be good to standardize terminology for this?

this is also what the definition of `MakeService` calls them:
https://github.com/tower-rs/tower/blob/6a25d322b5160c1033e8741ff5c12b175a83bc21/tower/src/make/make_service.rs#L15

tower/src/balance/mod.rs
```diff
@@ -1,4 +1,57 @@
...
+//! service, even when spare capacity is available to handle that request elsewhere.
+//!
+//! This module provides two pieces of middleware that helps with this type of load balancing:
+//!
+//! First, [`p2c`] implements the "[Power of Two Random Choices]" algorithm, a simple but robust
+//! technique for spreading load across services with only inexact load measurements. Use this if
+//! the set of available services is not within your control, and you simply want to spread load
+//! among that set of services.
+//!
+//!   [Power of Two Random Choices]: http://www.eecs.harvard.edu/~michaelm/postscripts/handbook2001.pdf
```
hawkw, line 15: nit: is it standard to indent these link references?
```suggestion
//! [Power of Two Random Choices]: http://www.eecs.harvard.edu/~michaelm/postscripts/handbook2001.pdf
```

tower/src/balance/mod.rs
```diff
@@ -1,4 +1,57 @@
...
+//! current load by tracking successful and unsuccessful calls to `poll_ready`, and uses an
+//! exponentially weighted moving average to add (using [`tower::make_service::MakeService`]) or
+//! remove (by dropping) services in response to increases or decreases in load. Use this if you
+//! are able to dynamically add more service endpoints to the system to handle added load.
+//!
+//! # Examples
+//!
+//! ```rust
+//! # #[cfg(feature = "util")]
+//! # #[cfg(feature = "load")]
```
hawkw, line 27: Does this actually work? My understanding is that the doctests are compiled in a separate crate and the crate's feature flags are not passed to it, so I _believe_ this means the doctest will _never_ be compiled

tower/src/balance/p2c/make.rs
```diff
@@ -12,15 +12,22 @@ use std::{
 };
 use tower_service::Service;
 
-/// Makes `Balancer`s given an inner service that makes `Discover`s.
+/// Construct load balancers over dynamic service sets produced by a wrapped "inner" service.
+///
+/// This is effectively an implementation of [`MakeService`](tower::make_service::MakeService),
+/// except that it forwards the service descriptors (`Target`) to an inner service (`S`), and
+/// expects that service to produce a service set in the form of a [`Discover`]. It then wraps the
+/// service set in a [`Balance`] before returning it as the "made" service.
```
hawkw, line 20: The links to `Discover` and `Balance` don't have references...do they actually work? I didn't think RustDoc would link type names in backticks without a reference.

tower/src/balance/p2c/make.rs
```diff
@@ -12,15 +12,22 @@ use std::{
 };
 use tower_service::Service;
 
-/// Makes `Balancer`s given an inner service that makes `Discover`s.
+/// Construct load balancers over dynamic service sets produced by a wrapped "inner" service.
```
hawkw, line 15: nit: I think this should be
```suggestion
/// Constructs load balancers over dynamic service sets produced by a wrapped "inner" service.
```

tower/src/balance/p2c/mod.rs
```diff
@@ -1,4 +1,32 @@
-//! A Power-of-Two-Choices Load Balancer
+//! This module implements the "[Power of Two Random Choices]" load balancing algorithm.
+//!
+//! It is a simple but robust technique for spreading load across services with only inexact load
+//! measurements. As its name implies, whenever a request comes in, it samples two ready services
+//! at random, and issues the request to whichever service is less loaded. How loaded a service is
+//! is determiend by the return value of [`Load`](tower::load::Load).
```
hawkw, line 6: typo:
```suggestion
//! is determined by the return value of [`Load`](tower::load::Load).
```

tower/src/balance/p2c/mod.rs
```diff
@@ -1,4 +1,32 @@
...
+//! > The algorithm randomly picks two services from the set of ready endpoints and
+//! > selects the least loaded of the two. By repeatedly using this strategy, we can
+//! > expect a manageable upper bound on the maximum load of any server.
+//! >
+//! > The maximum load variance between any two servers is bound by `ln(ln(n))` where
+//! > `n` is the number of servers in the cluster.
+//!
+//! The balance service and layer implementations rely on you providing the underlying set of
+//! services to balance requests across. They do so using the
+//! [`Discover`](tower::discover::Discover) trait, which is essentially a `Stream` that indicates
```
hawkw, line 19: Nit/TIOLI: I might consider introducing the term "service discovery" along with the discover trait? Maybe something like 
```suggestion
//! The balance service and layer implementations rely on _service discovery_ information providing
//! the underlying set of services to balance requests across. Service discovery is represented by
//! the [`Discover`](tower::discover::Discover) trait, which is essentially a `Stream` that indicates
```

tower/src/balance/p2c/mod.rs
```diff
@@ -1,4 +1,32 @@
...
+//! when services become available or go away. If you have a fixed set of services, considering
```
hawkw, line 20: typo:
```suggestion
//! when services become available or go away. If you have a fixed set of services, consider
```

tower/src/balance/p2c/service.rs
```diff
@@ -18,24 +18,15 @@ use tokio::sync::oneshot;
...
-/// > expect a manageable upper bound on the maximum load of any server.
-/// >
-/// > The maximum load variance between any two servers is bound by `ln(ln(n))` where
-/// > `n` is the number of servers in the cluster.
+/// See the [module-level documentation](..) for details.
 ///
 /// Note that `Balance` requires that the `Discover` you use is `Unpin` in order to implement
 /// `Service`. This is because it needs to be accessed from `Service::poll_ready`, which takes
 /// `&mut self`. You can achieve this easily by wrapping your `Discover` in [`Box::pin`] before you
 /// construct the `Balance` instance. For more details, see [#319].
```
hawkw, line 28: probably best saved for a follow-up PR, but should `Discover` just have an `Unpin` bound on it? Since the unpin bound is only on the `Service` impl for `Balance`, this will currently fail _very_ late, when a user tries to wrap a `Balance` in another layer or actually call a `Balance`, and it probably won't be super clear why `Service` isn't implemented. If `Discover` required `Self: Unpin`, the compiler error would occur when the user writes a bad `Discover` impl, which would be much clearer.

tower/src/balance/p2c/service.rs
```diff
@@ -94,10 +85,10 @@ where
     D::Service: Service<Req>,
     <D::Service as Service<Req>>::Error: Into<crate::BoxError>,
 {
-    /// Initializes a P2C load balancer from the provided randomization source.
-    pub fn new(discover: D, rng: SmallRng) -> Self {
+    /// Construct a load balancer that uses operating system entropy.
```
hawkw, line 88: Nit/TIOLI: I think these are usually written in the form
```suggestion
    /// Constructs a load balancer that uses operating system entropy.
```

tower/src/balance/p2c/service.rs
```diff
@@ -68,10 +59,10 @@ where
     }
 }
 
-#[pin_project]
 /// A Future that becomes satisfied when an `S`-typed service is ready.
 ///
-/// May fail due to cancelation, i.e. if the service is removed from discovery.
+/// May fail due to cancelation, i.e., if the service is removed from discovery.
```
hawkw, line 64: maybe worth linking to something in the `discover` module that's relevant to cancelation?

jonhoo:
@olix0r The bound is there just for tracing really, so that we can print the actual measured load. And I think that's a pretty valuable purpose. Really not sure what to do about that.

tower/src/balance/p2c/layer.rs
jonhoo: Yeah, I went with "service descriptor" here because I feel as though "target" is a hopefully non-descriptive name. It doesn't tell me anything about what that parameter _does_. It's true that it _might_ be a target (like an IP address), but I don't think that is _generally_ true. I'd actually be in favor of renaming target everywhere, but that's a longer discussion :sweat_smile: 

In the meantime, how would you feel about putting ("the `Target`") in parentheses?

tower/src/balance/mod.rs
jonhoo: Huh, looking at the spec apparently not. Not sure where I picked that up from, I've just always done it that way. I'll undo.

tower/src/balance/mod.rs
jonhoo: Yup, this works. I mentioned this in https://discordapp.com/channels/500028886025895936/500336346770964480/694636869656510484. See also https://github.com/rust-lang/cargo/issues/8060.

tower/src/balance/p2c/make.rs
jonhoo: Yup: https://github.com/rust-lang/rust/issues/43466. It's on by default on nightly, which is what docsrs uses.

tower/src/balance/p2c/service.rs
jonhoo: I think we should avoid `Unpin` bounds _wherever possible_. They are really sad, especially when we one day get generators that implement `Stream`. This also ties into the larger discussion in #319.

hawkw: Hmm, but if `Discover` is going to be polled in `poll_ready`, it _will_ always have to be `Unpin` unless `poll_ready` is changed to take a `Pin<&mut Self>` receiver. In that case, we would be making a breaking change anyway and could remove an `Unpin` bound from `Discover`.

jonhoo: I don't know if `Discover` will only ever be polled in `poll_ready` though? But you're right that if `poll_ready` later changes to be `Pin`, we could remove the bound then (although I think _removing_ the bound is backwards-compatible anyway?).

PR #462, opened 2020-06-24
port tower-buffer over to generic error approach

yaahc:


tower/src/buffer/error.rs
```diff
@@ -1,47 +1,12 @@
...
-    pub(crate) fn clone(&self) -> ServiceError {
-        ServiceError {
-            inner: self.inner.clone(),
-        }
-    }
-}
-
-impl fmt::Display for ServiceError {
-    fn fmt(&self, fmt: &mut fmt::Formatter) -> fmt::Result {
-        write!(fmt, "buffered service failed: {}", self.inner)
```
yaahc, line 35: formatting the source error in this errors Display impl and also returning the source instead of your source's source in the Error impl was a bug fwiw

yaahc:
The biggest issue with this is handling the inner service's error. Tower-buffer used to handle this by converting it to an error that has an Arc internally to make the error clone, and then type erasing _that_ error so it looks kinda sorta like the same error from a distance.

However, the core issue I'm trying to fix is forcing errors to go through `Box`, forcing them to go through `Arc` would be just as bad, the only way to service both concerns that I could come up with is to just push the complexity up the stack by requiring that the inner error type implements `Clone`. In the test code I reworked the mock error type to be an Arc based error type instead of Box based, and this is going to present issues for `eyre` as well because it isn't clone by default, though I'm hoping I can just add a feature that converts its internal boxes into arcs and it will work out fine.

This however does mean that you cannot just use `Box<dyn Error>` everywhere like you used to, if the service you're wrapping in a `Buffer` returns a `Box<dyn Error>` it will not compile, you need to change that type to one that impls Clone. We might want to provide an error type from `tower-buffer` for convenience that implements `Clone`, we could probably just steal the one from `tower-test` and export it instead of the old `ServiceError`.

Overall not great, but I still think its preferable to the current state of things.

hawkw:
It looks like CI is failing because there's a `BoxError` type alias that's never actually defined?

tower/src/buffer/future.rs
```diff
@@ -4,47 +4,87 @@ use super::{error::Closed, message};
...
 #[pin_project(project = ResponseStateProj)]
-#[derive(Debug)]
-enum ResponseState<T> {
-    Failed(Option<crate::BoxError>),
-    Rx(#[pin] message::Rx<T>),
-    Poll(#[pin] T),
+enum ResponseState<S, E2, Response>
+where
+    S: crate::Service<Response>,
+{
```
hawkw, line 41: Take it or leave it: I think that if this struct was generic over type parameters for the service's future and error types, rather than being generic over a `Service`, it would still be able to derive `Debug` (it's the `Response` type we don't care about implementing `Debug`), while still having the same number of type parameters (`T, E, E2` rather than `S, E2, Response`). Also, we wouldn't need to stick as many bounds everywhere for it.

tower/src/buffer/layer.rs
```diff
@@ -9,12 +9,13 @@ use tower_service::Service;
 /// which means that this layer can only be used on the Tokio runtime.
 ///
 /// See the module documentation for more details.
-pub struct BufferLayer<Request> {
+pub struct BufferLayer<Request, E2> {
     bound: usize,
     _p: PhantomData<fn(Request)>,
+    _e: PhantomData<E2>,
```
hawkw, line 15: This should be in a `fn` so that it doesn't effect the service's `Send`/`Sync`/etc impls.

hawkw, line 15: also i would probably just have one `PhantomData` for everything, 
```suggestion
    _p: PhantomData<fn(Request, E2)>,
```
having two phantom fields just adds noise IMO

yaahc:
ACK. I'll fix that in a sec,

tower/src/buffer/service.rs
```diff
@@ -13,18 +13,20 @@ use tower_service::Service;
 ///
 /// See the module documentation for more details.
 #[derive(Debug)]
-pub struct Buffer<T, Request>
+pub struct Buffer<S, Request, E2>
```
hawkw, line 16: Can `E2` have a default so that users who don't care about overriding `Box<dyn Error>` don't have to think about it?

seanmonstar:
> the core issue I'm trying to fix is forcing errors to go through Box, forcing them to go through Arc would be just as bad, the only way to service both concerns that I could come up with is to just push the complexity up the stack by requiring that the inner error type implements Clone.

Why is putting the error in an Arc so bad?

I think errors usually aren't `Clone`, so it that requirement would suddenly mean most services can't use `Buffer`.

tower/src/buffer/worker.rs
```diff
@@ -70,10 +70,11 @@ where
     ///
     /// If a `Message` is returned, the `bool` is true if this is the first time we received this
     /// message, and false otherwise (i.e., we tried to forward it to the backing service before).
+    #[allow(clippy::type_complexity)]
```
hawkw, line 73: i wonder if we can break this up a little with a type error?

tower/src/buffer/worker.rs
```diff
@@ -22,36 +20,38 @@ use tower_service::Service;
 /// implement (only call).
 #[pin_project]
 #[derive(Debug)]
-pub struct Worker<T, Request>
+pub struct Worker<S, Request, E2>
```
hawkw, line 23: Similarly, can this have a default so that I can still write `Worker<S, Request>` if I'm not overriding the error type?

tower/src/buffer/worker.rs
```diff
@@ -22,36 +20,38 @@ use tower_service::Service;
...
+    _e: PhantomData<E2>,
 }
 
-impl<T, Request> Worker<T, Request>
+impl<S, Request, E2> Worker<S, Request, E2>
 where
-    T: Service<Request>,
-    T::Error: Into<crate::BoxError>,
+    S: Service<Request>,
+    S::Error: Into<E2> + Clone,
```
hawkw, line 46: I'm really not a fan of adding a `Clone` bound here...in general, I don't think that a majority of the errors being used here will necessarily implement `Clone`, and this feels like it's essentially just punting the complexity of adding the `Arc` to the user?

hawkw: Also, I don't actually see where the `Clone` impl gets used? AFAICT we don't actually need this?

tower/src/buffer/error.rs
hawkw: FWIW, I think this is because the `Display` impl may have predated `Error::source` being useable...

tower/src/buffer/service.rs
yaahc: I'm not sure if we could default it in a meaningful way, the default only comes into play when you're using the constructor itself, not on any functions like `new` that then construct the `Buffer` internally.

yaahc:
> Why is putting the error in an Arc so bad?

@seanmonstar  I think the easiest way to demonstrate this is with how it would interfere with the way I want to use `eyre::Report` with `tower-buffer`.

My goal is to be able to write services that use `eyre::Report` that are then wrapped in a `tower-buffer`. With the move to generic error types we need to add the following bounds to Buffer `S::Error: Into<E2>, crate::buffer::error::Closed: Into<E2>`. These let us convert the inner error to the outer one.

However, if we're going to introduce the ability to be cloned to the inner error type with `Arc` instead of adding a `Clone` bound we end up needing to satisfy `Arc<S::Error>: Into<E2>` or we need to always return an `Arc<E2>` from the buffer itself. This would mean that the buffered service now returns an `Arc<eyre::Report>` which is no longer mutable, so we cant ever add more information to the error report or wrap the inner error type, which I'm pretty sure would cause issues for us down the line.

tower/src/buffer/worker.rs
yaahc: it gets used in `process_req`

hawkw:
> This would mean that the buffered service now returns an Arc<eyre::Report> which is no longer mutable, so we cant ever add more information to the error report or wrap the inner error type

Hmm, in general, Rust tends to build errors through composition, right? If an `Arc<eyre::Report>` still implements whatever traits are required to compose with other errors, can we not still compose other errors around it regardless of mutability?

tower/src/buffer/worker.rs
yaahc: I don't follow

tower/src/buffer/service.rs
hawkw: Yeah, but it also shows up if I want to write code that has a `Buffer` as a field or returns a `Buffer`.

hawkw: I might also want to consider having `new` fill in the default, and have a separate constructor for overriding the error type...but I could be convinced otherwise.

yaahc:
> Hmm, in general, Rust tends to build errors through composition, right? If an Arc<eyre::Report> still implements whatever traits are required to compose with other errors, can we not still compose other errors around it regardless of mutability?

This would be true except for the fact that `eyre::Report` is not an error in the same way `Box<dyn Error>` isn't an error and therefore they don't compose because they cannot implement std::error::Error, and eyre::Report contains additional context intended only for the end report that as of yet cannot be exposed via the error trait for further composition. Some day if we ever end up merging my generic member access RFC for the error trait this wont be an issue and it will be fine to compose reports, for now though the limitations of the error trait mean that if we want to capture extra context for reporting like notes and suggestions or custom sections in the output we can't force the error type of a service to itself be composable as  `std::error::Error`.

tower/src/buffer/service.rs
yaahc: > I might also want to consider having new fill in the default, and have a separate constructor for overriding the error type...but I could be convinced otherwise.

I have no desire to convince you otherwise, this sounds like a great idea.

yaahc: And added in the last commit

tower/src/buffer/future.rs
yaahc: When I try doing this I end up needing 4 parameters `Fut, T, E, E2` and it gives me an error about T being unused, which means I need to shove a phantom data into the ResponseState enum somehow I think but I'm not really sure how best to do so.

hawkw:
> eyre::Report contains additional context intended only for the end report that as of yet cannot be exposed via the error trait for further composition

If I understand this correctly, `eyre::Report` is only intended to be constructed by the scope that will be ultimately responsible for consuming the error (e.g. logging it, generating an error message, returning an HTTP error response, etc)? If that's the case, then a `Service` implementation should probably _never_ return an error reporting type, right? In general, a middleware service should expect its errors to be propagated through scopes that should be able to compose them with other errors --- I would expect a design using error reporting types should have a single service somewhere that consumes all lower-level error and produces error reporting types?

Let me know if I'm misunderstanding --- just trying to think about how this model fits into `tower`.

tower/src/buffer/future.rs
hawkw: What is `T` in that case? Previously, `T` was the future type, which appears twice in the enum...I don't see where the 4th param is coming from?

yaahc:
> If I understand this correctly, eyre::Report is only intended to be constructed by the scope that will be ultimately responsible for consuming the error (e.g. logging it, generating an error message, returning an HTTP error response, etc)? If that's the case, then a Service implementation should probably never return an error reporting type, right? In general, a middleware service should expect its errors to be propagated through scopes that should be able to compose them with other errors --- I would expect a design using error reporting types should have a single service somewhere that consumes all lower-level error and produces error reporting types?

I replied to this in discord but the gist of my reply was that its not meant to be constructed right before reporting, its meant to be constructed the moment you don't think anything else will be done other than reporting it, but as deeply in the stack as possible (or as close to the source error as possible) so that it can capture relevant context like backtrace frames and tracing spans via spantrace, and so that while its propagating back out of the stack more context can be added to the report as it becomes available.

yaahc:
Also, the `Clone` or not `Clone` bound issue is a moot point apparently. I tried testing out making `eyre::Report` clone by making it use `Arc`s internally and it broke most of the core functionality, so making this `Clone` doesn't even solve the problem I originally set out to solve. I'll have to figure out a way to deal with that, until then I'm gonna close this so it doesn't pollute the PR queue.

hawkw:
> > If I understand this correctly, eyre::Report is only intended to be constructed by the scope that will be ultimately responsible for consuming the error (e.g. logging it, generating an error message, returning an HTTP error response, etc)? If that's the case, then a Service implementation should probably never return an error reporting type, right? In general, a middleware service should expect its errors to be propagated through scopes that should be able to compose them with other errors --- I would expect a design using error reporting types should have a single service somewhere that consumes all lower-level error and produces error reporting types?
> 
> I replied to this in discord but the gist of my reply was that its not meant to be constructed right before reporting, its meant to be constructed the moment you don't think anything else will be done other than reporting it, but as deeply in the stack as possible (or as close to the source error as possible) so that it can capture relevant context like backtrace frames and tracing spans via spantrace, and so that while its propagating back out of the stack more context can be added to the report as it becomes available.

Yeah, I was kind of misunderstanding the intention of erorr reporting types like `eyre::Report`, thanks for clearing that up.

PR #467, opened 2020-08-29
ready-cache: Properly expose error source

olix0r:
The `ready_cache::error::Failed` type does not actually expose its inner
error type via `Error::source`. This change fixes this and improves
debug formatting.

### DIFF
--- a/tower/src/ready_cache/error.rs
+++ b/tower/src/ready_cache/error.rs
@@ -6,9 +6,12 @@ pub struct Failed<K>(pub K, pub crate::BoxError);
 
 // === Failed ===
 
-impl<K> std::fmt::Debug for Failed<K> {
+impl<K: std::fmt::Debug> std::fmt::Debug for Failed<K> {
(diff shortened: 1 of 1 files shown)


tower/src/ready_cache/error.rs
```diff
@@ -6,9 +6,12 @@ pub struct Failed<K>(pub K, pub crate::BoxError);
 
 // === Failed ===
 
-impl<K> std::fmt::Debug for Failed<K> {
+impl<K: std::fmt::Debug> std::fmt::Debug for Failed<K> {
```
jonhoo, line 9: Note that this is a backwards-incompatible change. So would have to be slated for next major release (which I guess we're in the process of preparing probably). cc @LucioFranco

hawkw: we are, in fact (and linkerd already depends on tower master)

LucioFranco: 👍 I am hoping to hop back on after tokio 0.3 so fine to push this onto master

PR #472, opened 2020-09-22
hedge: don't reserve slots for hedged requests

hdevalence:
`Hedge` wraps `S: Service + Clone` and is essentially implemented as
`Select<S, Delay<S>>`, whose `call` implementation creates two response
futures and selects the one which resolves first.  The second response
future is obtained from `Delay<S>::call`, which waits for some time
before calling `S::call`.

The current implementation of `Delay<S>::poll_ready` calls
`S::poll_ready` unconditionally.  However, the `S::call` corresponding
to that `poll_ready` only happens after much delay, if ever.  This
interacts badly with `Buffer`, which uses `poll_ready` to reserve a
channel slot.  When using `Hedge` with a `Buffer`'d service, the inner
`Buffer::poll_ready` is called twice, but only one slot is used
immediately.  This causes exhaustion of all of the `Buffer` slots.

Instead, return `Poll::Ready(Ok(()))` from `Delay::poll_ready` and use a
`Oneshot` inside of the `Delay` future.


hawkw:
Just to confirm, this fixes an issue in Zebra, correct?

hdevalence:
Yes, or it would, if the code that uses it was merged.  The Zebra context is here: https://github.com/ZcashFoundation/zebra/pull/1089

hawkw (approved):
this seems right to me!

PR #474, opened 2020-10-19
Upgrade to tokio 0.3

### LINKED ISSUE #473: update to tokio v0.3

LucioFranco:
- Upgrade runtime aspects to 0.3 (time)
- Continue to use tokio 0.2 sync until we can get the correct workarounds
- Move examples out of the crate and minor clean up
- Add `full` feature flag

Closes #473

### DIFF
--- a/tower/src/spawn_ready/future.rs
+++ b/tower/src/spawn_ready/future.rs
@@ -8,13 +8,13 @@ use std::{
     pin::Pin,
     task::{Context, Poll},
 };
-use tokio::sync::oneshot;
+use tokio02::sync::oneshot;
 use tower_service::Service;
 
 /// Drives a service to readiness.
 #[pin_project]
 #[derive(Debug)]
-pub struct BackgroundReady<T, Request> {
+pub(crate) struct BackgroundReady<T, Request> {
     service: Option<T>,
     tx: Option<oneshot::Sender<Result<T, crate::BoxError>>>,
     _req: PhantomData<Request>,
--- a/tower/src/spawn_ready/mod.rs
+++ b/tower/src/spawn_ready/mod.rs
@@ -1,7 +1,7 @@
 //! When an underlying service is not ready, drive it to readiness on a
 //! background task.
 
-pub mod future;
+mod future;
 mod layer;
 mod make;
 mod service;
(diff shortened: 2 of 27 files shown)


tower/src/spawn_ready/mod.rs
```diff
@@ -1,7 +1,7 @@
 //! When an underlying service is not ready, drive it to readiness on a
 //! background task.
 
-pub mod future;
+mod future;
```
jonhoo, line 4: Why make this no longer `pub`?

tower/src/spawn_ready/future.rs
```diff
@@ -8,13 +8,13 @@ use std::{
...
 };
-use tokio::sync::oneshot;
+use tokio02::sync::oneshot;
 use tower_service::Service;
 
 /// Drives a service to readiness.
 #[pin_project]
 #[derive(Debug)]
-pub struct BackgroundReady<T, Request> {
+pub(crate) struct BackgroundReady<T, Request> {
```
jonhoo, line 17: Why make this no longer `pub`?

jonhoo:
Can you link to the issues that track why we still need tokio `0.2`? Specifically I saw `sync` and `stream`?

tower/src/spawn_ready/mod.rs
LucioFranco: Mistake, will revert this

LucioFranco:
@jonhoo basically, mpsc/oneshots removed poll methods and because they are intrusive now it is quite difficult to make a poll based api for things. So to unblock this upgrade I am continuing to use `0.2` sync features while enabling tower to run on a 0.3 runtime.

I am not sure how I want to get around the changes for mpsc but when I do some unsafety miri complains. So another option is to just extract the old mpsc/oneshot and vendor it until we can get a proper fix.

jonhoo:
I'm really sad about the removal of `poll_` methods. Sometimes you just _have_ to implement `Future` manually, and then those are vital :cry: I think vendoring would probably be a mistake. It's too easy to let that code just sit around for a long time. And that extra complexity _will_ come back to bite us. I suppose bringing in `0.2` isn't _too_ bad as a stopgap given that we guard it with feature flags..

LucioFranco:
Closing this in favor of https://github.com/tower-rs/tower/pull/476

PR #476, opened 2020-10-26
update to Tokio 0.3

### LINKED ISSUE #473: update to tokio v0.3

hawkw:
This branch updates Tower to Tokio 0.3.

Unlike  #474, this branch uses Tokio 0.3's synchronization primitives,
rather than continuing to depend on Tokio 0.2. I think that we ought to
try to use Tokio 0.3's channels whenever feasible, because the 0.2
channels have pathological memory usage patterns in some cases (see
tokio-rs/tokio#2637). @LucioFranco let me know what you think of the
approach used here and we can compare notes!

For the most part, this was a pretty mechanical change: updating
versions in Cargo.toml, tracking feature flag changes, renaming
`tokio::time::delay` to `sleep`, and so on. Tokio's channel receivers
also lost their `poll_recv` methods, but we can easily replicate that by
enabling the `"stream"` feature and using `poll_next` instead.

The one actually significant change is that `tokio::sync::mpsc::Sender`
lost its `poll_ready` method, which impacts the way `tower::buffer` is
implemeted. When the buffer's channel is full, we want to exert
backpressure in `poll_ready`, so that callers such as load balancers
could choose to call another service rather than waiting for buffer
capacity. Previously, we did this by calling `poll_ready` on the
underlying channel sender.

Unfortunately, this can't be done easily using Tokio 0.3's bounded MPSC
channel, because it no longer exposes a polling-based interface, only an
`async fn ready`, which borrows the sender. Therefore, we implement our
own bounded MPSC on top of the unbounded channel, using a semaphore to
limit how many items are in the channel.

I factored out the code for polling a semaphore acquire future from
`limit::concurrency` into its own module, and reused it in `Buffer`.

Additionally, the buffer tests needed to be updated, because they
currently don't actually poll the buffer service before calling it. This
violates the `Service` contract, and the new code actually fails as a
result.

Closes #473 
Closes #474

Co-authored-by: Lucio Franco <luciofranco14@gmail.com>


tower/src/buffer/service.rs
```diff
@@ -87,12 +107,18 @@ where
     type Future = ResponseFuture<T::Future>;
 
     fn poll_ready(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
-        // If the inner service has errored, then we error here.
-        if let Err(_) = ready!(self.tx.poll_ready(cx)) {
-            Poll::Ready(Err(self.get_worker_error()))
-        } else {
-            Poll::Ready(Ok(()))
+        tracing::info!("poll");
```
LucioFranco, line 110: I had issues with this test previously, but also by the looks of these tracing statements you did too? Probably should remove these :)

tower/src/spawn_ready/future.rs
```diff
@@ -49,7 +49,11 @@ where
     fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
         let this = self.project();
 
-        if let Poll::Ready(_) = Pin::new(this.tx.as_mut().expect("illegal state")).poll_closed(cx) {
+        if let Poll::Ready(_) = {
+            let closed = this.tx.as_mut().expect("illegal state").closed();
+            tokio::pin!(closed);
+            closed.poll(cx)
+        } {
```
LucioFranco, line 56: nit: this is a bit hard to read, can we move out some of this logic from this block?

LucioFranco (changes requested):
LGTM just a few tracing things left over, ill close my PR :)

tower/src/buffer/service.rs
hawkw: oh whoops, didn't mean to leave those in!

hawkw:
> ill close my PR :)
@LucioFranco if there's anything you got that I missed, we could also merge one branch into the other?

LucioFranco:
@hawkw only last thing is my PR did some stuff with feature flags could you ensure you just copy that? beyond the tokio 0.2 sync stuff it fixed a few things and added a `full` flag.

tower/src/buffer/worker.rs
```diff
@@ -80,11 +80,11 @@ where
...
-        if let Some(mut msg) = self.current_message.take() {
-            // poll_closed returns Poll::Ready is the receiver is dropped.
-            // Returning Pending means it is still alive, so we should still
-            // use it.
-            if msg.tx.poll_closed(cx).is_pending() {
+        if let Some(msg) = self.current_message.take() {
+            // If the oneshot sender is closed, then the receiver is dropped,
+            // and nobody cares about the response. If this is the case, we
+            // should continue to the next request.
+            if !msg.tx.is_closed() {
```
hawkw, line 87: n.b. that `tokio` 0.3.2 is adding back `oneshot::Sender::poll_closed`, but I don't think we actually _need_ to poll here, since we immediately consume the sender if it isn't closed. we don't need to register interest in it closing, since we're not going to yield until we're done with that sender...so `is_closed` is probably slightly more efficient, too.

tower/src/spawn_ready/future.rs
```diff
@@ -49,7 +49,18 @@ where
     fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
         let this = self.project();
 
-        if let Poll::Ready(_) = Pin::new(this.tx.as_mut().expect("illegal state")).poll_closed(cx) {
+        // Is the channel sender closed?
+        // Note that we must actually poll the sender's closed future here,
+        // rather than just calling `is_closed` on it, since we want to be
+        // notified if the receiver is dropped.
+        let closed = {
+            // TODO(eliza): once `tokio` 0.3.2 is released, we can change this back
```
LucioFranco, line 57: open an issue for this?

LucioFranco (approved):
Awesome stuff @hawkw thanks for getting this done!

Issue #475, opened 2020-10-21
hedge middleware has fixed size for timing histogram


hdevalence:
The `Hedge` middleware uses [`hdrhistogram::Histogram::new_with_bounds`](https://docs.rs/hdrhistogram/7.1.0/hdrhistogram/struct.Histogram.html#method.new_with_bounds) with hardcoded values `1` (min, ms), `10_000` (max, ms), and `3` (significant figures). 

When using the `Hedge` middleware in front of a service that takes longer than 10 seconds to process a request, the middleware displays a big error message that it failed to update the histogram.

There seem to be a few potential ways this problem could be solved:

1.  Removing the error message and failing silently;
2.  Allowing user-specified bounds on the histogram (increases API complexity);
3.  Allow automatic resizing of the histogram (either by setting no bound, or by using [`Histogram::auto`](https://docs.rs/hdrhistogram/7.1.0/hdrhistogram/struct.Histogram.html#method.auto)).

It seems like (3) is the best in terms of correctness, and it would be a drop-in change.  The only downside is that resizing the histogram might cause a latency spike. (2) doesn't have this problem, but does make the API more complicated.

teor2345:
Another possible solution:
4. Clamp the values at the 10s mark

Not a great option, but at least it's quick.

hdevalence:
Maybe the cleanest thing to do from a maintainability standpoint would be to enable auto-resizing (3). If someone has latency issues in the future another API could always be added then.

hawkw:
> Maybe the cleanest thing to do from a maintainability standpoint would be to enable auto-resizing (3). If someone has latency issues in the future another API could always be added then.

That seems like the right approach to me.

hawkw:
Fixed by #484.

PR #484, opened 2020-11-18
hedge: use auto-resizing histograms

hdevalence:
The previous code used a fixed-size histogram with an upper bound of 10_000 ms
(10s).  This meant that the `Hedge` middleware would display errors when used
with services that take longer than 10s to complete a response.  Instead, use a
constructor that produces an auto-resizing histogram.  In the future, if the
auto-resizing behavior is an issue, Tower could add a second constructor for
the Hedge middleware that allows specifying bounds, but for now this change is
transparent and avoids spurious errors.

### DIFF
--- a/tower/src/hedge/rotating_histogram.rs
+++ b/tower/src/hedge/rotating_histogram.rs
@@ -19,10 +19,10 @@ pub struct RotatingHistogram {
 impl RotatingHistogram {
     pub fn new(period: Duration) -> RotatingHistogram {
         RotatingHistogram {
-            read: Histogram::<u64>::new_with_bounds(1, 10_000, 3)
-                .expect("Invalid histogram params"),
-            write: Histogram::<u64>::new_with_bounds(1, 10_000, 3)
-                .expect("Invalid histogram params"),
+            // Use an auto-resizing histogram to avoid choosing
+            // a maximum latency bound for all users.
+            read: Histogram::<u64>::new(3).expect("Invalid histogram params"),
+            write: Histogram::<u64>::new(3).expect("Invalid histogram params"),
             last_rotation: Instant::now(),
             period,
         }


hdevalence:
Closes #475

hawkw:
the CI failure is spurious (looks like [a tool could not be downloaded](https://github.com/tower-rs/tower/runs/1421384820)). i'll restart CI.

tower/src/hedge/rotating_histogram.rs
```diff
@@ -19,10 +19,10 @@ pub struct RotatingHistogram {
 impl RotatingHistogram {
     pub fn new(period: Duration) -> RotatingHistogram {
         RotatingHistogram {
-            read: Histogram::<u64>::new_with_bounds(1, 10_000, 3)
-                .expect("Invalid histogram params"),
-            write: Histogram::<u64>::new_with_bounds(1, 10_000, 3)
-                .expect("Invalid histogram params"),
+            // Use an auto-resizing histogram to avoid choosing
+            // a maximum latency bound for all users.
```
hawkw, line 23: maybe worth a note that this means it will resize automatically? not a blocker.

hawkw (approved):
seems good to me!

tower/src/hedge/rotating_histogram.rs
hdevalence: sorry, just to clarify you mean a note that `::new` creates an auto-resizing histogram, or?

hawkw: never mind, i can't read

hawkw:
Ah, hmm, looks like the `cargo-hack` CI job just always fails on PRs from forks? We should fix that. I'll just go ahead and merge this, though.

PR #485, opened 2020-11-23
Add and_then combinator

hlb8122:
**Motivation**
https://docs.rs/futures/0.3.8/futures/future/trait.TryFutureExt.html#method.and_then is a useful method on futures. Perhaps it'd be nice to replicate this for the `ServiceExt` API.

**TODO**
- [x] Documentation
- [ ] Are the generics sane?

### DIFF
--- a/tower/src/util/and_then.rs
+++ b/tower/src/util/and_then.rs
@@ -1,0 +1,81 @@
+use std::marker::PhantomData;
+use std::task::{Context, Poll};
+
+use futures_core::TryFuture;
+use futures_util::{future::AndThen as AndThenFut, TryFutureExt};
+use tower_layer::Layer;
+use tower_service::Service;
+
+/// Service returned by the [`and_then`] combinator.
+///
+/// [`and_then`]: crate::util::ServiceExt::and_then
+#[derive(Clone, Debug)]
+pub struct AndThen<S, F, Fut> {
+    inner: S,
+    f: F,
+    _fut: PhantomData<Fut>,
+}
+
+impl<S, F, Fut> AndThen<S, F, Fut> {
+    /// Creates a new `AndThen` service.
+    pub fn new(inner: S, f: F) -> Self {
+        AndThen {
+            f,
+            inner,
+            _fut: Default::default(),
+        }
+    }
+}
+
+impl<S, F, Request, Fut> Service<Request> for AndThen<S, F, Fut>
+where
+    S: Service<Request>,
+    F: FnOnce(S::Response) -> Fut + Clone,
+    Fut: TryFuture<Error = S::Error>,
+{
+    type Response = Fut::Ok;
+    type Error = Fut::Error;
+    type Future = AndThenFut<S::Future, Fut, F>;
+
+    fn poll_ready(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
+        self.inner.poll_ready(cx)
+    }
+
+    fn call(&mut self, request: Request) -> Self::Future {
+        self.inner.call(request).and_then(self.f.clone())
+    }
+}
+
+/// A [`Layer`] that produces a [`AndThen`] service.
+///
+/// [`Layer`]: tower_layer::Layer
+#[derive(Debug)]
+pub struct AndThenLayer<F, Fut> {
+    f: F,
+    _fut: PhantomData<Fut>,
+}
+
+impl<F, Fut> AndThenLayer<F, Fut> {
+    /// Creates a new [`AndThenLayer`] layer.
+    pub fn new(f: F) -> Self {
+        AndThenLayer {
+            f,
+            _fut: Default::default(),
+        }
+    }
+}
+
+impl<S, F, Fut> Layer<S> for AndThenLayer<F, Fut>
+where
+    F: Clone,
+{
+    type Service = AndThen<S, F, Fut>;
+
+    fn layer(&self, inner: S) -> Self::Service {
+        AndThen {
+            f: self.f.clone(),
+            inner,
+            _fut: Default::default(),
+        }
+    }
+}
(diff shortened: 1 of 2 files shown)


tower/src/util/and_then.rs
```diff
@@ -0,0 +1,81 @@
...
+
+/// Service returned by the [`and_then`] combinator.
+///
+/// [`and_then`]: crate::util::ServiceExt::and_then
+#[derive(Clone, Debug)]
+pub struct AndThen<S, F, Fut> {
+    inner: S,
+    f: F,
+    _fut: PhantomData<Fut>,
+}
```
hawkw, line 17: I don't think it's necessary for the `AndThen` type to be generic over the response future's type. If we remove `Fut` from the type of `AndThen` and `AndThenLayer`, we don't need to use `PhantomData`, and the type of the `AndThen` service is smaller. This may have an impact on compile times when there are large nested service stacks.

I can open a PR against your branch with a version that removes the type parameter, if you like --- I made the change myself before commenting to make sure it does, in fact, compile.

hlb8122: 👍 that'd be awesome. It's only there because I assumed it'd be required from its presence in the futures AndThen. Let's get rid of it.

hawkw: done! https://github.com/hlb8122/tower/pull/1

hlb8122:
@LucioFranco Just in case you missed the discussion here https://github.com/hlb8122/tower/pull/1

tower/src/util/and_then.rs
```diff
@@ -0,0 +1,70 @@
...
+    pub fn new(inner: S, f: F) -> Self {
+        AndThen { f, inner }
+    }
+}
+
+impl<S, F, Request, Fut> Service<Request> for AndThen<S, F>
+where
+    S: Service<Request>,
+    F: FnOnce(S::Response) -> Fut + Clone,
+    Fut: TryFuture<Error = S::Error>,
```
hawkw, line 28: It's a bit of a bummer that this can't change the error type. I think it might be better to use a `From` bound, and convert the error type returned by `poll_ready`, the way I did in my PR adding a `then` combinator (which is essentially a synchronous version of this), https://github.com/tower-rs/tower/pull/500.

hawkw (commented):
Overall, this looks good! I had one suggestion around error types.

Also, it would be nice to add a `ServiceBuilder` method for pushing an `and_then` layer, similar to the ones for other combinators.

tower/src/util/and_then.rs
hlb8122: @hawkw Adding the `From` bound here is a little more verbose than in `then` because [AndThen](https://docs.rs/futures/0.3.8/futures/future/trait.TryFutureExt.html#method.and_then) constrains the return `Error` type.

We could do this by changing `call` to `self.inner.call(request).and_then(self.f.clone()).map_err(Into::into)` and, as a result, `type Future`. Should I do that?

hawkw: That seems correct to me.

hlb8122: Given it a shot.

tower/src/util/and_then.rs
```diff
@@ -0,0 +1,74 @@
...
+impl<S, F, Request, Error, Fut> Service<Request> for AndThen<S, F>
+where
+    S: Service<Request>,
+    S::Error: Into<Error>,
+    F: FnOnce(S::Response) -> Fut + Clone,
+    Fut: TryFuture<Error = Error>,
+{
+    type Response = Fut::Ok;
+    type Error = Error;
+    type Future = AndThenFut<ErrIntoFut<S::Future, Error>, Fut, F>;
```
hawkw, line 36: let's change this to a named type re-exported from `util::future`, the way the other combinators do. I added a macro to make this easier --- if you rebase onto master, you can follow the example of PR #509

hawkw, line 36: I think some of the generic parameters aren't necessary here:
```suggestion
impl<S, F, Request, Fut> Service<Request> for AndThen<S, F>
where
    S: Service<Request>,
    S::Error: Into<Fut::Error>,
    F: FnOnce(S::Response) -> Fut + Clone,
    Fut: TryFuture,
{
    type Response = Fut::Ok;
    type Error = Fut::Error;
    type Future = AndThenFut<ErrIntoFut<S::Future, Fut::Error>, Fut, F>;
```

hawkw (commented):
overall, this looks good. let's change the response future type to a newtype. 

i also had a small style suggestion, but you can take it or leave it.

tower/src/util/and_then.rs
hlb8122: I will get an unnecessary parameter passed you eventually...

hlb8122: This doesn't seem possible
```rust
opaque_future! {
    /// Response future from [`AndThen`] services.
    ///
    /// [`AndThen`]: crate::util::AndThen
    pub type AndThenFuture<F1, F2: TryFuture, N> = future::AndThen<future::ErrInto<F1, F2::Error>, F2, N>;
}
```

Is this acceptable?

```rust
opaque_future! {
    /// Response future from [`AndThen`] services.
    ///
    /// [`AndThen`]: crate::util::AndThen
    pub type AndThenFuture<F1, F2, F2Error, N> = future::AndThen<future::ErrInto<F1, F2Error>, F2, N>;
}
```

hawkw: Either is fine --- you're right that the macro doesn't support trait bounds currently. We could try to hack that in, but I think it would be fine to manually write our own newtype.

Remember that if you do that, you should write `pub struct AndThenFuture<...>(...)` and write a `Future` impl for it, rather than `pub type AndThenFuture...`. This is what the macro would expand to.

hlb8122: How's that look?

hawkw (approved):
looks good to me, thank you!

PR #487, opened 2020-12-11
layer: fix+improve doc links

GeorgeHahn:
- Fixes a link to the `Service` trait on the `tower_layer` module page. ([current page](https://docs.rs/tower-layer/0.3.0/tower_layer/index.html))
- Adds a link to `Service` on the `Layer` trait page. ([current page](https://docs.rs/tower-layer/0.3.0/tower_layer/trait.Layer.html))

These changes use old style links - intra-rustdoc links won't work here because `tower`/`tower_service` aren't referenced by this crate.

### DIFF
--- a/tower-layer/src/lib.rs
+++ b/tower-layer/src/lib.rs
@@ -12,6 +12,8 @@
 //! allows other services to be composed with the service that implements layer.
 //!
 //! A middleware implements the [`Layer`] and [`Service`] trait.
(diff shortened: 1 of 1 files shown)


hawkw (approved):
This looks good to me, thanks! If we update from the latest master, we should get a successful CI build and then I'll be happy to merge this.

Issue #455, opened 2020-05-26
propogate panics through `Buffer`


yaahc:
## Problem

In zebra we're seeing a panic during one of our services `poll_ready` which kills the thread its running in but not the entire application.

https://github.com/ZcashFoundation/zebra/issues/394

The issue seems to be caused by `Buffer` not propagating the panic.

## Proposed Solution

* change `Buffer::new` to store the `JoinHandle` rather than discarding it
* change `poll_ready` on `Buffer` to propogate errors from the join handle in the case where the tx to the spawned future fails, rather than creating a new error.

yaahc:
also, I plan on submitting a PR to implement this feature and test it against our code to verify that it does in fact solve our bug and cause the panics to get propagated correctly.

yaahc:
Further testing showed that we were actually discarding the synthesized error that represents the background thread being unreachable, so even if Buffer propagated panics this would still have been a bug, but I think ideally I'd still like to see Buffer propagate panics.

saleemrashid:
This ties into the error handling/API design section of my proposal in #450

LucioFranco:
Seems like a good idea! Would tie is pretty directly to tokio but I think that is totally fair right now :)

LucioFranco:
@yaahc did we ever get this fixed and can we close this? Or should we keep it open and try to get some fix in `0.4`?

yaahc:
I don't think this ever got fixed. We ended up switching to `panic=abort` in zebra so it was less important for us to fix this, probably still a good idea for `0.4` but not a breaking change afaik.

yaahc:
Okay, this is probably needed actually, we ran into an issue with `panic = abort` that will force us to disable it until I can add a new feature in the compiler and have it make it to stable...

I'll start working on this right away

davidpdrsn:
@yaahc I would like to look into this but I can't quite reproduce it. Probably just something I'm misunderstanding.

In this example the panic in `poll_ready` does get propagated back to main:

```rust
use std::future::Future;
use std::{
    pin::Pin,
    task::{Context, Poll},
};
use tower::{buffer::Buffer, BoxError};
use tower::{Service, ServiceExt};

#[tokio::main]
async fn main() {
    let mut service = Buffer::new(MyService, 100);

    let res = service.ready().await.unwrap().call("bar").await.unwrap();

    eprintln!("panic not propagated")
}

struct MyService;

impl Service<&'static str> for MyService {
    type Response = &'static str;
[...]

PR #488, opened 2020-12-11
start working on panic propagation

yaahc:
This is an extremely rough sketch ATM but I'm starting to work on the panic propagation issue https://github.com/tower-rs/tower/issues/455.

Right now I'm pretty unhappy with the initial approach and am re-evaluating. Also, I ran into a lot of compiler errors just working with the `master` copy of tower and I'm unsure whether that was my fault or if the same problem would occur in CI, so I decided I'd open the PR now so I could get feedback and see how the CI goes.

The things that are currently frustrating:

* cannot clone a `JoinHandle`
* cannot propagate panics when worker is spawned via `pair`

Part of me wonders if the worker itself should be catching panics and propagating it through its own `Handle` type.

### DIFF
--- a/tower/Cargo.toml
+++ b/tower/Cargo.toml
@@ -24,7 +24,7 @@ keywords = ["io", "async", "non-blocking", "futures", "service"]
 edition = "2018"
 
 [features]
-default = ["log", "buffer"]
+default = ["log"]
 log = ["tracing/log"]
 balance = ["discover", "load", "ready-cache", "make", "rand", "slab", "tokio/stream"]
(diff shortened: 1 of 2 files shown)


tower/Cargo.toml
```diff
@@ -24,7 +24,7 @@ keywords = ["io", "async", "non-blocking", "futures", "service"]
 edition = "2018"
 
 [features]
-default = ["log"]
+default = ["log", "buffer"]
```
yaahc, line 27:
```suggestion
default = ["log"]
```

yaahc:
I doubt I'm going to continue working on this any time soon, so closing for now.

Issue #267, opened 2019-04-23
util: add LayerFn which implements `Layer` for a closure


seanmonstar:
Many layers are *very* simple, and it'd be useful to have a way to create one from a closure.

### Example implementation

```rust
#[derive(Clone, Copy, Debug)]
pub struct LayerFn<F>(F);

impl<F, S, Out> Layer<S> for LayerFn<F>
where
    F: Fn(S) -> Out,
{
    type Service = Out;

    fn layer(&self, inner: S) -> Self::Service {
        (self.0)(inner)
    }
}
```

Need to decide on an exposed name:

- `tower::layer_fn()`
- `tower::layer::from_fn()` (similar to `std::iter::from_fn`)
- `tower::layer::mk()` (take from finagle, less clear)

carllerche:
`layer_fn` to mirror `service_fn` 👍

seanmonstar:
I'm kinda partial to following `std::iter::from_fn`...

carllerche:
Should `service_fn` be renamed then?

PR #491, opened 2020-12-25
Add `layer_fn`

davidpdrsn:
Resolves https://github.com/tower-rs/tower/issues/267

I went with `layer_fn` over `layer::from_fn` because changing `service_fn` would be a breaking change. But I don't mind changing it if you think thats more appropriate 😊

### DIFF
--- a/tower-layer/src/layer_fn.rs
+++ b/tower-layer/src/layer_fn.rs
@@ -1,21 +1,23 @@
 use super::Layer;
 
-/// Returns a new `LayerFn` with the given closure.
-pub fn layer_fn<T>(f: T) -> LayerFn<T> {
+/// Returns a new `impl Layer` with the given closure.
+pub fn layer_fn<F, S, T>(f: F) -> impl Layer<S, Service = T> + Clone
+where
+    F: Fn(S) -> T + Clone,
+{
     LayerFn { f }
 }
 
-/// A `Layer` implemented by a closure.
 #[derive(Clone, Copy, Debug)]
-pub struct LayerFn<F> {
+struct LayerFn<F> {
     f: F,
 }
 
-impl<F, S, Out> Layer<S> for LayerFn<F>
+impl<F, S, T> Layer<S> for LayerFn<F>
 where
-    F: Fn(S) -> Out,
+    F: Fn(S) -> T,
 {
-    type Service = Out;
+    type Service = T;
 
     fn layer(&self, inner: S) -> Self::Service {
         (self.f)(inner)
--- a/tower-layer/src/layer_fn.rs
+++ b/tower-layer/src/layer_fn.rs
@@ -2,11 +2,63 @@ use super::Layer;
 use std::fmt;
 
 /// Returns a new `LayerFn` with the given closure.
+///
+/// # Example
+/// ```rust
+/// # use tower::Service;
+/// # use std::task::{Poll, Context};
+/// # use tower_layer::{Layer, layer_fn};
+/// # use std::fmt;
+/// # use std::convert::Infallible;
+/// #
+/// // A middleware that logs requests before forwarding them to another service
+/// pub struct LogService<S> {
+///     target: &'static str,
+///     service: S,
+/// }
+///
+/// impl<S, Request> Service<Request> for LogService<S>
+/// where
+///     S: Service<Request>,
+///     Request: fmt::Debug,
+/// {
+///     type Response = S::Response;
+///     type Error = S::Error;
+///     type Future = S::Future;
+///
+///     fn poll_ready(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
+///         self.service.poll_ready(cx)
+///     }
+///
+///     fn call(&mut self, request: Request) -> Self::Future {
+///         // Log the request
+///         println!("request = {:?}, target = {:?}", request, self.target);
+///
+///         self.service.call(request)
+///     }
+/// }
+///
+/// // A `Layer` that wraps services in `LogService`
+/// let log_layer = layer_fn(|service| {
+///     LogService {
+///         service,
+///         target: "tower-docs",
+///     }
+/// });
+///
+/// // An example service. This one uppercases strings
+/// let uppercase_service = tower::service_fn(|request: String| async move {
+///     Ok::<_, Infallible>(request.to_uppercase())
+/// });
+///
+/// // Wrap our servic in a `LogService` so requests are logged.
+/// let wrapped_service = log_layer.layer(uppercase_service);
+/// ```
 pub fn layer_fn<T>(f: T) -> LayerFn<T> {
     LayerFn { f }
 }
 
-/// A `Layer` implemented by a closure.
+/// A `Layer` implemented by a closure. See the docs for `layer_fn` for more details.
 #[derive(Clone, Copy)]
 pub struct LayerFn<F> {
     f: F,
--- a/tower-layer/src/layer_fn.rs
+++ b/tower-layer/src/layer_fn.rs
@@ -1,7 +1,14 @@
 use super::Layer;
 use std::fmt;
 
-/// Returns a new `LayerFn` with the given closure.
+/// Returns a new [`LayerFn`] that implements [`Layer`] by calling the
+/// given function.
+///
+/// The [`Layer::layer`] method takes a type implementing [`Service`] and
+/// returns a different type implementing [`Service`]. In many cases, this can
+/// be implemented by a function or a closure. The [`LayerFn`] helper allows
+/// writing simple [`Layer`] implementations without needing the boilerplate of
+/// a new struct implementing [`Layer`].
 ///
 /// # Example
 /// ```rust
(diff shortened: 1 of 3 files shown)


tower-layer/src/layer_fn.rs
```diff
@@ -0,0 +1,23 @@
+use super::Layer;
+
+/// Returns a new `LayerFn` with the given closure.
+pub fn layer_fn<T>(f: T) -> LayerFn<T> {
```
olix0r, line 4: Note that there's no need for `LayerFn` to be a public type. This can return an anonymous type like:

```suggestion
pub fn layer_fn<S, T, F>(f: F) -> impl Layer<S, Service = T> + Clone
where
    F: Fn(S) -> T,
    F: Clone,
{
```

davidpdrsn: True but that means users wont be able to name the type in type aliases which I find pretty convenient.

olix0r: The majority of cases can't be named, anyway, since you're usually passing in an `impl Fn(S) -> T`. You can only name the thing when it's a pure `fn(S) -> T`.

I don't think you _have to_ omit the `LayerFn` type; but having used a helper like this extensively in our codebase, we have never actually used the named type in practice.

Just a suggestion, though. Feel free to ignore ;)

hawkw: FWIW, I *think* (though I could be wrong) that it actually wouldn't be a breaking change to replace the `impl Layer` with a named type in the future if there's a need for it — since the `impl Layer` is unnameable, I don't think it would be possible for code to depend on the return type being an `impl Layer` rather than a named type, as long as it still implements `Layer`.

So, if we wanted to, we *could* merge this with `impl Layer` and add a named type if users actually need it. 

On the other hand, @davidpdrsn, if you *know* you have a use-case for being able to name the type, that's demand for it. :)

On the *third* hand, I believe `service_fn` returns a nameable type, so it might be better to be consistent...or change `service_fn` to return `impl Service`.

davidpdrsn: I don't have a specific use case in mind for naming `LayerFn` right now and the use case I have for `ServiceFn` could technically be replaced with `impl Service`:

```rust
pub fn not_found<R>() -> ServiceFn<fn(R) -> Ready<Result<Response<Body>, Infallible>>> {
    // ...
}
```

I could also write

```rust
pub fn not_found<R>() -> impl tower::Service<
    R,
    Response = Response<Body>,
    Error = Infallible,
    Future = Ready<Result<Response<Body>, Infallible>>,
> + Clone {
    // ...
}
```

That is just a bit unwieldy so I went for `ServiceFn` but ultimately it doesn't matter.

So I don't feel strongly about it and I think you're right that changing `impl Layer` to `LayerFn` in the future wont be a breaking change 🤔 I guess if users really need to name the layer type they might as well define their own type and implement `Layer` for it. Don't think thats a big hassle.

I'll change the PR to use `impl Layer` instead.

davidpdrsn: Changed in d86e41e375d65.

olix0r (approved):
lgtm! thanks ;)

tower-layer/src/layer_fn.rs
```diff
@@ -0,0 +1,25 @@
...
+
+/// Returns a new `impl Layer` with the given closure.
+pub fn layer_fn<F, S, T>(f: F) -> impl Layer<S, Service = T> + Clone
+where
+    F: Fn(S) -> T + Clone,
+{
+    LayerFn { f }
+}
+
+#[derive(Clone, Copy, Debug)]
```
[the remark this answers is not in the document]
hawkw: As an aside, I'm not sure if the `Debug` impl is valuable here, even if we weren't returning `impl Layer`. AFAIK, no type implementing `Fn` is `Debug`, and user code can't implement the `Fn` trait for its own types, so I don't think the derived impl will ever work.

It might be better to provide a `Debug` impl that doesn't depend on the function being `Debug`. For example, we could do something like this:

```rust 
impl<F> fmt::Debug for LayerFn<F> {
    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
        f.debug_struct("LayerFn")
            .field("f", &format_args!("<{}>", std::any::type_name::<F>())
            .finish()
    }
}
``` 
(but, this would also require us to return a concrete, named type...)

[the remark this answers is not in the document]
olix0r: Could we return `impl Layer<_> + Clone + fmt::Debug`?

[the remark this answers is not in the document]
olix0r: I'm fine with exposing the named type, but I don't think there's a whole lot of value in preserving `Copy`... `Clone` should be Just Fine imo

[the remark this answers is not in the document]
davidpdrsn: Doesn't sound like there is much gained by using `impl Layer` but potential for loosing out on stuff. Let me know if I should revert to `LayerFn` 😊

If so I'll also add the `Debug` impl like you suggested @hawkw. I always add `#[derive(Debug)]` out of habit.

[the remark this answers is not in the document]
hawkw: I think it's probably a good idea to have a `Debug` that works, or else anything holding this layer also can't easily be `Debug`. I think it should be possible to return `impl Layer<_> + Clone + fmt::Debug`, though, as @olix0r suggested.

I don't care too strongly about whether or not we return a named type from this. I'm just not really sure if `impl Layer<_>` has any actual _benefit_. It doesn't make this code a whole lot more concise, since we still have to have a type implementing `Layer`, it just lets us not make it be `pub`.

tower-layer/src/layer_fn.rs
```diff
@@ -0,0 +1,25 @@
+use super::Layer;
+
+/// Returns a new `impl Layer` with the given closure.
+pub fn layer_fn<F, S, T>(f: F) -> impl Layer<S, Service = T> + Clone
```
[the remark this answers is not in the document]
davidpdrsn: Yes ofc. Added one in bbf93a7

tower-layer/src/layer_fn.rs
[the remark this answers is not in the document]
davidpdrsn: Hehe okay. I've reverted to exposing and returning `LayerFn` and added the debug impl suggested.

tower-layer/src/layer_fn.rs
```diff
@@ -0,0 +1,104 @@
...
+/// });
+///
+/// // Wrap our servic in a `LogService` so requests are logged.
+/// let wrapped_service = log_layer.layer(uppercase_service);
+/// ```
+pub fn layer_fn<T>(f: T) -> LayerFn<T> {
+    LayerFn { f }
+}
+
+/// A `Layer` implemented by a closure. See the docs for `layer_fn` for more details.
```
hawkw, line 61: minor docs nit: we should probably link to the `layer_fn` function here:
```suggestion
/// A `Layer` implemented by a closure. See the docs for [`layer_fn`] for more details.
```

tower-layer/src/layer_fn.rs
```diff
@@ -0,0 +1,104 @@
...
+            .field("f", &format_args!("<{}>", std::any::type_name::<F>()))
+            .finish()
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[allow(dead_code)]
```
hawkw, line 90: why is the `allow(dead_code)` necessary here?

tower-layer/src/layer_fn.rs
```diff
@@ -0,0 +1,104 @@
+use super::Layer;
+use std::fmt;
+
+/// Returns a new `LayerFn` with the given closure.
+///
```
hawkw, line 5: Since this doc comment is the main documentation for `LayerFn` (the `LayerFn` type's docs just refer the reader to here), I think this should probably explain what `layer_fn` is *for*, not just what this function returns.

Maybe:
```suggestion
/// Returns a new [`LayerFn`] that implements [`Layer`] by calling the
/// given function.
///
/// The [`Layer::layer`] method takes a type implementing [`Service`] and
/// returns a different type implementing [`Service`]. In many cases, this can
/// be implemented by a function or a closure. The [`LayerFn`] helper allows
/// writing simple [`Layer`] implementations without needing the boilerplate of
/// a new struct implementing [`Layer`].
///
```
or something like that?

I realize this is basically copied from the docs for `service_fn`, which are basically identical...but I think the `service_fn` docs could use some improvement as well. :)

hawkw (approved):
LGTM overall — I had some docs suggestions.

tower-layer/src/layer_fn.rs
davidpdrsn: Otherwise it complains that `WrappedService.inner` is never read.

hawkw: ah gotcha, thanks!

Issue #253, opened 2019-04-16
MakeConnection: add into_service and as_service


seanmonstar:
This allows having bounds for a `C: MakeConnection<T>`, and then being able to easily convert back into a `Service<T>` so one can use `ServiceExt` methods.

PR #492, opened 2020-12-25
Add `MakeService::into_service` and `MakeService::as_service`

davidpdrsn:
Resolves https://github.com/tower-rs/tower/issues/253

Adds `MakeService::into_service` and `MakeService::as_service` which converts `MakeService`s into `Service`s. `into_service` consumes `self` and `as_service` borrows `self` mutably.

### DIFF
--- a/tower/src/make/make_service.rs
+++ b/tower/src/make/make_service.rs
@@ -1,5 +1,6 @@
 use crate::sealed::Sealed;
 use std::future::Future;
+use std::marker::PhantomData;
 use std::task::{Context, Poll};
 use tower_service::Service;
 
@@ -37,6 +38,45 @@ pub trait MakeService<Target, Request>: Sealed<(Target, Request)> {
 
     /// Create and return a new service value asynchronously.
     fn make_service(&mut self, target: Target) -> Self::Future;
+
+    /// Consume this `MakeService` and convert it into a `Service`.
+    ///
+    /// # Example
+    /// ```
+    /// use std::convert::Infallible;
+    /// use tower::Service;
+    /// use tower::make::MakeService;
+    /// use tower::service_fn;
+    ///
+    /// # fn main() {
+    /// # async {
+    /// // A `MakeService`
+    /// let make_service = service_fn(|make_req: ()| async {
+    ///     Ok::<_, Infallible>(service_fn(|req: String| async {
+    ///         Ok::<_, Infallible>(req)
+    ///     }))
+    /// });
+    ///
+    /// // Convert the `MakeService` into a `Service`
+    /// let mut svc = make_service.into_service();
+    ///
+    /// // Make a new service
+    /// let mut new_svc = svc.call(()).await.unwrap();
+    ///
+    /// // Call the service
+    /// let res = new_svc.call("foo".to_string()).await.unwrap();
+    /// # };
+    /// # }
+    /// ```
+    fn into_service(self) -> IntoService<Self, Request>
+    where
+        Self: Sized,
+    {
+        IntoService {
+            make: self,
+            _marker: PhantomData,
+        }
+    }
 }
 
 impl<M, S, Target, Request> Sealed<(Target, Request)> for M
@@ -63,5 +103,29 @@ where
 
     fn make_service(&mut self, target: Target) -> Self::Future {
         Service::call(self, target)
+    }
+}
+
+#[derive(Debug, Clone)]
+pub struct IntoService<M, Request> {
+    make: M,
+    _marker: PhantomData<Request>,
+}
+
+impl<M, S, Target, Request> Service<Target> for IntoService<M, Request>
+where
+    M: Service<Target, Response = S>,
+    S: Service<Request>,
+{
+    type Response = M::Response;
+    type Error = M::Error;
+    type Future = M::Future;
+
+    fn poll_ready(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
+        self.make.poll_ready(cx)
+    }
+
+    fn call(&mut self, target: Target) -> Self::Future {
+        self.make.make_service(target)
     }
 }
--- a/tower/src/make/make_service.rs
+++ b/tower/src/make/make_service.rs
@@ -77,6 +77,48 @@ pub trait MakeService<Target, Request>: Sealed<(Target, Request)> {
             _marker: PhantomData,
         }
     }
+
+    /// Convert this `MakeService` into a `Service` without consuming the original `MakeService`.
+    ///
+    /// # Example
+    /// ```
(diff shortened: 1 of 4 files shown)


davidpdrsn:
One thought I had was that it would be nice if `IntoService` and `AsService` were public types so users can name them in type aliases and such. Not sure where to expose them though. `tower::make::{IntoService, AsService}` doesn't feel right and if we go with `tower::make::make_service::{IntoService, AsService}` it feels like `MakeService` should be moved in there as well.

tower/src/make/make_service.rs
```diff
@@ -65,3 +147,51 @@ where
         Service::call(self, target)
     }
 }
+
+#[derive(Debug, Clone)]
```
hawkw, line 151: Unfortunately, I think we should probably provide manual `Clone` and `Debug` impls for this, rather than deriving them. The derived implementations require _all_ type parameters of the type deriving `Clone` or `Debug` to implement those traits. However, we don't actually need the request type to implement `Clone` to clone an `IntoService` or implement `Debug` to format one — it's only necessary because of the `PhantomData`.

Therefore, I think we should provide an `impl<M: Clone, Request> Clone for IntoService<M, Request>` and an `impl<M: fmt::Debug, Request> fmt::Debug for IntoService<M, Request>`, rather than deriving them.

tower/src/make/make_service.rs
```diff
@@ -65,3 +147,51 @@ where
...
+    fn poll_ready(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
+        self.make.poll_ready(cx)
+    }
+
+    fn call(&mut self, target: Target) -> Self::Future {
+        self.make.make_service(target)
+    }
+}
+
+#[derive(Debug)]
```
hawkw, line 175: similarly, this should probably not be derived IMO.

tower/src/make/make_service.rs
```diff
@@ -65,3 +147,51 @@ where
         Service::call(self, target)
     }
 }
+
+#[derive(Debug, Clone)]
+pub struct IntoService<M, Request> {
```
hawkw, line 152: let's make sure to add doc comments for these structs as well. it's fine if they just say 
```suggestion
/// Service returned by [`MakeService::into_service`][into]. 
/// 
/// See the documentation on [`into_service`][into] for details.
///
/// [into]: MakeService::into_service
#[derive(Debug, Clone)]
pub struct IntoService<M, Request> {
```
or something

tower/src/make/make_service.rs
```diff
@@ -65,3 +147,51 @@ where
...
+        self.make.poll_ready(cx)
+    }
+
+    fn call(&mut self, target: Target) -> Self::Future {
+        self.make.make_service(target)
+    }
+}
+
+#[derive(Debug)]
+pub struct AsService<'a, M, Request> {
```
hawkw, line 176: similarly, a quick doc comment would be good
```suggestion
/// Service returned by [`MakeService::as_service`][as]. 
/// 
/// See the documentation on [`as_service`][as] for details.
///
/// [as]: MakeService::as_service
#[derive(Debug)]
pub struct AsService<'a, M, Request> {
```

hawkw (changes requested):

tower/src/make/make_service.rs
davidpdrsn: Ah yes ofc. Fixed in 4b1dd1505be29

tower/src/make/make_service.rs
davidpdrsn: 4b1dd1505be29 👍

davidpdrsn:
Should we expose them as `MakeServiceIntoService` and `MakeServiceAsService`? Both a bit wordy 🤔

hawkw:
> Should we expose them as `MakeServiceIntoService` and `MakeServiceAsService`? Both a bit wordy thinking

as long as they  live in the `make` module and aren't re-exported at the crate root, I think  `make::IntoService` and `make::AsService` are better.

tower/src/make/make_service.rs
```diff
@@ -65,3 +148,93 @@ where
...
+    type Error = M::Error;
+    type Future = M::Future;
+
+    fn poll_ready(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
+        self.make.poll_ready(cx)
+    }
+
+    fn call(&mut self, target: Target) -> Self::Future {
+        self.make.make_service(target)
+    }
```
hawkw, line 200: TIOLI: these could probably be inlined...compiler might be smart enough to figure that out now though...

hawkw (approved):
great, this looks good to me! thanks!

davidpdrsn:
> > Should we expose them as `MakeServiceIntoService` and `MakeServiceAsService`? Both a bit wordy thinking
> 
> as long as they live in the `make` module and aren't re-exported at the crate root, I think `make::IntoService` and `make::AsService` are better.

I was thinking if we wanted to make the types public since the `make` module is currently private.

tower/src/make/make_service.rs
davidpdrsn: 6a753cc4f3383 👍

hawkw:
> I was thinking if we wanted to make the types public since the `make` module is currently private.

Hmm...let's just make the `make` module public, then. We should continue re-exporting `MakeService` from the root of the crate, though, and maybe add `#[doc(inline)]` on the re-export.

davidpdrsn:
👌

tower/src/make/mod.rs
```diff
@@ -1,7 +1,9 @@
 //! Trait aliases for Services that produce specific types of Responses.
 
 mod make_connection;
-mod make_service;
+
+pub mod make_service;
 
 pub use self::make_connection::MakeConnection;
+#[doc(inline)]
 pub use self::make_service::MakeService;
```
hawkw, line 9: Sorry, I misunderstood what you were saying earlier. This is the `tower::make` module, which is public, and contains a private module named `make_service`. I thought you were saying the `make` module itself was not public.

Let's not make `tower::make::make_service` pub. Let's just re-export `AsService` and `IntoService` _in_ the `tower::make` module:
```suggestion
mod make_service;

pub use self::make_connection::MakeConnection;
pub use self::make_service::{MakeService, AsService, IntoService};
```

I think this is fine --- it's clear that these are `MakeService` adapters because they're in the `tower::make` module.

hawkw (changes requested):

tower/src/make/mod.rs
davidpdrsn: Haha all good

hawkw (approved):
looks good to me!

PR #493, opened 2020-12-26
buffer: Add Clone + Copy impls for BufferLayer

olix0r:
It's generally the case that layers need to be shareable. There's no
reason that `BufferLayer` should not implement both `Clone` and `Copy`.

This change adds manual `Clone` and `Copy` implementations for
`BufferLayer`.


hawkw (approved):
looks good to me! thanks!

PR #494, opened 2020-12-26
layer: Add a layer_fn helper

olix0r:
It's frequently unnecessary to create concrete types for all layer
implementations.

This change adds a `layer_fn` helper, like `service_fn`, that supports
creating layer implementations from a simple `Fn`.

This is adapted from a utility that we use heavily in the `linkerd2-proxy`
repo.


olix0r:
Dupe of #491

Issue #264, opened 2019-04-23
util: a type that implements `Service<T>` for a `Future<Item=impl Service<T>>`


seanmonstar:
A user may wish to materialize a `Service<T>` immediately, even though they only have a `Future` that resolves to a `Service<T>` eventually. As an example name, it's similar to finagle's `Service.pending`.

### Example

```rust
// MakeService<&str, Request>
let mk_svc = some_http_client();
let dst = "wat";
// Service<Request>
let svc = pending(mk_svc, dst);

svc.oneshot(Request::default());
```

Example implementation: https://github.com/linkerd/linkerd2-proxy/pull/236/files#diff-a38c43fc40e2d0e1164fe070ba8ebac4

davidpdrsn:
I gave this a bit of thought but not quite sure I understand the example.

Is it something like this you're thinking of

```rust
pub fn pending<F, S, E, R>(make_service: F) -> Pending<F, R>
where
    F: Future<Output = Result<S, E>>,
    S: Service<R, Error = E>,
{
    Pending {
        make_service,
        _marker: PhantomData,
    }
}

#[pin_project]
#[derive(Debug, Copy, Clone)]
pub struct Pending<F, R> {
    #[pin]
    make_service: F,
    _marker: PhantomData<R>,
}

impl<F, S, E, R> Future for Pending<F, R>
where
    F: Future<Output = Result<S, E>>,
    S: Service<R, Error = E>,
{
    type Output = Result<S, E>;

    fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
        self.project().make_service.poll(cx)
    }
}
```

where the usage would be

```rust
let pending_svc = pending(async {
    Ok::<_, Infallible>(service_fn(|_req: ()| async {
        Ok::<_, Infallible>(String::from("foo"))
    }))
});
let mut svc = pending_svc.await.unwrap();
let res = svc.call(()).await.unwrap();
```

Or should `Pending` implement `MakeService` and be used more like this:

```rust
let mut pending_svc = service_fn(|_req: ()| async {
    Ok::<_, Infallible>(service_fn(|_req: ()| async {
        Ok::<_, Infallible>(String::from("foo"))
    }))
});
let mut svc = pending_svc.make_service(()).await.unwrap();
let res = svc.call(()).await.unwrap();
```

olix0r:
[...]

PR #495, opened 2020-12-27
Add `Pending` which impls `Service` for `Future<Output = impl Service>`

davidpdrsn:
Resolves https://github.com/tower-rs/tower/issues/264#issuecomment-751400523

Based on https://github.com/linkerd/linkerd2-proxy/blob/1be301f2b0605367be733b4cd96de88108b8889b/linkerd/stack/src/future_service.rs

I thought `Pending` was a better name than `FutureService` but thats up for bike shedding ofc.

Also thought it would be nice to avoid the `Unpin` bound but not sure there is a way around that without always using `Box::pin` or using unsafe.

### DIFF
--- a/tower/src/util/future_service.rs
+++ b/tower/src/util/future_service.rs
@@ -6,43 +6,43 @@ use std::{
 };
 use tower_service::Service;
 
-/// Returns a new `Pending` for the given future.
+/// Returns a new `FutureService` for the given future.
 ///
 /// # Example
 /// ```
 /// use tower::{service_fn, Service, ServiceExt};
-/// use tower::util::pending;
+/// use tower::util::future_service;
 /// use std::convert::Infallible;
 ///
 /// # fn main() {
 /// # async {
-/// let mut pending_svc = pending(Box::pin(async {
+/// let mut future_of_a_service = future_service(Box::pin(async {
 ///     let svc = service_fn(|_req: ()| async { Ok::<_, Infallible>("ok") });
 ///     Ok::<_, Infallible>(svc)
 /// }));
 ///
-/// let svc = pending_svc.ready_and().await.unwrap();
+/// let svc = future_of_a_service.ready_and().await.unwrap();
 /// let res = svc.call(()).await.unwrap();
 /// # };
 /// # }
 /// ```
-pub fn pending<F, S, R, E>(future: F) -> Pending<F, S>
+pub fn future_service<F, S, R, E>(future: F) -> FutureService<F, S>
 where
     F: Future<Output = Result<S, E>> + Unpin,
     S: Service<R, Error = E>,
 {
-    Pending {
+    FutureService {
         inner: State::Future(future),
     }
 }
 
 /// A type that implements `Service` for a `Future` that produces a `Service`.
 #[derive(Clone)]
-pub struct Pending<F, S> {
+pub struct FutureService<F, S> {
     inner: State<F, S>,
 }
 
-impl<F, S> fmt::Debug for Pending<F, S>
+impl<F, S> fmt::Debug for FutureService<F, S>
 where
     S: fmt::Debug,
 {
@@ -65,7 +65,7 @@ enum State<F, S> {
     Service(S),
 }
 
-impl<F, S, R, E> Service<R> for Pending<F, S>
+impl<F, S, R, E> Service<R> for FutureService<F, S>
 where
     F: Future<Output = Result<S, E>> + Unpin,
     S: Service<R, Error = E>,
@@ -99,14 +99,14 @@ where
 #[cfg(test)]
 mod tests {
     use super::*;
-    use crate::util::{pending, ServiceExt};
+    use crate::util::{future_service, ServiceExt};
     use crate::Service;
     use futures::future::{ready, Ready};
     use std::convert::Infallible;
 
     #[tokio::test]
     async fn pending_service_debug_impl() {
-        let mut pending_svc = pending(ready(Ok(DebugService)));
+        let mut pending_svc = future_service(ready(Ok(DebugService)));
 
         assert_eq!(format!("{:?}", pending_svc), "Pending { service: Pending }");
 
--- a/tower/src/util/future_service.rs
+++ b/tower/src/util/future_service.rs
@@ -44,7 +44,7 @@ where
     S: Service<R, Error = E>,
 {
     FutureService {
-        inner: State::Future(future),
+        state: State::Future(future),
     }
 }
 
@@ -53,7 +53,7 @@ where
 /// See `future_service` for more details.
 #[derive(Clone)]
 pub struct FutureService<F, S> {
-    inner: State<F, S>,
+    state: State<F, S>,
 }
 
 impl<F, S> fmt::Debug for FutureService<F, S>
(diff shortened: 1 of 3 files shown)


tower/src/util/pending.rs
```diff
@@ -0,0 +1,141 @@
...
+
+#[derive(Clone)]
+enum Inner<F, S> {
+    Future(F),
+    Service(S),
+}
+
+impl<F, S, R, E> Service<R> for Pending<F, S>
+where
+    F: Future<Output = Result<S, E>> + Unpin,
```
tesaguri, line 70: The `Unpin` can simply be removed since `poll_ready` is not `Pin`ning `self`.

tower/src/util/pending.rs
```diff
@@ -0,0 +1,141 @@
...
+    fn poll_ready(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
+        loop {
+            self.inner = match &mut self.inner {
+                Inner::Future(fut) => {
+                    let fut = Pin::new(fut);
+                    let svc = match fut.poll(cx) {
+                        Poll::Ready(Ok(svc)) => svc,
+                        Poll::Ready(Err(e)) => return Poll::Ready(Err(e)),
+                        Poll::Pending => return Poll::Pending,
+                    };
```
tesaguri, line 86:
```suggestion
                    let svc = futures_core::ready!(fut.poll(cx)?);
```

tesaguri:
The name `Pending` is already used by [`std::future::Pending`][pending], a future which never resolves, and I think using that name for a service which do resolve is confusing.

[pending]: https://doc.rust-lang.org/1.48.0/std/future/struct.Pending.html

tower/src/util/pending.rs
```diff
@@ -0,0 +1,141 @@
...
+{
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        #[derive(Debug)]
+        struct Pending;
+
+        let mut f = f.debug_struct("Pending");
+        let f = match &self.inner {
+            Inner::Future(_) => f.field("service", &Pending),
+            Inner::Service(svc) => f.field("service", svc),
+        };
```
hawkw, line 57: take it or leave it: I have a slight preference for formatting this a little bit more faithfully to its actual implementation, so that the enum is preserved. for example, when pending, it could look like
```
Pending { inner: Inner::Future(...) }
```
and when complete,
```
Pending { inner: Inner::Service(MyService { whatever }) }
```

tower/src/util/pending.rs
```diff
@@ -0,0 +1,141 @@
+use std::fmt;
+use std::{
+    future::Future,
+    pin::Pin,
+    task::{Context, Poll},
+};
+use tower_service::Service;
+
+/// Returns a new `Pending` for the given future.
+///
```
hawkw, line 10: may be worth a comment explaining that the `Unpin` bound is required, because the future will be polled in `poll_ready`, which does not have a pinned receiver...

tower/src/util/pending.rs
```diff
@@ -0,0 +1,141 @@
...
+        let f = match &self.inner {
+            Inner::Future(_) => f.field("service", &Pending),
+            Inner::Service(svc) => f.field("service", svc),
+        };
+        f.finish()
+    }
+}
+
+#[derive(Clone)]
+enum Inner<F, S> {
```
hawkw, line 63: nit, take it or leave it: might be a little more descriptive to name this `State`? but `Inner` is fine too

hawkw (commented):
Overall, this looks pretty good. Personally, though, I agree with @tesagurl's comment

> The name `Pending` is already used by [`std::future::Pending`](https://doc.rust-lang.org/1.48.0/std/future/struct.Pending.html), a future which never resolves, and I think using that name for a service which do resolve is a bit confusing.

IMO, the name `FutureService` is more descriptive of what this actually does. And, it's worth noting that I think the name `Pending` was suggested back in the `futures` 0.1 days, when it was `Async::NotReady` rather than `Poll::Pending`, so there was less of a name collision.

olix0r:
> The name Pending is already used by std::future::Pending, a future which never resolves, and I think using that name for a service which do resolve is a bit confusing.

For what it's worth, this type was initially called `Pending` in our codebase and I renamed it to `FutureService` because the word _pending_ is, as noted, pretty overloaded. I don't really love `FutureService` either, though. Perhaps `FromFuture`, corresponding to a `ServiceBuilder::from_future`?

tower/src/util/pending.rs
davidpdrsn: 🤯

tower/src/util/pending.rs
davidpdrsn: 🤯

tower/src/util/pending.rs
davidpdrsn: Sure thing. Fixed in 3a55869

tower/src/util/pending.rs
davidpdrsn: Good point. Fixed in eca981b.

davidpdrsn:
Uh yeah I didn't remember `std::future::Pending` was a thing. I've renamed it to `FutureService`. I think `FromFuture` is perhaps a bit too vague.

tower/src/util/pending.rs
davidpdrsn: 👌 Changed in 77fce4df59f16

tower/src/util/future_service.rs
```diff
@@ -0,0 +1,163 @@
...
+/// # Example
+/// ```
+/// use tower::{service_fn, Service, ServiceExt};
+/// use tower::util::future_service;
+/// use std::convert::Infallible;
+///
+/// # fn main() {
+/// # async {
+/// let mut future_of_a_service = future_service(Box::pin(async {
+///     let svc = service_fn(|_req: ()| async { Ok::<_, Infallible>("ok") });
```
hawkw, line 23: take it or leave it: it _might_ be worth showing this actually await something? otherwise, the example code shows an unnecessary future (this could just synchronously return the service). we could even just have an `async fn new_service` or something whose body is hidden. just a thought.

tower/src/util/future_service.rs
```diff
@@ -0,0 +1,163 @@
...
+///     Ok::<_, Infallible>(svc)
+/// }));
+///
+/// let svc = future_of_a_service.ready_and().await.unwrap();
```
hawkw, line 27: also, take it or leave it, but I might structure the example with creating the future and wrapping it with `future_service` on separate lines, so that it's clearer what's going on. something like this:
```suggestion
/// // A future which outputs a type implementing `Service`.
/// let future_of_a_service = async {
///     let svc = service_fn(|_req: ()| async { Ok::<_, Infallible>("ok") });
///     Ok::<_, Infallible>(svc)
/// };
///
/// // Wrap the future with a `FutureService`, allowing it to be used
/// // as a service without awaiting the future's completion:
/// let mut svc = future_service(Box::pin(future_of_a_service));
///
/// // Now, when we wait for the service to become ready, it will
/// // drive the future to completion internally.
/// let svc = svc.ready_and().await.unwrap();
```

tower/src/util/future_service.rs
```diff
@@ -0,0 +1,163 @@
...
+/// ```
+///
+/// # Regarding the `Unpin` bound
+///
+/// The `Unpin` bound on `F` is necessary because the future will be polled in
+/// `Service::poll_ready` which doesn't have a pinned receiver (it takes `&mut self` and not `self:
+/// Pin<&mut Self>`). So we cannot put the future into a `Pin` without requiring `Unpin`.
+///
+/// This will most likely come up if you're calling `future_service` with an async block. In that
+/// case you can use `Box::pin(async { ... })` as shown in the example.
```
hawkw, line 40: :+1: thanks for the comment!

hawkw (approved):
This looks good to me overall! I had some small docs suggestions.

PR #497, opened 2020-12-28
balance: Remove MakeBalance::from_rng

olix0r:
`MakeBalance`'s use of `SmallRng` is problematic: since it clones the
`SmallRng` between `Balance` instances, each instance will have the same
state and produce an identical sequence of values. This probably isn't
_dangerous_, though it is certainly unexpected.

This change removes the `MakeBalance::from_rng` and
`MakeBalanceLayer::from_rng` helpers. The `MakeBalance` service now uses
the default RNG via `Balance::new`. `Balance::new` now creates its
`SmallRng` from the `thread_rng` instead of the default entropy source,
as the default entropy source may use the slower `getrandom`. From the
[`rand` docs][from_entropy]:

> In case the overhead of using getrandom to seed many PRNGs is an
> issue, one may prefer to seed from a local PRNG, e.g.
> from_rng(thread_rng()).unwrap().

Finally, this change updates the balancer to the most recent version of
`rand`, v0.8.0.

This seems like a change we should make before the next release/version
bump of `tower`.

[from_entropy]: https://docs.rs/rand/0.8.0/rand/trait.SeedableRng.html#method.from_entropy

### DIFF
--- a/tower/examples/tower-balance.rs
+++ b/tower/examples/tower-balance.rs
@@ -113,7 +113,8 @@ fn gen_disco() -> impl Discover<
                 let svc = tower::service_fn(move |_| {
                     let start = Instant::now();
 
-                    let maxms = latency.as_millis();
+                    let maxms = u64::from(latency.subsec_nanos() / 1_000 / 1_000)
+                        .saturating_add(latency.as_secs().saturating_mul(1_000));
                     let latency = Duration::from_millis(rand::thread_rng().gen_range(0..maxms));
 
                     async move {
--- a/tower/examples/tower-balance.rs
+++ b/tower/examples/tower-balance.rs
@@ -113,9 +113,18 @@ fn gen_disco() -> impl Discover<
                 let svc = tower::service_fn(move |_| {
                     let start = Instant::now();
 
+<<<<<<< HEAD
                     let maxms = u64::from(latency.subsec_nanos() / 1_000 / 1_000)
                         .saturating_add(latency.as_secs().saturating_mul(1_000));
                     let latency = Duration::from_millis(rand::thread_rng().gen_range(0..maxms));
+||||||| parent of c1418ac... fixup gen_range
+                    let maxms = u64::from(latency.subsec_nanos() / 1_000 / 1_000)
+                        .saturating_add(latency.as_secs().saturating_mul(1_000));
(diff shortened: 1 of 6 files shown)


hawkw:
Looks like [the examples need to be updated](https://github.com/tower-rs/tower/pull/497/checks?check_run_id=1618583321#step:4:192).

tower/examples/tower-balance.rs
```diff
@@ -113,9 +113,18 @@ fn gen_disco() -> impl Discover<
...
+                        .saturating_add(latency.as_secs().saturating_mul(1_000));
+                    let latency = Duration::from_millis(rand::thread_rng().gen_range(0..maxms));
+||||||| parent of c1418ac... fixup gen_range
                     let maxms = u64::from(latency.subsec_nanos() / 1_000 / 1_000)
                         .saturating_add(latency.as_secs().saturating_mul(1_000));
                     let latency = Duration::from_millis(rand::thread_rng().gen_range(0, maxms));
+=======
+                    let maxms = latency.as_millis();
+                    let latency = Duration::from_millis(rand::thread_rng().gen_range(0..maxms));
+>>>>>>> c1418ac... fixup gen_range
```
hawkw, line 127: looks like there's a merge conflict that slipped in here?

hawkw (approved):
This seems good to me once the examples compile again!

It does seem like it could be possible to allow constructing `MakeBalance` instances with an overridden RNG, perhaps by adding a constructor that takes an `FnMut() -> Rng`? That way, the RNG is overridable at the `MakeService` level as well as the `Service` level, but with a way of making a _new_ RNG with its own state. But, we can add that in a follow-up if we decide it's even worth having. In practice, I'm not really sure if there's all that much of a need for this --- maybe if you want to mock out randomness for deterministic testing? I dunno.

olix0r:
@hawkw We still support `Balance::from_rng`, so tests that use an actual balancer are still feasible. I don't think it's really worth supporting this on the Make/Layer at this point (especially if we have to support an additional generic).

hawkw:
> @hawkw We still support `Balance::from_rng`, so tests that use an actual balancer are still feasible. I don't think it's really worth supporting this on the Make/Layer at this point (especially if we have to support an additional generic).

that's fair, this is fine!

PR #499, opened 2020-12-28
util: unify and rename combinators

### LINKED ISSUE #498: combinator service unification
Currently, the `ServiceExt` trait has `with` and `try_with` methods for composing a `Service` with functions that modify the request type, and `map_err` and `map_ok` methods for composing a service with functions that modify the response type. Meanwhile, `ServiceBuilder` has `map_request` for composing a service with a request mapping function, and `map_response` for composing a service with a response-mapping function. These combinators are very similar in purpose, but are implemented using different middleware types that essentially duplicate the same behavior.

Before releasing 0.4, we should probably unify these APIs. In particular, it would be good to de-duplicate the middleware service types, and to unify the naming. 

Personally, I think the `ServiceExt` versions of `map_X` have a slightly better design, since they take `FnOnce + Clone`, rather than taking a `Fn` and `Arc`ing it internally. This avoids the overhead of `Arc` for mapping functions that are trivially clonable (e.g. function pointers)  but still allows the user to use `Arc`s when needed.

I'm open to bikeshedding what naming to settle on. `with` has an analogy with the `futures` crate's `Sink` methods, but I personally find `map_request` and `map_response` a *bit* clearer in the `tower` context. However, if we want to have both `Err` and `Ok` versions of these combinators, which I think we do, `map_response_err` and `map_response_ok` gets a bit wordy...perhaps we could just have `map_response` and `map_err`?

hawkw:
Currently, the `ServiceExt` trait has `with` and `try_with` methods for
composing a `Service` with functions that modify the request type, and
`map_err` and `map_ok` methods for composing a service with functions
that modify the response type. Meanwhile, `ServiceBuilder` has
`map_request` for composing a service with a request mapping function,
and `map_response` for composing a service with a response-mapping
function. These combinators are very similar in purpose, but are
implemented using different middleware types that essentially duplicate
the same behavior.

Before releasing 0.4, we should probably unify these APIs. In
particular, it would be good to de-duplicate the middleware service
types, and to unify the naming.

This commit makes the following changes:
- Rename the `ServiceExt::with` and `ServiceExt::try_with` combinators
  to `map_request` and `try_map_request`
  - Rename the `ServiceExt::map_ok` combinator to `map_response`
- Unify the `ServiceBuilder::map_request` and `ServiceExt::map_request`
  combinators to use the same `Service` type
- Unify the `ServiceBuilder::map_response` and
  `ServiceExt::map_response` combinators to use the same `Service` type
- Unify the `ServiceBuilder::map_err` and `ServiceExt::map_err`
  combinators to use the same `Service` type
- Only take `FnOnce + Clone` when in response/err combinators, which
  require cloning into the future type. `MapRequest` and `TryMapRequest`
  now take `FnMut(Request)`s and don't clone them every time they're
  called
- Reexport future types for combinators where it makes sense.
- Add a `try_map_request` method to `ServiceBuilder`

Closes #498


hawkw:
cc @hlb8122 --- sorry for hacking up your code, but i think this should improve the overall API surface a bit!

tower/src/builder/mod.rs
```diff
@@ -197,11 +197,23 @@ impl<L> ServiceBuilder<L> {
...
     where
-        F: Fn(R1) -> R2,
+        F: FnMut(R1) -> R2 + Clone,
     {
         self.layer(crate::util::MapRequestLayer::new(f))
     }
 
+    /// Fallibly one request type to another, or to an error.
+    #[cfg(feature = "util")]
+    pub fn try_map_request<F, R1, R2, E>(
```
seanmonstar, line 207: The naming here feels like the unfortunate conjunction of two worlds... Usually `map` is infallible, and `and_then` or `then` is fallible, but we don't want this to happen *after* the service (implied by "service then closure"). There isn't anything like `before`... Maybe something like `filter` or `filter_request`?

tower/src/builder/mod.rs
```diff
@@ -197,11 +197,23 @@ impl<L> ServiceBuilder<L> {
         f: F,
     ) -> ServiceBuilder<Stack<crate::util::MapRequestLayer<F>, L>>
     where
-        F: Fn(R1) -> R2,
+        F: FnMut(R1) -> R2 + Clone,
```
seanmonstar, line 200: Is the `Clone` actually required here? I'd think it's only required if you want to be able to clone the `MapRequest<S, F>`.

hawkw: The `Clone` bound is required by the `Layer`, since `Layer::layer` takes `&self`. The same layer can be used to wrap any number of services, so the `MapRequestLayer` has to clone the function each time it creates a service. 

Similarly, the same `ServiceBuilder` _could_ actually be used to wrap any number of root services...although I notice that [the current `ServiceBuilder::service` method takes `self` by value](https://github.com/tower-rs/tower/blob/57e8412ae4a31343e78363288e1cd9de4898b229/tower/src/builder/mod.rs#L237-L238). We should maybe change that for 0.4?

tower/src/builder/mod.rs
hawkw: Hmm, I don't love this naming either. However, I'm not sure about `filter`, since we already have a [`Filter` middleware](https://github.com/tower-rs/tower/blob/master/tower/src/filter/mod.rs) which takes an async predicate...

tower/src/builder/mod.rs
seanmonstar: Oh right, forgot about `Layer` and `&self`. Carry on.

tower/src/builder/mod.rs
seanmonstar: Uh huh... welp... I dunno, has this method been requested (with an example of need)? If not, my not-very-strong-opinion would be to just leave it off (it seems people could use `tower::filter` if needed?).

hawkw: It was added in https://github.com/tower-rs/tower/pull/396 --- @hlb8122, are you actually using `try_with` presently, and is there a motivation for using it instead of the `tower::filter` module?

It would be nice if we could add a `tower::filter::Predicate` impl for synchronous functions as well, but that would conflict with the blanket impl for functions returning `Future`:
https://github.com/tower-rs/tower/blob/704bfa819e3301e7aa0a35b6c09042d0b05670e4/tower/src/filter/predicate.rs#L15-L25

This feels like a case where the `futures` 0.1 `IntoFuture` trait could come in handy, by making it possible to automatically convert a synchronous `Result` into a ready future, so users don't have to wrap it. Then, the blanket impl could take `Fn(Request) -> IntoFuture<...>`...

hawkw: FWIW, I'm fine with removing the `try_map_request`/`try_with` methods and middleware entirely, too. It was just here when I sat down to start cleaning this up, so I erred on the side of keeping it.

hlb8122: I'm not using `try_with`. I remember reading that @hdevalence was using some of the service combinators, maybe he'd like to comment on the usefulness perhaps.

A weakish defense of fallible mapping of requests would be: There might be some  transforms that are easier written as one transform using `try_with`/`try_map_request` rather than `with`/`map_request` then `filter`?

hawkw: That's a good point — a `try_map_request`/`try_with` allows changing the request type _and_ rejecting requests in one place, which the `tower::filter` interface doesn't. 

I guess the concern, though, is just that it doesn't feel great to have a bunch of different ways to do similar things that don't *quite* overlap. I think it would be good to have both synchronous and asynchronous filtering/fallible request mapping, but they should ideally have consistent API surfaces. 

I'm not opposed to the `filter` naming scheme for fallible-request-mapping APIs — it does seem much less wordy than `try_map_request`. maybe it would be good to have both synchronous and asynchronous versions of `filter` and expose them as combinators as well as layers?

LucioFranco (approved):

tower/src/builder/mod.rs
hawkw: I think we should punt on this in *this* PR and address `filter`/`try_with` in a follow-up (see #502).

PR #500, opened 2020-12-28
util: Add `then` combinator

hawkw:
Currently, `ServiceExt` and `ServiceBuilder` provide combinators for
mapping successful responses to other responses, and mapping errors to
other errors, but don't provide a way to map between `Ok` and `Err`
results.

For completeness, this branch adds a new `then` combinator, which takes
a function from `Result` to `Result` and applies it when the service's
future completes. This can be used for recovering from some errors or
for rejecting some `Ok` responses. It can also be used for behaviors
that should be run when a service's future completes regardless of
whether it completed successfully or not.

Depends on #499


olix0r:
This seems somewhat incongruous with [`Futures::then`](https://docs.rs/futures/0.3.8/futures/future/trait.FutureExt.html#method.then), which takes a closure that produces another _future_. Should this behave similarly? This would permit more flexible uses, like calling another future in the `then` block.

hawkw:
> This seems somewhat incongruous with [`Futures::then`](https://docs.rs/futures/0.3.8/futures/future/trait.FutureExt.html#method.then), which takes a closure that produces another _future_. Should this behave similarly? This would permit more flexible uses, like calling another future in the `then` block.
> 
> What we have here seems to be more like `Futures::map`

This is a good point --- I got the naming mixed up. `then` should definitely be async, and a synchronous version of this can be implemented on top of it.

tower/src/util/mod.rs
```diff
@@ -215,6 +220,216 @@ pub trait ServiceExt<Request>: tower_service::Service<Request> {
...
+    /// service's future completes, regardless of whether it succeeded or failed.
+    ///
+    /// This method can be used to change the [`Response`] type of the service
+    /// into a different type. It can also be used to change the [`Error`] type
+    /// of the service. However, because the `map_result` function is not applied
+    /// to the errors returned by the service's [`poll_ready`] method, it must
+    /// be possible to convert the service's [`Error`] type into the error type
+    /// returned by the `map_result` function. This is trivial when the function
+    /// returns the same error type as the service, but in other cases, it can
+    /// be useful to use [`BoxError`] to erase differing error types.
```
LucioFranco, line 248: Is box error a type we have now?

tower/src/util/mod.rs
```diff
@@ -327,6 +542,17 @@ pub trait ServiceExt<Request>: tower_service::Service<Request> {
     {
         TryMapRequest::new(self, f)
     }
+
+    /// TODO(eliza): docs
```
LucioFranco, line 546: Missing docs

tower/src/util/then.rs
```diff
@@ -0,0 +1,77 @@
...
+    /// Creates a new `Then` service.
+    pub fn new(inner: S, f: F) -> Self {
+        Then { f, inner }
+    }
+}
+
+impl<S, F, Request, Response, Error, Fut> Service<Request> for Then<S, F>
+where
+    S: Service<Request>,
+    Error: From<S::Error>,
```
LucioFranco, line 39: I thought we were going with Into?

LucioFranco (approved):
LGTM, minor comments

tower/src/util/mod.rs
hawkw: https://github.com/tower-rs/tower/blob/1a845433177b31ebc6daff765ee81468c42d6676/tower/src/lib.rs#L85-L86

tower/src/util/then.rs
hawkw: oh whoops
```suggestion
    S::Error: Into<Error>,
```

PR #503, opened 2020-12-29
util: Add BoxService layer helpers

olix0r:
This change adds `BoxService::layer` and `UnsyncBoxService::layer`
helpers to provide a convenience helper for boxing services in layered
stacks.

### DIFF
--- a/tower/src/util/boxed/sync.rs
+++ b/tower/src/util/boxed/sync.rs
@@ -42,7 +42,7 @@ impl<T, U, E> BoxService<T, U, E> {
     }
 
     /// Returns a [`Layer`] for wrapping a [`Service`] in a `BoxService` middleware.
-    pub fn layer<S>() -> impl Layer<S, Service = Self> + Clone
+    pub fn layer<S>() -> LayerFn<fn(S) -> Self>
     where
         S: Service<T, Response = U, Error = E> + Send + 'static,
         S::Future: Send + 'static,
--- a/tower/src/util/boxed/sync.rs
+++ b/tower/src/util/boxed/sync.rs
@@ -1,3 +1,4 @@
+use tower_layer::{layer_fn, Layer};
 use tower_service::Service;
 
 use std::fmt;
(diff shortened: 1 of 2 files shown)


tower/src/util/boxed/sync.rs
```diff
@@ -39,6 +40,15 @@ impl<T, U, E> BoxService<T, U, E> {
         let inner = Box::new(Boxed { inner });
         BoxService { inner }
     }
+
+    #[allow(missing_docs)]
```
hawkw, line 44: Can we add a short doc comment for this? Just something like
```suggestion
    /// Returns a [`Layer`] for wrapping a [`Service`] in a `BoxService` middleware.
```

tower/src/util/boxed/sync.rs
```diff
@@ -39,6 +40,15 @@ impl<T, U, E> BoxService<T, U, E> {
...
+    pub fn layer<S>() -> impl Layer<S, Service = Self> + Clone
```
hawkw, line 45: Since `Self::new` is not a closure, this actually can return a named type rather than an `impl Layer`. I think this could just be
```suggestion
    pub fn layer<S>() -> LayerFn<fn(S) -> Self>
```

tower/src/util/boxed/unsync.rs
```diff
@@ -33,6 +34,15 @@ impl<T, U, E> UnsyncBoxService<T, U, E> {
         let inner = Box::new(UnsyncBoxed { inner });
         UnsyncBoxService { inner }
     }
+
+    #[allow(missing_docs)]
```
hawkw, line 38: similarly:
```suggestion
    /// Returns a [`Layer`] for wrapping a [`Service`] in an `UnsyncBoxService` middleware.
```

tower/src/util/boxed/unsync.rs
```diff
@@ -33,6 +34,15 @@ impl<T, U, E> UnsyncBoxService<T, U, E> {
...
+    pub fn layer<S>() -> impl Layer<S, Service = Self> + Clone
```
hawkw, line 39: and again
```suggestion
    pub fn layer<S>() -> LayerFn<fn(S) -> Self>
```

hawkw (approved):
looks good to me, thanks!

PR #504, opened 2021-01-04
Make `ServiceBuilder::service` take `self` by reference

### LINKED ISSUE #501: consider making `ServiceBuilder::service` take `&self`
Currently, the `ServiceBuilder::service` method takes `self` by value. This means that a `ServiceBuilder` with a bunch of layers in it can only be used to wrap one `Service` with its middleware.

However, the `Layer::layer` method only needs an `&self` receiver --- the layers themselves can be reused. All `ServiceBuilder::service` does is _call_ `Layer::layer`, so there's no real reason for `service` to move `self`.

We should reconsider taking `self` here. If we took `&self`, the same middleware could be applied to any number of services, making `ServiceBuilder` a more useful primitive for shared middleware configurations or for wrapping every service that comes from a `MakeService`, or similar.

See:

> Similarly, the same `ServiceBuilder` _could_ actually be used to wrap any number of root services...although I notice that [the current `ServiceBuilder::service` method takes `self` by value](https://github.com/tower-rs/tower/blob/57e8412ae4a31343e78363288e1cd9de4898b229/tower/src/builder/mod.rs#L237-L238). We should maybe change that for 0.4?
>
> _Originally posted by @hawkw in https://github.com/tower-rs/tower/pull/499#discussion_r549514710_

davidpdrsn:
Fixes #501

### DIFF
--- a/tower/CHANGELOG.md
+++ b/tower/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Added `MakeService::into_service` and `MakeService::as_service` for
   converting `MakeService`s into `Service`s.
+- Make `ServiceBuilder::service` take `self` by reference rather than by value.
 
 ### Changed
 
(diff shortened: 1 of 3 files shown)


hawkw (approved):
looks good to me, thank you!

davidpdrsn:
Forgot to update changelog 😊

PR #505, opened 2021-01-04
Fix clippy warnings

davidpdrsn:
Thought I might as well fix the clippy warnings. Feel free to close this if you don't want the git churn.

### DIFF
--- a/tower/src/balance/p2c/service.rs
+++ b/tower/src/balance/p2c/service.rs
@@ -78,7 +78,6 @@ enum Error<E> {
     Canceled,
 }
 
-#[allow(clippy::len_without_is_empty)]
 impl<D, Req> Balance<D, Req>
 where
     D: Discover,
@@ -107,6 +106,11 @@ where
     /// Returns the number of endpoints currently tracked by the balancer.
     pub fn len(&self) -> usize {
         self.services.len()
+    }
+
+    /// Returns whether or not the balancer is empty.
+    pub fn is_empty(&self) -> bool {
+        self.len() == 0
     }
 }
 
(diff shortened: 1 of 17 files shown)


hawkw (approved):

tower/src/balance/p2c/service.rs
```diff
@@ -78,6 +78,7 @@ enum Error<E> {
     Canceled,
 }
 
+#[allow(clippy::len_without_is_empty)]
```
hawkw, line 81: should we just consider adding an `is_empty` method?

hawkw (approved):

hawkw:
One other note: if we're going to start caring about clippy lints (which I think we ought to), I think we should probably add a Clippy check step on CI. This will stop us from going and adding new code that clippy doesn't like, and having to do another big "fix clippy warnings" PR to make them go away. We can do that separately, though.

tower/src/balance/p2c/service.rs
davidpdrsn: Sure. I just wasn't sure if it was appropriate here. Not very familiar with this API. Added `is_empty` in 58821d7826a07

hawkw (approved):

tower/src/balance/p2c/service.rs
```diff
@@ -107,6 +107,11 @@ where
     pub fn len(&self) -> usize {
         self.services.len()
     }
+
+    /// Returns whether or not the balancer is empty.
+    pub fn is_empty(&self) -> bool {
+        self.len() == 0
```
olix0r, line 113: I would expect this to trigger clippy. Should this be:
```suggestion
        self.services.is_empty()
```

olix0r: I'm not sure if ReadyCache actually has an `is_empty`, but it probably should if it does not...

davidpdrsn: Uh good catch. Fixed in 5dc604458ccf7

PR #508, opened 2021-01-06
rewrite `tower::filter`

### LINKED ISSUE #502: decide the future of `tower::filter`
## Motivation

It was pointed out that there is currently some overlap between the `try_with` `Service` combinator and `tower::filter` middleware (see https://github.com/tower-rs/tower/pull/499#discussion_r549522471). `try_with` synchronously maps from a `Request` -> `Result<DifferentRequest, Error>`, while `tower::filter` _asynchronously_ maps from a `&Request` to a `Result<(), Error>`. The key differences are:
 - `try_with` takes a request by value, and allows the predicate to return a *different* request value
 - `try_with` also permits changing the _type_ of the request
 - `try_with` is synchronous, while `tower::filter` is asynchronous
 - `tower::filter` has a `Predicate` trait, which can be implemented by more than just functions. For example, a struct with a `HashSet` could implement `Predicate` by failing requests that match the values in the hashset.
 
It definitely seems like there's demand for both synchronous and asynchronous request filtering. However, the APIs we have currently differ pretty significantly. It would be nice to make them more consistent with each other.

As an aside, `tower::filter` [does not seem all that widely used](https://github.com/search?l=&p=1&q=%22tower%3A%3Afilter%22+extension%3Ars&ref=advsearch&type=Code). 

Meanwhile, `linkerd2-proxy` defines its own `RequestFilter` middleware, using a [predicate trait](https://github.com/linkerd/linkerd2-proxy/blob/24bee8cbc5413b4587a14bea1e2714ce1f1f919a/linkerd/stack/src/request_filter.rs#L8-L12) that's essentially in between `tower::filter` and `ServiceExt::try_with`:
- it's synchronous, like `try_with`
- it allows modifying the type of the request, like `try_with`
- it uses a trait for predicates, rather than a `Fn`, like `tower::filter`
- it uses a similar naming scheme to `tower::filter` ("filtering" rather than "with"/"map")

## Proposal

I think we should try to unify these APIs into something with a more consistent interface. I suggest the following:
- continue using the "Filter"/"Predicate" terminology rather than the function composition terminology ("with"/"map")
- take `Request` types by value rather than by reference, and allow returning a new request type
- continue providing traits that can be implemented by things other than functions
- provide both sync and async filtering APIs

It's theoretically possible to have one API for both sync and async filtering, but this would require a trait like `IntoFuture` that can be implemented by both `Future<Output = Result<T, E>` and `Result<T, E>`. A predicate trait could return something implementing that trait, and `Result` could have an impl that turns itself into a ready future. This would let us have one API for both sync and async predicates. However, it introduces some complexity in the type signatures, and requires a slight added overhead (polling the immediately ready future when the predicate is synchronous). The added complexity of using an `IntoFuture`-like trait might be less of an issue when the standard library stabilizes `IntoFuture`...

On the other hand, we could have separate sync and async filter APIs. The main disadvantage of _that_ approach is that we'd then need to bikeshed a consistent naming scheme that makes both the distinction and the similarity similar. Historically, we've not been great at naming things efficiently ;)

cc @LucioFranco @seanmonstar @olix0r @hlb8122  --- what do you all think?

https://github.com/linkerd/linkerd2-proxy/blob/main/linkerd/stack/src/request_filter.rs

hawkw:
## Motivation

It was pointed out that there is currently some overlap between the
`try_with` `Service` combinator and `tower::filter` middleware (see https://github.com/tower-rs/tower/pull/499#discussion_r549522471 ).
`try_with` synchronously maps from a `Request` ->
`Result<DifferentRequest, Error>`, while `tower::filter`
_asynchronously_ maps from a `&Request` to a `Result<(), Error>`. The
key differences are: - `try_with` takes a request by value, and allows
the predicate to return a *different* request value - `try_with` also
permits changing the _type_ of the request - `try_with` is synchronous,
while `tower::filter` is asynchronous - `tower::filter` has a
`Predicate` trait, which can be implemented by more than just functions.
For example, a struct with a `HashSet` could implement `Predicate` by
failing requests that match the values in the hashset.

It definitely seems like there's demand for both synchronous and
asynchronous request filtering. However, the APIs we have currently
differ pretty significantly. It would be nice to make them more
consistent with each other.

As an aside, `tower::filter` [does not seem all that widely used][1].

Meanwhile, `linkerd2-proxy` defines its own `RequestFilter` middleware,
using a [predicate trait][2] that's essentially in between `tower::filter` and
`ServiceExt::try_with`: - it's synchronous, like `try_with` - it allows
modifying the type of the request, like `try_with` - it uses a trait for
predicates, rather than a `Fn`, like `tower::filter` - it uses a similar
naming scheme to `tower::filter` ("filtering" rather than "with"/"map").

[1]: https://github.com/search?l=&p=1&q=%22tower%3A%3Afilter%22+extension%3Ars&ref=advsearch&type=Code
[2]: https://github.com/linkerd/linkerd2-proxy/blob/24bee8cbc5413b4587a14bea1e2714ce1f1f919a/linkerd/stack/src/request_filter.rs#L8-L12

## Solution

This branch rewrites `tower::filter` to make the following changes:

* Predicates are synchronous by default. A separate `AsyncFilter` type
  and an `AsyncPredicate` trait are available for predicates returning
  futures.
* Predicates may now return a new `Request` type, allowing `Filter` and
  `AsyncFilter` to subsume `try_map_request`.
* Predicates may now return any error type, and errors are now converted
  into `BoxError`s.

Closes #502


tower/src/macros.rs
```diff
@@ -0,0 +1,24 @@
...
+            $actual: std::future::Future,
+        {
+            type Output = <$actual as std::future::Future>::Output;
+            #[inline]
+            fn poll(self: std::pin::Pin<&mut Self>, cx: &mut std::task::Context<'_>) -> std::task::Poll<Self::Output> {
+                self.project().0.poll(cx)
+            }
+        }
+    }
+}
```
hawkw, line 24: The idea behind this thing is that we probably don't want to expose type aliases for `futures_util` futures in public APIs, for stability reasons. Since type aliases don't hide the aliased type, users could refer to these as the nested chain of `futures_util` combinator types. Thus, if we change the specific combinator futures, that's potentially a breaking change.

I'll wrap other exposed combinator futures in `opaque_future!` when this merges.

tower/Cargo.toml
```diff
@@ -29,7 +29,7 @@ log = ["tracing/log"]
 balance = ["discover", "load", "ready-cache", "make", "rand", "slab", "tokio-stream"]
 buffer = ["tokio/sync", "tokio/rt", "tokio-stream"]
 discover = []
-filter = []
+filter = ["futures_util"]
```
hawkw, line 32:
```suggestion
filter = ["futures-util"]
```

tower/src/filter/predicate.rs
```diff
@@ -1,25 +1,69 @@
...
+    /// The type of requests returned by `check`.
+    ///
+    /// This request is forwarded to the inner service if the predicate
+    /// succeeds.
+    type Request;
+    /// The type of errors returned by `check`.
+    ///
+    /// If the predicate fails, this is returned rather than calling the inner
+    /// service.
+    type Error: Into<crate::BoxError>;
```
olix0r, line 17: What's the value of having an `Error` type parameter when they're always coerced to the box error? What's the downside to having the traits simply require that boxed errors are returned?

hawkw: hmm, yeah, we could just remove that, good point! the only downside is that the user has to box the error, which adds a bit of code to the predicate implementation. with this approach, we can automatically box the error.

on the other hand, if the user code returns errors via `?`, they don't have to write `Box::new` etc either...

olix0r: Right. And the `.map_err(Into::into)` may be less onerous than having extra `type Error = MyError` defns?

I don't have a strong opinion on this either way, but I wanted to raise the question to get some consensus on which pattern is better.

tower/src/filter/mod.rs
```diff
@@ -1,55 +1,146 @@
...
+    type Future = ResponseFuture<T::Response, T::Future>;
+
+    fn poll_ready(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
+        self.inner.poll_ready(cx).map(|res| res.map_err(Into::into))
+    }
+
+    fn call(&mut self, request: Request) -> Self::Future {
+        ResponseFuture(match self.predicate.check(request) {
+            Ok(request) => Either::Right(self.inner.call(request).err_into()),
+
```
LucioFranco, line 94: odd newline?

LucioFranco (approved):
LGTM

tower/src/builder/mod.rs
```diff
@@ -239,6 +239,42 @@ impl<L> ServiceBuilder<L> {
         self.layer(crate::timeout::TimeoutLayer::new(timeout))
     }
 
+    /// Conditionally reject requests based on `predicate`.
+    ///
+    /// `predicate` must implement the [`Predicate`] trait.
```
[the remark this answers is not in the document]
hawkw: i don't think it makes sense to use a bulleted list with only one entry?

PR #510, opened 2021-01-07
service: Improve example in docs

davidpdrsn:
- Fixed a couple of typos that I happen to discover.
- Makes the `Timeout` example from the `Service` docs runnable and adds some comments to the code.

### DIFF
--- a/tower-service/src/lib.rs
+++ b/tower-service/src/lib.rs
@@ -163,7 +163,7 @@ use std::task::{Context, Poll};
 ///     // Errors may be either `Expired` if the timeout expired, or the inner service's
 ///     // `Error` type. Therefore, we return a boxed `dyn Error` trait object to erase
 ///     // the error's type.
-///     type Error = Box<dyn Error>;
+///     type Error = Box<dyn Error + Send + Sync>;
 ///     type Future = Pin<Box<dyn Future<Output = Result<Self::Response, Self::Error>>>>;
 ///
 ///     fn poll_ready(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
--- a/tower-service/src/lib.rs
+++ b/tower-service/src/lib.rs
@@ -47,7 +47,6 @@ use std::task::{Context, Poll};
 /// # use std::task::{Poll, Context};
 /// # use std::future::Future;
 /// # use tower_service::Service;
-///
 /// use http::{Request, Response, StatusCode};
 ///
 /// struct HelloWorld;
@@ -71,7 +70,7 @@ use std::task::{Context, Poll};
 ///             .status(StatusCode::OK)
 ///             .body(body)
 ///             .expect("Unable to create `http::Response`");
-///         
+///
 ///         // create a response in a future.
 ///         let fut = async {
 ///             Ok(resp)
@@ -112,7 +111,7 @@ use std::task::{Context, Poll};
 ///
 /// Take timeouts as an example:
 ///
-/// ```rust,ignore
+/// ```rust
 /// use tower_service::Service;
 /// use tower_layer::Layer;
 /// use futures::FutureExt;
@@ -120,16 +119,14 @@ use std::task::{Context, Poll};
 /// use std::task::{Context, Poll};
 /// use std::time::Duration;
 /// use std::pin::Pin;
+/// use std::fmt;
+/// use std::error::Error;
 ///
-///
+/// // our timeout service which wraps another service
 /// pub struct Timeout<T> {
 ///     inner: T,
 ///     timeout: Duration,
 /// }
-///
-/// pub struct TimeoutLayer(Duration);
-///
-/// pub struct Expired;
 ///
 /// impl<T> Timeout<T> {
 ///     pub fn new(inner: T, timeout: Duration) -> Timeout<T> {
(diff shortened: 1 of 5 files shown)


tower-service/src/lib.rs
```diff
@@ -139,48 +137,84 @@ use std::task::{Context, Poll};
...
+/// impl Error for Expired {}
+///
+/// // We can implement `Service` for `Timeout<T>` if `T` is a `Service`
 /// impl<T, Request> Service<Request> for Timeout<T>
 /// where
 ///     T: Service<Request>,
 ///     T::Future: 'static,
-///     T::Error: From<Expired> + 'static,
-///     T::Response: 'static
+///     T::Error: Into<Box<dyn Error>> + 'static,
```
LucioFranco, line 157:
```suggestion
///     T::Error: Into<Box<dyn Error + Send + Sync>> + 'static,
```

Because trait obj are a PITA

tower-service/src/lib.rs
```diff
@@ -139,48 +137,84 @@ use std::task::{Context, Poll};
...
+///     T::Response: 'static,
 /// {
+///     // `Timeout` doesn't modify the response type, so we use `T`'s response type
 ///     type Response = T::Response;
-///     type Error = T::Error;
+///     // Errors may be either `Expired` if the timeout expired, or the inner service's
+///     // `Error` type. Therefore, we return a boxed `dyn Error` trait object to erase
+///     // the error's type.
+///     type Error = Box<dyn Error>;
```
LucioFranco, line 165:
```suggestion
///     type Error = Box<dyn Error + Send + Sync>;
```

tower-service/src/lib.rs
```diff
@@ -139,48 +137,84 @@ use std::task::{Context, Poll};
...
-///     T::Error: From<Expired> + 'static,
-///     T::Response: 'static
+///     T::Error: Into<Box<dyn Error>> + 'static,
+///     T::Response: 'static,
 /// {
+///     // `Timeout` doesn't modify the response type, so we use `T`'s response type
 ///     type Response = T::Response;
-///     type Error = T::Error;
+///     // Errors may be either `Expired` if the timeout expired, or the inner service's
+///     // `Error` type. Therefore, we return a boxed `dyn Error` trait object to erase
```
LucioFranco, line 163:
```suggestion
///     // `Error` type. Therefore, we return a boxed `dyn Error + Send + Sync` trait object to erase
```

tower-service/src/lib.rs
```diff
@@ -139,48 +137,84 @@ use std::task::{Context, Poll};
...
+///         // and an error will be returned and `fut` will be dropped and not polled again
+///         //
+///         // we have to box the errors so the types match
+///         let f = async move {
+///             tokio::select! {
+///                 res = fut => {
+///                     res.map_err(|err| err.into())
+///                 },
+///                 _ = timeout => {
+///                     Err(Box::new(Expired) as Box<dyn Error>)
```
LucioFranco, line 193:
```suggestion
///                     Err(Box::new(Expired) as Box<dyn Error + Send + Sync>)
```

tower-service/src/lib.rs
```diff
@@ -139,48 +137,84 @@ use std::task::{Context, Poll};
...
 ///
 ///     fn call(&mut self, req: Request) -> Self::Future {
-///         let timeout = tokio_timer::delay_for(self.timeout)
-///             .map(|_| Err(Self::Error::from(Expired)));
-///
-///         let fut = Box::pin(self.inner.call(req));
-///         let f = futures::select(fut, timeout)
-///             .map(|either| either.factor_first().0);
+///         // Create a future that completes after `self.timeout`
+///         let timeout = tokio::time::sleep(self.timeout);
```
LucioFranco, line 176: I wonder if we should use https://docs.rs/tokio/1.0.1/tokio/time/fn.timeout.html but I also kinda like that we show how to use `select!` and a `async move {}` instead of named types.

tower-service/src/lib.rs
```diff
@@ -139,48 +137,84 @@ use std::task::{Context, Poll};
...
+///
+///         // Call the inner service and get a future that resolves to the response
+///         let fut = self.inner.call(req);
+///
+///         // wrap those two futures in another future that completes when either one completes
```
LucioFranco, line 181: Nit: Can we keep the same style as above, capital letters to start and periods to end sentences?

tower/src/steer/mod.rs
```diff
@@ -26,7 +26,7 @@
 //!     let res = Response::new("Hello, World!".to_string());
 //!     Ok::<_, Infallible>(res)
 //! });
-//! // we have to box the service so its type gets erased and we can put in a `Vec` with other
+//! // we have to box the service so its type gets erased and we can put it in a `Vec` with other
```
LucioFranco, line 29: Same here.

tower-service/src/lib.rs
davidpdrsn: Yes. I’ll fix that later today.

davidpdrsn:
@LucioFranco fixed

tower-service/src/lib.rs
davidpdrsn: I picked `sleep` because it matches how the current timeout middleware works but I'll let you decide which is more appropriate.

LucioFranco:
@davidpdrsn can you resolve the conflicts then we can get this merged!

davidpdrsn:
@LucioFranco Done 😊

LucioFranco (approved):
:shipit: 🔥

PR #515, opened 2021-01-09
Fix all the docs links

davidpdrsn:
This goes through all the files and adds/fixes docs links. Also fixes typos and wordings in a few places.

### DIFF
--- a/tower-layer/src/identity.rs
+++ b/tower-layer/src/identity.rs
@@ -3,21 +3,25 @@ use std::fmt;
 
 /// A no-op middleware.
 ///
-/// When wrapping a `Service`, the `Identity` layer returns the provided
+/// When wrapping a [`Service`], the [`Identity`] layer returns the provided
 /// service without modifying it.
+///
+/// [`Service`]: https://docs.rs/tower-service/latest/tower_service/trait.Service.html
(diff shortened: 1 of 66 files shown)


tower-layer/src/identity.rs
```diff
@@ -3,21 +3,25 @@ use std::fmt;
 
 /// A no-op middleware.
 ///
-/// When wrapping a `Service`, the `Identity` layer returns the provided
+/// When wrapping a [`Service`], the [`Identity`] layer returns the provided
```
hawkw, line 6: nit/TIOLI: does this type's docs _need_ to link to its own definition? you're already reading its docs...

hawkw (approved):
looks good to me! we should maybe consider adding a CI check for broken doc links to make this easier in the future...

tower-layer/src/identity.rs
davidpdrsn: Yeah I did consider that as well however I don’t think it hurts and it’s just a single pair of `[]`. Also made it easier to search for everything inside backticks and make them into links.

davidpdrsn:
I've added `#![deny(broken_intra_doc_links)]` to all crates now.

PR #521, opened 2021-01-11
filter: expose predicate check as methods on `Filter` and `AsyncFilter`

hawkw:
## Motivation

In Linkerd 2, we have our own `RequestFilter` middleware. Now that Tower
0.4 has been released, including a new synchronous request filtering
middleware, we'd like to migrate to Tower's filter middleware. However,
there's one hitch: we implement additional traits for our filtering
middleware --- in particular, we use a `NewService` trait, which is
essentially a synchronous version of `MakeService`, for services that
can be constructed immediately. We'd like to be able to implement that
trait for `tower::filter::Filter` services as well.

## Solution

This branch adds new `Filter::check` and `AsyncFilter::check` methods
which expose the inner `Predicate::check` and `AsyncPredicate::check`
methods on the filter service's predicate. This can be used to call into
the predicate directly, allowing external traits to be implemented for
`Filter`/`AsyncFilter` in downstream code. Of course, the additional
traits must be defined in the crate providing implementaitons for
`tower::filter`, since `Filter` and `AsyncFilter` would be foreign
types, but in our use case, at least, this isn't an issue.

### DIFF
--- a/tower/src/filter/mod.rs
+++ b/tower/src/filter/mod.rs
@@ -72,6 +72,14 @@ impl<T, U> Filter<T, U> {
     pub fn layer(predicate: U) -> FilterLayer<U> {
         FilterLayer::new(predicate)
     }
+
+    /// Check a `Request` value against this filter's predicate.
+    pub fn check<Request>(&mut self, request: Request) -> Result<U::Request, BoxError>
+    where
+        U: Predicate<Request>,
+    {
+        self.predicate.check(request)
+    }
 }
 
 impl<T, U, Request> Service<Request> for Filter<T, U>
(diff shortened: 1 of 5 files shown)


tower/src/filter/mod.rs
```diff
@@ -72,6 +72,14 @@ impl<T, U> Filter<T, U> {
     pub fn layer(predicate: U) -> FilterLayer<U> {
         FilterLayer::new(predicate)
     }
+
+    /// Check a `Request` value against this filter's predicate.
+    pub fn check<Request>(&mut self, request: Request) -> Result<U::Request, BoxError>
```
LucioFranco, line 77: only nit I have is we have a lot of `Request` in here, maybe change the generic to be `R`?

LucioFranco (approved):

PR #525, opened 2021-01-12
util: impl `Clone` for `AndThen`, `MapRequest` and `MapErr` layers

hawkw:
The `Layer` types for `AndThen`, `MapRequest`, and `MapErr` middleware
were missing `Clone` impls, which the `MapResponse`, `MapResult`, and
`Then` middleware's layer types *do* have. These layers should all be
`Clone` when the function is `Clone`.

I've added `derive(Clone)` to all these layer types.

### DIFF
--- a/tower/src/util/and_then.rs
+++ b/tower/src/util/and_then.rs
@@ -47,7 +47,7 @@ where
 /// A [`Layer`] that produces a [`AndThen`] service.
 ///
 /// [`Layer`]: tower_layer::Layer
-#[derive(Debug)]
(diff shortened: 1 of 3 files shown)


LucioFranco (approved):

olix0r (approved):
You could probably derive `Copy` as well

hawkw:
> You could probably derive `Copy` as well

yes, although it's a potential compatibility issue if we ever add _additional_ fields that are not trivially `Copy`. i don't think that's too likely, but i punted on copy for now.

PR #529, opened 2021-01-14
tower: expand on usage docs, mention compatible libraries

hawkw:
This branch adds to the "Usage" section in the `tower` crate's README
and lib.rs docs. In particular, I've added a discussion of the various
use-cases for Tower, and a list of crates that support Tower.

Let me know if I'm missing anything!


davidpdrsn (approved):
Really good explanation! I'm sure this will help people wondering how to use Tower.

PR #531, opened 2021-01-15
util: Implement `Layer` for `Either<A, B>`

davidpdrsn:
I ran into a use case for this today which meant I had to make my own `Either` type. I guess makes sense for Tower to provide this directly.

My use case was that I wanted to optionally apply some layer depending on some config. I had an `Option<T>` which I wanted to map into a `Either<LayerUsingT, Identity>` and then treat that as a `Layer` in a larger stack.


olix0r (approved):
+1 LGTM

hawkw (approved):

Issue #530, opened 2021-01-15
Feature "full" does not exists in version 4.3


ava57r:
Hello
When I set 

`tower = { version = "0.4", features = ["full"] }`

from https://github.com/tower-rs/tower/blob/master/tower/src/lib.rs#L66

It does not work.

```
the package `test` depends on `tower`, with features: `full` but `tower` does not have these features.
```

string "full" does not exists in https://github.com/tower-rs/tower/blob/master/tower/Cargo.toml `[features]` section.

PR #532, opened 2021-01-15
Add "full" feature that turns on all other features

davidpdrsn:
Fixes https://github.com/tower-rs/tower/issues/530

### DIFF
--- a/tower/CHANGELOG.md
+++ b/tower/CHANGELOG.md
@@ -11,6 +11,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - **util**: Implement `Clone` for `FilterLayer`.
 - **timeout**: Implement `Clone` for `TimeoutLayer`.
 - **limit**: Implement `Clone` for `RateLimitLayer`.
+
+### Fixed
 - Added "full" feature which turns on all other features.
 
 # 0.4.3 (January 13, 2021)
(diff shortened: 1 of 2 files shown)


tower/CHANGELOG.md
```diff
@@ -3,6 +3,12 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+# Unreleased
+
+### Added
+- Added "full" feature which turns on all other features.
```
hawkw, line 10: nit/TIOLI: I might mark this as "Fixed", since the docs in previous releases mentioned this existing, but it was missing?

tower/Cargo.toml
```diff
@@ -25,6 +25,25 @@ edition = "2018"
 
 [features]
 default = ["log"]
+full = [
+  "log",
```
hawkw, line 29: the "log" feature is already turned on by default, so I don't think it should be part of this feature. That way, if users want to enable all the _middleware_ without using the `log` crate, they can write
```rust
tower = { version = "0.4", default_features = false, features = ["full"] }
```

davidpdrsn: Makes sense 👍 Fixed in bf1c665

tower/CHANGELOG.md
davidpdrsn: Ah right. I didn't realize it used to exist. Fixed in 3f4127a9e901a.

LucioFranco (approved):

Issue #262, opened 2019-04-23
util: type that implements `Service<T>` by cloning the shared value


seanmonstar:
A user may wish to satisfy a `MakeService<T, Req>` requirement with an existing `Service<Req>`, and have a clone used each time.

### Example

```rust
let double = service_fn(|req: u32| {
    Ok(req * 2)
});

let mk_svc = shared(double);

let one = mk_svc
    .oneshot("one")
    .and_then(|svc| svc.oneshot(1));

let two = mk_svc
    .oneshot("two")
    .and_then(|svc| svc.oneshot(2));
```

For an example implementation: https://github.com/linkerd/linkerd2-proxy/pull/236/files#diff-5de50f8a9530a389d2ebe255278303ab

### Possible names:

- Shared
- Const(ant)
- ???

carllerche:
`CloneService`?

davidbarsky:
I like `Shared`, but maybe `Fanout` could be a reasonable name?

I'd like to also nominate having this be a part of 0.1, as it solves the debatable misuse of `tower-buffer` as a cheap clone.

seanmonstar:
> I'd like to also nominate having this be a part of 0.1, as it solves the debatable misuse of tower-buffer as a cheap clone.

I believe they serve different purposes. The proposed type here is to allow some existing `impl Service + Clone` to act as an `impl MakeService`. Whether a buffer is used or not shouldn't affect this.

davidpdrsn:
I gave this a shot thinking it would be something like

```rust
pub fn shared<S, Request>(
    svc: S,
) -> impl MakeService<
    (),
    Request,
    Service = S,
    Response = S::Response,
    Error = S::Error,
    MakeError = Infallible,
>
where
    S: Service<Request> + Clone,
{
    service_fn(move |_| {
        let svc = svc.clone();
        async move { Ok(svc) }
    })
}
```

And while that works if you're only calling `.make_service()`:

```rust
#[tokio::test]
async fn using_make_service() {
    let echo_service = service_fn(|req: String| async {
        Ok::<_, Infallible>(req)
    });

    // Create a `MakeService` from `echo_service`
    let mut make_service = shared(echo_service);

    // Make a new instance of `echo_service`
    let mut new_svc = make_service.make_service(()).await.unwrap();

    // Call it
    let res = new_svc.call("foo".to_string()).await.unwrap();
    assert_eq!(res, "foo");
}
```

It doesn't work if going through `.into_service()`:

```rust
[...]

PR #533, opened 2021-01-17
make: Add `Shared`

davidpdrsn:
Fixes https://github.com/tower-rs/tower/issues/262

`Shared` is a `MakeService` that produces services by cloning an inner service.

I was wondering if it makes sense that `Shared` implements `Service` where the request is `()` since cloning doesn't require anything. Accepting some generic `Target` requires making a new trait which complicates the API a bit. Feels like at that point users would be better off writing their own custom implementation.


tower/src/make/make_service.rs
```diff
@@ -247,3 +247,80 @@ where
...
+}
+
+impl<S> Shared<S> {
+    /// Create a new [`Shared`] from a service.
+    pub fn new(service: S) -> Self {
+        Self { service }
+    }
+}
+
+impl<S> Service<()> for Shared<S>
```
olix0r, line 264: Couldn't this be generalized to any arbitrary target (that is just ignored)?

davidpdrsn: It could but could that perhaps be confusing to users looking at the docs?

hawkw: IMO, I agree with @olix0r --- this should _probably_ be `Service<T>` for an arbitrary `T`, since that way, a `Shared` can be passed into functions expecting _any_ `MakeService`. The current impl can _only_ be passed intop functions expecting a `MakeService<()>`. 

If it's possible to do that easily, I think this would be better.

LucioFranco: Would we also not want to enforce that `S: impl Service`?

olix0r: @LucioFranco I don't think that's strictly necessary. You may have a Service that returns a discovery stream, for instance, or some non-service response. Also, that bound would require binding the request type via phantomdata, which seems better to avoid.

davidpdrsn: Changing the impl to `Service<T>` leads to some inference errors in https://github.com/tower-rs/tower/blob/david/shared/tower/src/make/make_service.rs#L297:

```rust
error[E0282]: type annotations needed
   --> tower/src/make/make_service.rs:300:22
    |
300 |         poll_fn(|cx| MakeService::poll_ready(&mut shared, cx))
    |                      ^^^^^^^^^^^^^^^^^^^^^^^ cannot infer type for type parameter `Target` declared on the trait `MakeService`

error[E0282]: type annotations needed
   --> tower/src/make/make_service.rs:314:26
    |
314 |         let mut shared = MakeService::into_service(shared);
    |                          ^^^^^^^^^^^^^^^^^^^^^^^^^ cannot infer type for type parameter `Target` declared on the trait `MakeService`
````

Not sure if there is a way around those 🤔

Comment (davidpdrsn) [reply tower/src/make/make_service.rs]: Other than calling `MakeService::<(), _>::poll_ready(&mut shared, cx)` of course, but that isn't very elegant.

**Review by olix0r — APPROVED** [review]
lgtm

File: tower/src/make/make_service.rs [hunk 2]
```diff
@@ -247,3 +247,80 @@ where
...
+            .await
+            .unwrap();
+        let mut svc = shared.call(()).await.unwrap();
+
+        poll_fn(|cx| svc.poll_ready(cx)).await.unwrap();
+        let res = svc.call("foo").await.unwrap();
+
+        assert_eq!(res, "foo");
+    }
+}
```
hawkw, line 326: nit/TIOLI: this might be enough code to justify sticking it in a separate file (but I'd not expose the module and just reexport it from `make`)...

tower/src/make/make_service.rs
```diff
@@ -247,3 +247,80 @@ where
         self.make.make_service(target)
     }
 }
+
+/// A [`MakeService`] that produces services by cloning an inner service.
```
hawkw, line 251: It would be nice to have examples here, IMO...maybe with `buffer` to make a service `Clone`?

hawkw (approved):
this looks good to me! i'd like to see some examples but it's not really a blocker...

one thought: I wonder if we want to also add a `ServiceExt::shared` for turning `Service`s implementing `Clone` into `Shared`s? We could add that in a separate PR, though.

tower/src/make/make_service.rs
hawkw: i'm not sure if the type inference issues are _that_ big a deal, because I imagine in practice (e.g. not in tests), these are mostly going to be passed into functions taking `T: MakeService<Whatever, Req>` or similar?

tower/src/make/make_service.rs
davidpdrsn: 👍  7685a0761e82b

tower/src/make/make_service.rs
davidpdrsn: Sure thing. I haven't used `MakeService` much myself so not quite sure what a decent example might be? Any suggestions?

davidpdrsn: I added an example in 57776bf8f6aad

PR #535, opened 2021-01-18
Add clone impls to layers

davidpdrsn:
I ran into a missing `Clone` impl for `TimeoutLayer` today. Figure I would add that and look for other layers that probably should impl `Clone`.

Not sure if these were oversights or if there were reasons for not implementing `Clone`.


hawkw:
looks like the `cargo-deny` check hit a spurious network error or something, i went ahead and restarted it.

hawkw (approved):
:+1:

olix0r (approved):
thanks!

PR #536, opened 2021-01-19
spawn-ready: Remove useless `MakeSpawnReady` type

olix0r:
The `spawn_ready` module has two notable (public API) problems:

1. `MakeSpawnReady` is useless, as it doesn't use the target type in any
   novel way. It's nothing more than a `MapResponse`.
2. The `SpawnReadyLayer` type produces a `MakeSpawnReady` (which is, as
   mentioned, useless).

This change removes the `spawn_ready::make` module and modifies
`SpawnReadyLayer` to produce `SpawnReady` instances directly.


olix0r:
This PR depends on https://github.com/tower-rs/tower/pull/538

olix0r:
Again, this is a breaking change; but I wanted to open this for discussion. I'm responsible for the current mess and would love to clean it up eventually.

hawkw:
> the current state of `SpawnReadyLayer` is so convoluted that I can't imagine anyone is actually using it

Could be worth searching GitHub to see if it shows up in the wild...this wouldn't give us a correct negative signal, as there might be any number of private projects or projects on other git hosts using it, but we could at least get a positive signal if it turns out someone *is* using this...

davidpdrsn:
Not quite sure what your plan is but gut my feeling is that we shouldn't be afraid of bumping the version to 0.5 even if this is the only breaking change.

olix0r:
From a quick search, there appear to be no uses outside of this repo: https://github.com/search?q=SpawnReadyLayer&type=code

Though, I'm fine with bumping to 0.5 in this PR? (Should that happen in this PR or a distinct followup?)

davidpdrsn (approved):
I'm not very familiar with this area of the code but I _think_ this looks correct

olix0r:
@hawkw ptal

also: should i bump version to 0.5 in this PR or should that be done when prepping the release?

hawkw:
I haven't looked too closely at the actual change & its rationale yet, but I did want to weigh in on the breaking change: I think that we actually _can_ afford to bump versions for the `tower` crate fairly liberally, as long as we don't break `tower-service` or `tower-layer`. Those crates contain the traits that are necessary for compatibility, while `tower` itself doesn't. Thus, few other libraries expose `tower` itself (rather than `tower-layer` or `tower-service`) in their public APIs, and in many cases, it should be okay (though not ideal) to have multiple versions of `tower` in a dependency tree.

That said, I think we should try to get approval from everyone before shipping a breaking change --- at least, it would be nice to get a +1 from @LucioFranco as well. I'd also prefer to do a deprecation for `MakeSpawnReady` in an 0.4 release before we remove it in 0.5.

olix0r:
> I'd also prefer to do a deprecation for MakeSpawnReady in an 0.4 release before we remove it in 0.5.

I don't know how we do this gracefully when the intent is to change the signature of `SpawnReadyLayer`...

hawkw:
> > 
> 
> I don't know how we do this gracefully when the intent is to change the signature of `SpawnReadyLayer`...

this is what i get before weighing in before actually reading the change :P

> Also, I don't really think it's _important_ to do a deprecation cycle as we have no evidence it's actually used anywhere.

I don't think it's important either :) I just thought that, from the PR title, we were just doing a simple removal, so it seemed like we _could_ do a deprecation cycle quite easily, and we may as well do one. My bad!

tower/src/spawn_ready/service.rs
```diff
@@ -1,4 +1,4 @@
-use super::future::ResponseFuture;
+use super::{future::ResponseFuture, SpawnReadyLayer};
```
hawkw, line 1: clippy says this import is unused, because the `SpawnReady::layer` function references it from `super`:
https://github.com/tower-rs/tower/blob/49d77909515dbc4f5962f4cdfe5eaac7a89836b2/tower/src/spawn_ready/service.rs#L35-L37

we should remove this (or remove the `super` in `Layer`, I don't really care).
```suggestion
use super::future::ResponseFuture;
```

hawkw (approved):
I'm on board with this provided that @LucioFranco (and maybe also @jonhoo?) agree that shipping an 0.5 is reasonable.

Also, we may want to wait to do a quick pass to make sure there's nothing _else_ that ought to be in a breaking release before publishing.

olix0r:
Per discussion with @LucioFranco: It would be great if we didn't do a breaking release with _only_ such a minor change. So, I've marked this in https://github.com/tower-rs/tower/milestone/3 so we can bundle a set of breaking changes together.

hawkw:
now that we're prepping v0.5, let's get this merged!

hawkw:
@olix0r looks like this has some conflicts that will need to be resolved manually now...

olix0r:
@hawkw merged master

PR #538, opened 2021-01-19
spawn-ready: Avoid oneshot allocations

olix0r:
This change modifies the `SpawnReady` service to track a
`tokio::task::JoinHandle` instead of creating a `tokio::sync::oneshot`
on each pending poll. This avoid unnecessary allocation.

The `spawn_ready` test is also fixed to avoid using `thread::sleep`,
instead using `tokio::time::pause`.


tower/src/spawn_ready/service.rs
```diff
@@ -1,59 +1,57 @@
...
 
-impl<T, Request> Service<Request> for SpawnReady<T>
+impl<S, Req> Service<Req> for SpawnReady<S>
 where
-    T: Service<Request> + Send + 'static,
-    T::Error: Into<crate::BoxError>,
-    Request: Send + 'static,
+    Req: 'static,
+    S: Service<Req> + Send + 'static,
+    S::Error: Into<BoxError> + 'static,
```
LucioFranco, line 39: Is this additional lifetime a breaking change?

olix0r: Turns out it's not actually needed; removed.

LucioFranco (approved):

PR #541, opened 2021-01-20
Idea: Add `ServiceFn::poll_ready_fn`

davidpdrsn:
This is totally just an idea I had so deciding not merge this is cool 😊

Using `service_fn` together with `impl Service` is nice for making middlewares because you don't have to hand-code futures and can instead use `async/await`. However currently `ServiceFn::poll_ready` always return `Poll::Ready(Ok(())`. There is no way to customize that. That isn't great for middlewares which sometimes deals with back pressure or perhaps propagates it from their inner service.

This adds `ServiceFn::poll_ready_fn` which takes a closure that will be called in `poll_ready`. For `call` it uses the closure given to `service_fn`.

### DIFF
--- a/tower/src/util/service_fn.rs
+++ b/tower/src/util/service_fn.rs
@@ -30,3 +30,35 @@ where
         (self.f)(req)
     }
 }
+
+impl<T> ServiceFn<T> {
+    pub fn poll_ready_fn<F>(self, f: F) -> PollReadyFn<Self, F>
+    where
+        Self: Sized,
+    {
+        PollReadyFn { svc: self, f }
+    }
+}
+
+pub struct PollReadyFn<S, F> {
+    svc: S,
+    f: F,
+}
+
+impl<R, S, F> Service<R> for PollReadyFn<S, F>
+where
+    S: Service<R>,
+    F: FnMut(&mut Context<'_>) -> Poll<Result<(), S::Error>>,
+{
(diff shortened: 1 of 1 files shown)


hawkw:
This is interesting, and I think it could be worth having.

One concern about the potential usefulness of something like this, though, is that most non-trivial implementations of `poll_ready` depend on some kind of state that might be shared with the `call` method. For example, we might want to poll a channel in `poll_ready` and send a message on that channel in `call`, or poll one or more inner services in `poll_ready` and call them in `call`. However, we can't easily share state between `poll_ready` and `call` with two separate closures. Instead, they would need an `Arc<Mutex<TheSharedState>>` or something, which is cloned into both closures...which is a lot of unnecessary overhead and kind of defeats the point of having `poll_ready` and `call` defined on the _same_ data.  

So, the question I have is...have you tried this out in practice, and did it actually work for use-cases other than a `poll_ready` impl that always returns `Ready`? If this doesn't actually work in practice, I might have some other, similar ideas.

davidpdrsn:
I have done some more testing with this and I think you're right that it isn't very practical. If you're using this to write a middleware it pretty much always requires cloning the inner service to move it into both the `service_fn` and `poll_ready_fn` closures. That leaves you with two clones of a service and no guarantee that that readiness is shared and coordinated across the clones.

Therefore I'm gonna close this.

PR #542, opened 2021-01-20
util: Add `ServiceExt::map_future`

davidpdrsn:
I ran into a thing today where I wanted to write a middleware that wraps
all futures produced by a service in a `tracing::Span`. So something
like `self.inner.call(req).instrument(span)`.

Afaik all the combinators we have today receive the value produced by
the future and not the future itself. At that point its too late to call
`.instrument`. So I thought this made sense to add a combinator for.

If you agree I'll go write some docs and clean things up.

### DIFF
--- a/tower/src/util/map_future.rs
+++ b/tower/src/util/map_future.rs
@@ -1,27 +1,48 @@
 use std::{
+    fmt,
     future::Future,
     task::{Context, Poll},
 };
 use tower_layer::Layer;
 use tower_service::Service;
 
+/// [`Service`] returned by the [`map_future`] combinator.
+///
+/// [`map_future`]: crate::util::ServiceExt::map_future
+#[derive(Clone)]
 pub struct MapFuture<S, F> {
     inner: S,
     f: F,
 }
 
-#[derive(Debug, Clone)]
-pub struct MapFutureLayer<F> {
-    f: F,
-}
-
 impl<S, F> MapFuture<S, F> {
+    /// Creates a new [`MapFuture`] service.
     pub fn new(inner: S, f: F) -> Self {
         Self { inner, f }
     }
 
+    /// Returns a new [`Layer`] that produces [`MapFuture`] services.
+    ///
+    /// This is a convenience function that simply calls [`MapFutureLayer::new`].
+    ///
+    /// [`Layer`]: tower_layer::Layer
     pub fn layer(f: F) -> MapFutureLayer<F> {
-        MapFutureLayer { f }
+        MapFutureLayer::new(f)
+    }
+
+    /// Get a reference to the inner service
+    pub fn get_ref(&self) -> &S {
+        &self.inner
+    }
+
+    /// Get a mutable reference to the inner service
+    pub fn get_mut(&mut self) -> &mut S {
+        &mut self.inner
+    }
+
+    /// Consume `self`, returning the inner service
+    pub fn into_inner(self) -> S {
+        self.inner
     }
 }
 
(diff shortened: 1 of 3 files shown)


hawkw:
> I ran into a thing today where I wanted to write a middleware that wraps
> all futures produced by a service in a `tracing::Span`

(fwiw, for this use case in particular, i *really* should work on upstreaming the `tracing`/`tower` integration code I've written into `tower` itself)

I think this is probably worth having for the general case, though it's kind of hard to explain why you would want it until you already understand why you would want it...not totally sold on the naming but I guess it's probably the best option...

davidpdrsn:
> (fwiw, for this use case in particular, i really should work on upstreaming the tracing/tower integration code I've written into tower itself)

Perhaps 😅

> I think this is probably worth having for the general case, though it's kind of hard to explain why you would want it until you already understand why you would want it...not totally sold on the naming but I guess it's probably the best option...

Yes agreed. Writing the docs was also a bit tricky without involving `tracing::Instrument`.

I'm not sold on the name either. I also considered `map_call` but IMO thats worse.

Regardless, I've cleaned things up so ready for review 👀

tower/src/util/map_future.rs
```diff
@@ -0,0 +1,105 @@
...
+
+impl<R, S, F, T, E, Fut> Service<R> for MapFuture<S, F>
+where
+    S: Service<R>,
+    F: FnMut(S::Future) -> Fut,
+    E: From<S::Error>,
+    Fut: Future<Output = Result<T, E>>,
+{
+    type Response = T;
+    type Error = E;
```
hawkw, line 57: TIOLI: we could reduce the number of type parameters here using `TryFuture`:
```suggestion
impl<R, S, F, Fut> Service<R> for MapFuture<S, F>
where
    S: Service<R>,
    F: FnMut(S::Future) -> Fut,
    Fut: TryFuture,
    Fut::Error: From<S::Error>,
{
    type Response = Fut::Ok;
    type Error = Fut::Error;
```
not a big deal though...

tower/src/util/map_future.rs
```diff
@@ -0,0 +1,105 @@
...
+}
+
+impl<S, F> fmt::Debug for MapFuture<S, F>
+where
+    S: fmt::Debug,
+{
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("MapFuture")
+            .field("inner", &self.inner)
+            .field("f", &format_args!("<{}>", std::any::type_name::<F>()))
```
hawkw, line 76: nice, maybe we should do this for other combinator types with function parameters! (in a separate PR).

incidentally, I tried this out in the playground and it looks like `type_name` actually returns the name of function pointers, and the location of closures, which is quite nice: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=655a2434c5ffe8a85a1126ac24d1fe9d maybe we should actually omit the `<>`s?

tower/src/util/map_future.rs
```diff
@@ -0,0 +1,105 @@
...
+            .finish()
+    }
+}
+
+/// A [`Layer`] that produces a [`MapFuture`] service.
+///
+/// [`Layer`]: tower_layer::Layer
+#[derive(Debug, Clone)]
```
hawkw, line 84: this should probably also have a manual `Debug` impl rather than deriving it, since `F` is probably not going to ever be `Debug`.

hawkw (approved):
this looks good to me! i had a few thoughts...

tower/src/util/map_future.rs
davidpdrsn: Interesting! How does this work in terms of not exposing things from the `futures` crate?

tower/src/util/map_future.rs
davidpdrsn: Good idea. I made an issue https://github.com/tower-rs/tower/issues/545

`<>` removed in 4dc5ff0ba21fe.

tower/src/util/map_future.rs
davidpdrsn: 👍 5c15ea73ab37b

PR #548, opened 2021-01-27
service: Clarify subtly around cloning and readiness

davidpdrsn:
Adds a sections to the `Service` docs that mentions the subtlety around cloning services and readiness.


LucioFranco (approved):
Yeah, good catch and we should add this. We may also want to add it to the possible lists of panic that if you clone and don't poll ready that instance then it could panic. We could also ensure that our `tower-test` mock service covers this, we already do this somewhat but I don't remember if it catches this specific case.

davidpdrsn:
> We could also ensure that our `tower-test` mock service covers this, we already do this somewhat but I don't remember if it catches this specific case.

Thats a good idea. I'll look into that.

lilyball:
The examples here don't seem to call `self.inner.poll_ready()`, so isn't it still broken?

Issue #549, opened 2021-01-27
Tracing could be an optional dependency


rcoh:
As discussed on discord, tracing is only used in a few services (and even in those services, you might consider config-deffing it out?) In any case, it doesn't need to be a required dependency for the entire `tower` crate.

PR #551, opened 2021-01-27
Only pull in tracing when necessary

davidpdrsn:
Fixes https://github.com/tower-rs/tower/issues/549

Tested with `cargo hack check --each-feature --no-dev-deps`


hawkw (approved):
lgtm, thanks!

PR #555, opened 2021-02-09
Add `util::option_layer` and `ServiceBuilder::option_layer`

### LINKED ISSUE #553: Add `util::option_layer` and `ServiceBuilder::option_layer`
Hi, thanks for `tower`. [`kube`](https://github.com/clux/kube-rs) is now using it under the hood ([replaced `reqwest` with `tower` + `hyper`](https://github.com/clux/kube-rs/pull/394)) :tada:

We have a conditional Layer to handle certain authentication methods and found [`ServiceBuilderExt::optional_layer`](https://github.com/hyperium/tonic/blob/cb9ed1851873343ca77f5b9254bfb1d0835d80ed/tonic/src/transport/service/layer.rs) in `tonic` very useful. Can we have that included in `tower`?

We use it like the following:

```rust
let maybe_auth = match &auth {
    Authentication::None => None,
    Authentication::Basic(s) => {
        default_headers.insert(
            AUTHORIZATION,
            HeaderValue::from_str(s).map_err(Error::InvalidBasicAuth)?,
        );
        None
    }
    Authentication::Token(s) => {
        default_headers.insert(
            AUTHORIZATION,
            HeaderValue::from_str(s).map_err(Error::InvalidBearerToken)?,
        );
        None
    }
    Authentication::RefreshableToken(r) => Some(AuthLayer::new(r.clone())),
};
let common = ServiceBuilder::new()
    // ... omitted
    .map_request(move |r| set_default_headers(r, default_headers.clone()))
    .into_inner();
ServiceBuilder::new()
    .layer(common)
    .optional_layer(maybe_auth)
    .service(client)
```

kazk:
Closes #553


tower/src/builder/mod.rs
```diff
@@ -134,6 +134,16 @@ impl<L> ServiceBuilder<L> {
         }
     }
 
+    /// Optionally add a new layer `T` into the [`ServiceBuilder`].
```
davidpdrsn, line 137: Maybe an example here would be nice.

tower/src/util/mod.rs
```diff
@@ -938,3 +940,12 @@ pub trait ServiceExt<Request>: tower_service::Service<Request> {
 }
 
 impl<T: ?Sized, Request> ServiceExt<Request> for T where T: tower_service::Service<Request> {}
+
+/// Convert an `Option<Layer>` into a `Layer`
```
davidpdrsn, line 944: I think we should also have a small example here.

davidpdrsn:
Thanks 👍 

Should probably also be mentioned in the [changelog](https://github.com/tower-rs/tower/blob/master/tower/CHANGELOG.md).

tower/src/util/mod.rs
```diff
@@ -938,3 +940,12 @@ pub trait ServiceExt<Request>: tower_service::Service<Request> {
...
+pub fn option_layer<L>(layer: Option<L>) -> Either<L, Identity> {
```
hawkw, line 945: An alternative to having a special wrapper function would be  to just add an `impl<L: Layer> for Option<L>`. That might have slightly nicer API ergonomics, since _all_ `Option<Layer>`s just _are_ layers, rather than having to wrap them in `option_layer` calls.

But, the downside to that is that this would have to be implemented in the `tower_layer` crate, rather than in `tower`. This means that we couldn't use `Either`, which is defined in `tower`...

kazk:
Thanks for reviewing. Added minimal examples with `TimeoutLayer` and updated changelog.

kazk:
Fixed doctests. Sorry for the noise.

tower/src/util/mod.rs
davidpdrsn: I actually like that `option_layer` is explicit about whats happening. When looking at large layer stack it might be obvious which layers are optional and which aren't. Is `TimeoutLayer` applied

davidpdrsn: I actually like that `option_layer` is explicit about whats happening. When looking at large layer stack being built with `ServiceBuilder` it might be obvious which layers are optional and which aren't. Is this `TimeoutLayer` applied all the time or only sometimes? Would have to check the type.

That combined with the annoyances of having to copy+paste `Either` (or depend on the [either](https://crates.io/crates/either) crate) makes me think this approach is fine.

hawkw: @davidpdrsn Yeah, I think that makes sense. I'm also in favor of avoiding `tower-service` changes when they're not absolutely necessary. We can always add an impl in `tower-service` later if we end up deciding that _is_ a good idea...

hawkw: for API consistency's sake, WDYT about adding an `option_service` (or something) that takes an `Option<T: Service>` and returns an [`Optional<T>`](https://docs.rs/tower/0.4.4/tower/util/struct.Optional.html)? That way, we'd have `option_layer` and `option_service` functions. 

I'd add that in a separate PR, though.

tower/src/util/mod.rs
```diff
@@ -938,3 +940,32 @@ pub trait ServiceExt<Request>: tower_service::Service<Request> {
...
+/// # #[cfg(feature = "timeout")]
+/// # use tower::timeout::TimeoutLayer;
+/// # #[cfg(all(feature = "timeout", feature = "util"))]
+/// # async fn wrap<S>(svc: S) where S: Service<(), Error = &'static str> + 'static + Send, S::Future: Send {
+/// # let timeout = Some(Duration::new(10, 0));
+/// // Layer to apply a timeout if configured
+/// let maybe_timeout = option_layer(timeout.map(TimeoutLayer::new));
+/// # ServiceBuilder::new()
+/// #     .layer(maybe_timeout)
+/// #     .service(svc);
```
hawkw, line 960: TIOLI: it might be worth actually showing this being used:
```suggestion
///  ServiceBuilder::new()
///      .layer(maybe_timeout)
///      .service(svc);
```

tower/src/util/mod.rs
```diff
@@ -938,3 +940,32 @@ pub trait ServiceExt<Request>: tower_service::Service<Request> {
...
 
 impl<T: ?Sized, Request> ServiceExt<Request> for T where T: tower_service::Service<Request> {}
+
+/// Convert an `Option<Layer>` into a [`Layer`].
+///
+/// ```
+/// # use std::time::Duration;
+/// # use tower::Service;
+/// # use tower::builder::ServiceBuilder;
+/// # use tower::util::option_layer;
```
hawkw, line 950: TIOLI: in general, I think we do tend to show the type or function in question being imported, in these examples:
```suggestion
/// use tower::util::option_layer;
```

tower/src/builder/mod.rs
```diff
@@ -134,6 +134,33 @@ impl<L> ServiceBuilder<L> {
...
+    ///
+    /// ```
+    /// # use std::time::Duration;
+    /// # use tower::Service;
+    /// # use tower::builder::ServiceBuilder;
+    /// # #[cfg(feature = "timeout")]
+    /// # use tower::timeout::TimeoutLayer;
+    /// # #[cfg(all(feature = "timeout", feature = "util"))]
```
hawkw, line 145: as i mentioned on the free function's docs, the `cfg` attributes don't actually work:
```suggestion
    /// # use tower::timeout::TimeoutLayer;
```

hawkw (changes requested):
I had some minor nits about the examples. Other than that, this change looks great and I'll happily merge it once the docs feedback is addressed.

Thanks for working on this @kazk!

tower/src/util/mod.rs
davidpdrsn: Yeah that sounds good.

hawkw (approved):
looks great, thanks again!

PR #557, opened 2021-02-10
spawn_ready: propagate `tracing` spans

hawkw:
Currently, when using `SpawnReady`, the current `tracing` span is not
propagated to the spawned task when the inner service is not ready. This
means that any traces emitted by the inner service's `poll_ready` occur
in their own root span, rather than the span the `SpawnReady` service
was polled in.

This branch fixes this by propagating the current trace span to the
spawned task.

This means that "spawn-ready" now enables the "tracing" feature. In the
future, we may want to consider feature-flagging `tracing` separately
from the middleware implementation that contain tracing instrumentation,
but doing so would break traces if the feature flag isn't enabled. This
doesn't break API compatibility, but it *does* break functionality, so
we may not want to do that until the next breaking change release.

I also added tests for span propagation. I realized the same test could
easily be applied to `Buffer`, which also propagates `tracing` spans, to
guard against regressions, so I also added a test for buffer.


tower/tests/spawn_ready/main.rs
```diff
@@ -43,3 +43,22 @@ async fn when_inner_fails() {
...
+    let _t = support::trace_init();
+
+    let span = tracing::info_span!("my_span");
+
+    let service = support::AssertSpanSvc::new(span.clone());
+    let mut service = SpawnReady::new(service);
+    let result =
+        tokio::spawn(async move { service.ready_and().await?.call(()).await }.instrument(span));
+
+    time::sleep(time::Duration::from_millis(100)).await;
```
olix0r, line 62: why is the sleep necessary? perhaps a comment explaining it?

tower/tests/support.rs
```diff
@@ -33,3 +36,66 @@ impl<I> Stream for IntoStream<mpsc::UnboundedReceiver<I>> {
         self.project().0.poll_recv(cx)
     }
 }
+
+#[derive(Clone, Debug)]
+pub struct AssertSpanSvc {
+    span: tracing::Span,
+    polled: bool,
+}
+pub struct AssertSpanError(String);
```
olix0r, line 45:
```suggestion
}

pub struct AssertSpanError(String);
```

tower/tests/spawn_ready/main.rs
```diff
@@ -43,3 +43,22 @@ async fn when_inner_fails() {
...
+
+    time::pause();
+    let _t = support::trace_init();
+
+    let span = tracing::info_span!("my_span");
+
+    let service = support::AssertSpanSvc::new(span.clone());
+    let mut service = SpawnReady::new(service);
+    let result =
+        tokio::spawn(async move { service.ready_and().await?.call(()).await }.instrument(span));
```
olix0r, line 60:
```suggestion
        tokio::spawn(service.oneshot(()).instrument(span));
```

tower/tests/buffer/main.rs
```diff
@@ -346,6 +346,26 @@ async fn wakes_pending_waiters_on_failure() {
...
+    let _t = support::trace_init();
+
+    let span = tracing::info_span!("my_span");
+
+    let service = support::AssertSpanSvc::new(span.clone());
+    let (mut service, worker) = Buffer::pair(service, 5);
+    let worker = tokio::spawn(worker);
+
+    let result =
+        tokio::spawn(async move { service.ready_and().await?.call(()).await }.instrument(span));
```
olix0r, line 363:
```suggestion
        tokio::spawn(service.oneshot(()).instrument(span));
```

olix0r (approved):

tower/tests/spawn_ready/main.rs
olix0r: If we really do need a sleep, can we use `time::pause`?

hawkw: there's a `time::pause` here: https://github.com/tower-rs/tower/blob/0408e315b5eef267e070d9aa37dbae900c1676b0/tower/tests/spawn_ready/main.rs#L52

I added this because the other `SpawnReady` tests have a sleep...although after looking at it more closely I don't think it's necessary in any case. I'll remove them in a separate branch.

hawkw: to elaborate: i'm _guessing_ the sleep was originally added to give the background task time to run, and then someone did a pass where they came through and slapped a `time::pause` in every test that has a sleep in it...so now, i think the sleep basically just does nothing.

olix0r: ah, yeah, that's me

hawkw: lol...i think a more robust solution would be just changing these tests to use `ready_and().await` instead of sleeping?

olix0r: i could see sleeping where we poll once (as i imagine this was initially), but with await it seems totally unnecessary.

PR #560, opened 2021-02-13
builder: Add `ServiceBuilder::layer_fn`

davidpdrsn:
This gem as just [sitting in Tonic][sit] waiting to be upstreamed.

[sit]: https://github.com/hyperium/tonic/blob/master/tonic/src/transport/service/layer.rs

### DIFF
--- a/tower/src/builder/mod.rs
+++ b/tower/src/builder/mod.rs
@@ -159,6 +159,15 @@ impl<L> ServiceBuilder<L> {
         self.layer(crate::util::option_layer(layer))
     }
 
+    /// Add a [`Layer`] built from a function that accepts a service and returns another service.
+    ///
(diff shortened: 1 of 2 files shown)


tower/src/builder/mod.rs
```diff
@@ -159,6 +159,15 @@ impl<L> ServiceBuilder<L> {
         self.layer(crate::util::option_layer(layer))
     }
 
+    /// Add a [`Layer`] built from a function that accepts a service and returns another service.
+    ///
+    /// See the documentation for [`layer_fn`] for more details.
+    ///
```
hawkw, line 165: nit/TIOLI: an example _might_ be nice, but the link to `layer_fn` is probably sufficient...

hawkw (approved):
lovely, thank you!

PR #562, opened 2021-02-15
Multiplex middleware

davidpdrsn:
This is a middleware that I've been thinking about for some time and I would like to hear what people think.

For tower-http I would like to provide some basic routing functionality. The goal isn't to compete with warp but to provide something for users who just want to route requests between a couple services.

That can in theory be done with `Steer` but I think it has a few limitations:

- The services must all be the same type.
- You cannot use `ServiceBuilder` to compose things. You have to build each service separately and then put them all in a `Vec`.
- Returning indexes into a `Vec` feels quite low level and error prone.

`Multiplex` serves a similar purpose but only dispatches to one of two services. It has a `P: Picker` that returns `Pick::First` or `Pick::Second` which is how we know which service to call.

This would allow you compose a service by doing something like this (where `method` and `path` are methods for building pickers that work on `Request<B>`):

```rust
ServiceBuilder::new()
    .layer(CompressionLayer::new()) // middleware applied to all lower services
    .multiplex(service_fn(root), method(Method::GET).and(path("/")))
    .multiplex(service_fn(users_index), method(Method::GET).and(path("/users")))
    .multiplex(service_fn(users_show), method(Method::GET).and(path("/users/:id")))
    .multiplex(service_fn(users_create), method(Method::POST).and(path("/users")))
    .service(service_fn(not_found));
```

The main downside to this API is that you have `O(n)` time to route a request, where `n` is the number of services. Getting to the bottom service requires checking all the ones above it. `Steer` doesn't have that limitation. However afaik warp works the same way since filters are built and composed individually, though I might be mistaking. In practice I don't think this would be a problem unless users have well beyond 100 leaf services.

### DIFF
--- a/tower/src/multiplex/service.rs
+++ b/tower/src/multiplex/service.rs
@@ -1,0 +1,100 @@
+use super::{Pick, Picker};
+use crate::util::Either;
+use crate::BoxError;
+use futures_util::ready;
+use pin_project::pin_project;
+use std::task::{Context, Poll};
+use std::{future::Future, pin::Pin};
+use tower_service::Service;
+
+/// TODO(david): docs
+#[derive(Debug, Clone, Copy)]
+pub struct Multiplex<A, B, P> {
+    picker: P,
+    first: A,
+    first_ready: bool,
+    second: B,
+    second_ready: bool,
+}
+
+impl<A, B, P> Multiplex<A, B, P> {
+    /// TODO(david): docs
+    pub fn new(first: A, second: B, picker: P) -> Self {
+        Self {
+            picker,
+            first,
+            first_ready: false,
+            second,
+            second_ready: false,
+        }
+    }
+}
+
+impl<R, P, A, B> Service<R> for Multiplex<A, B, P>
+where
+    P: Picker<R>,
+    A: Service<R>,
+    A::Error: Into<BoxError>,
+    B: Service<R, Response = A::Response>,
+    B::Error: Into<BoxError>,
+{
+    type Response = A::Response;
+    type Error = BoxError;
+    type Future = ResponseFuture<A::Future, B::Future>;
+
+    fn poll_ready(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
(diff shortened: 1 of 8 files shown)


olix0r:
I'm generally a fan of this approach. I could probably bikeshed names; but I don't think `multiplex` is particularly bad as far as difficult names go...

My hangup with _multiplex_ is that, to me, it means it can support more than one in-flight request. I.e. i think of request multiplexing and not service composition. Something like `on_predicate` might more clearly indicate "this service is used when this predicate is matched on the request" which is less abstract of a concept.

davidpdrsn:
Yeah I’m up for bike shedding the name. The name multiplex was inspired by this example in tonic https://github.com/hyperium/tonic/blob/master/examples/src/multiplex/server.rs

tower/src/multiplex/service.rs
```diff
@@ -0,0 +1,100 @@
...
+    A: Service<R>,
+    A::Error: Into<BoxError>,
+    B: Service<R, Response = A::Response>,
+    B::Error: Into<BoxError>,
+{
+    type Response = A::Response;
+    type Error = BoxError;
+    type Future = ResponseFuture<A::Future, B::Future>;
+
+    fn poll_ready(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
```
LucioFranco, line 45: I feel like this style of polling will be limited for a general implementation. Tonic and Warp can get away because we can guarantee that our server services generally don't need to be polled ready (they are always ready).

davidpdrsn: In my experience it hasn't been an issue since the services you combine are at the leaves and generally always are ready. But it is something to be aware of.

I'm not sure there is a nice way around it 🤔

LucioFranco:
Do we maybe want to move this into a separate crate that we could treat as unstable kinda like tokio-util?

davidpdrsn:
Yeah thats probably a good idea. I have to do some more experimentation in tower-http regardless.

I'll close this for now.

PR #563, opened 2021-02-21
util: Add example for `service_fn`

davidpdrsn:


tower/src/util/service_fn.rs
```diff
@@ -4,11 +4,52 @@ use std::task::{Context, Poll};
 use tower_service::Service;
 
 /// Returns a new [`ServiceFn`] with the given closure.
+///
+/// This lets you build a [`Service`] from an async function that returns a [`Result`].
```
hawkw, line 8: do these need link references?

hawkw (approved):
great, this looks awesome!

tower/src/util/service_fn.rs
hawkw: ...oh, we have `deny(broken_intra_doc_links)` on already, lol...i guess if these were broken, we'd know.

PR #564, opened 2021-02-21
builder: Add `ServiceBuilder::service_fn`

davidpdrsn:
A small convenience for doing `.service(service_fn(handler_function))`.

### DIFF
--- a/tower/src/builder/mod.rs
+++ b/tower/src/builder/mod.rs
@@ -459,6 +459,22 @@ impl<L> ServiceBuilder<L> {
     {
         self.layer.layer(service)
     }
+
+    /// Wrap the async function `F` with the middleware provided by this [`ServiceBuilder`]'s
+    /// [`Layer`]'s, returning a new [`Service`].
+    ///
+    /// This is a convenience for doing `.service(service_fn(handler_function))`.
+    ///
+    /// [`Layer`]: crate::Layer
+    /// [`Service`]: crate::Service
+    #[cfg(feature = "util")]
+    #[cfg_attr(docsrs, doc(cfg(feature = "util")))]
+    pub fn service_fn<F>(self, f: F) -> L::Service
+    where
+        L: Layer<crate::util::ServiceFn<F>>,
+    {
+        self.service(crate::util::service_fn(f))
+    }
 }
 
 impl<L: fmt::Debug> fmt::Debug for ServiceBuilder<L> {
--- a/tower/src/builder/mod.rs
+++ b/tower/src/builder/mod.rs
@@ -474,20 +474,22 @@ impl<L> ServiceBuilder<L> {
     ///
     /// This is a convenience method which is equivalent to calling
     /// [`ServiceBuilder::service`] with a [`service_fn`], like this:
+    ///
     /// ```rust
-    /// # use tower::{ServiceBuilder, ServiceFn};
+    /// # use tower::{ServiceBuilder, service_fn};
     /// # async fn handler_fn(_: ()) -> Result<(), ()> { Ok(()) }
     /// # let _ = {
     /// ServiceBuilder::new()
     ///     // ...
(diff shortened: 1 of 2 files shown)


hawkw (approved):

tower/src/builder/mod.rs
```diff
@@ -459,6 +459,22 @@ impl<L> ServiceBuilder<L> {
     {
         self.layer.layer(service)
     }
+
+    /// Wrap the async function `F` with the middleware provided by this [`ServiceBuilder`]'s
+    /// [`Layer`]'s, returning a new [`Service`].
+    ///
+    /// This is a convenience for doing `.service(service_fn(handler_function))`.
+    ///
```
hawkw, line 467: nit: might be nice to have an example?

tower/src/builder/mod.rs
```diff
@@ -459,6 +459,22 @@ impl<L> ServiceBuilder<L> {
     {
         self.layer.layer(service)
     }
+
+    /// Wrap the async function `F` with the middleware provided by this [`ServiceBuilder`]'s
+    /// [`Layer`]'s, returning a new [`Service`].
+    ///
+    /// This is a convenience for doing `.service(service_fn(handler_function))`.
```
hawkw, line 466: nit: might be nice if this linked to `service_fn` someplace?

tower/src/builder/mod.rs
davidpdrsn: Sure thing!

tower/src/builder/mod.rs
davidpdrsn: Good idea. I made `.service(service_fn(handler_function))` link to it. Doesn't think that totally makes sense but not sure where else to put it 🤷

tower/src/builder/mod.rs
```diff
@@ -459,6 +459,47 @@ impl<L> ServiceBuilder<L> {
     {
         self.layer.layer(service)
     }
+
+    /// Wrap the async function `F` with the middleware provided by this [`ServiceBuilder`]'s
+    /// [`Layer`]'s, returning a new [`Service`].
+    ///
+    /// This is a convenience for doing [`.service(service_fn(handler_function))`][`service_fn`].
```
hawkw, line 466: nit, take it or leave it: maybe something like this is slightly nicer:
```suggestion
    /// This is a convenience method which is equivalent to calling
    /// [`ServiceBuilder::service`] with a [`service_fn`], like this:
    /// ```rust
    /// # use tower::{ServiceBuilder, ServiceFn};
    /// # async fn handler_fn(_: ()) -> Result<(), ()> { Ok(()) }
    /// # let _ = {
    /// ServiceBuilder::new()
    ///     // ...
    ///     .service(service_fn(handler_fn))
    /// # };
```

tower/src/builder/mod.rs
```diff
@@ -459,6 +459,47 @@ impl<L> ServiceBuilder<L> {
...
+    ///
+    /// let response = svc.oneshot("foo").await?;
+    ///
+    /// assert_eq!(response, "foo");
+    /// # Ok(())
+    /// # }
+    /// ```
+    ///
+    /// [`Layer`]: crate::Layer
+    /// [`Service`]: crate::Service
```
hawkw, line 493: do these links actually work with the reexports, or do we need to link to the original crates? could be good to test that out in a docs build.

hawkw (approved):
I commented on some potential docs improvements, but other than that, this looks great!

tower/src/builder/mod.rs
hawkw: again, turns out we're already checking intra-docs links on CI, and i don't need to worry about this, lol.

tower/src/builder/mod.rs
davidpdrsn: Much nicer!

PR #567, opened 2021-02-24
Rename `ServiceExt::ready_and` to `ServiceExt::ready`

davidbarsky:
This PR renames:
- `ServiceExt::ready_and` to `ServiceExt::ready`
- the `ReadyAnd` future to `Ready`
-  the associated documentation to refer to `ServiceExt::ready` and `ReadyAnd`.

My recollection of the original conversation surrounding the introduction of the `ServiceExt::ready_and` combinator in https://github.com/tower-rs/tower/pull/427 was that it was meant to be a temporary workaround for the unchainable `ServiceExt::ready` combinator until the next breaking release of the Tower crate. The unchainable `ServiceExt::ready` combinator was removed, but `ServiceExt::ready_and` was not renamed. I believe, but am not 100% sure, that this was an oversight. 

I also recognize that this PR might have a large impact. To reduce the impact, I'd be happy to modify this PR to introduce `ServiceExt::ready` and deprecate `ServiceExt::ready_and`.

### DIFF
--- a/tower/src/util/mod.rs
+++ b/tower/src/util/mod.rs
@@ -30,7 +30,7 @@ pub use self::{
     map_result::{MapResult, MapResultLayer},
     oneshot::Oneshot,
     optional::Optional,
-    ready::{ReadyAnd, ReadyOneshot},
+    ready::{Ready, ReadyOneshot},
     service_fn::{service_fn, ServiceFn},
     then::{Then, ThenLayer},
 };
@@ -61,11 +61,11 @@ pub mod future {
 /// adapters
 pub trait ServiceExt<Request>: tower_service::Service<Request> {
     /// Yields a mutable reference to the service when it is ready to accept a request.
-    fn ready_and(&mut self) -> ReadyAnd<'_, Self, Request>
+    fn ready(&mut self) -> Ready<'_, Self, Request>
     where
         Self: Sized,
     {
-        ReadyAnd::new(self)
+        Ready::new(self)
     }
 
     /// Yields the service when it is ready to accept a request.
@@ -222,7 +222,7 @@ pub trait ServiceExt<Request>: tower_service::Service<Request> {
     /// // Call the new service
     /// let id = 13;
     /// let name = new_service
-    ///     .ready_and()
+    ///     .ready()
     ///     .await?
     ///     .call(id)
     ///     .await?;
@@ -289,7 +289,7 @@ pub trait ServiceExt<Request>: tower_service::Service<Request> {
     /// // Call the new service
     /// let id = 13;
     /// let code = new_service
-    ///     .ready_and()
+    ///     .ready()
     ///     .await?
     ///     .call(id)
     ///     .await
@@ -390,7 +390,7 @@ pub trait ServiceExt<Request>: tower_service::Service<Request> {
     /// // Call the new service
     /// let id = 13;
     /// let name = new_service
-    ///     .ready_and()
+    ///     .ready()
     ///     .await?
     ///     .call(id)
     ///     .await?;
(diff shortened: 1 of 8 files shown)


hawkw:
> I also recognize that this PR might have a large impact. To reduce the impact, I'd be happy to modify this PR to introduce `ServiceExt::ready` and deprecate `ServiceExt::ready_and`.

let's go ahead and do this, please — that way, we can land this now, and remove `ready_and` in 0.5.

davidpdrsn (approved):
Yes! I've also thought that `ready_and` was an awkward name.

davidbarsky:
> let's go ahead and do this, please — that way, we can land this now, and remove ready_and in 0.5.

Done! I've also opened https://github.com/tower-rs/tower/issues/568 as a reminder.

jonhoo (approved):

tower/src/util/mod.rs
```diff
@@ -30,7 +30,7 @@ pub use self::{
     map_result::{MapResult, MapResultLayer},
     oneshot::Oneshot,
     optional::Optional,
-    ready::{ReadyAnd, ReadyOneshot},
+    ready::{Ready, ReadyAnd, ReadyOneshot},
```
hawkw, line 33: i think we need to stick an `allow(deprecated)` here...it might have to go at the top of the `use` statement though, I can't remember...
```suggestion
    #[allow(deprecated)]
    ready::{Ready, ReadyAnd, ReadyOneshot},
```

tower/src/util/mod.rs
```diff
@@ -61,6 +61,18 @@ pub mod future {
...
+    {
+        Ready::new(self)
+    }
+
+    /// Yields a mutable reference to the service when it is ready to accept a request.
+    #[deprecated(
+        since = "0.4.6",
+        note = "Please use the ServiceExt::ready method instead"
+    )]
     fn ready_and(&mut self) -> ReadyAnd<'_, Self, Request>
```
hawkw, line 76: we need to stick an `allow(deprecated)` somewhere so this doesn't warn due to the use of the `ReadyAnd` type.

tower/src/util/mod.rs
```diff
@@ -61,6 +61,18 @@ pub mod future {
...
+    where
+        Self: Sized,
+    {
+        Ready::new(self)
+    }
+
+    /// Yields a mutable reference to the service when it is ready to accept a request.
+    #[deprecated(
+        since = "0.4.6",
+        note = "Please use the ServiceExt::ready method instead"
```
hawkw, line 74: nit: going off the stdlib's example, i think these usually start with a lowercase letter, since they're part of a sentence, include backticks for Rust type names (which renders in the RustDoc). like:
```suggestion
        note = "please use the `ServiceExt::ready` method instead"
```

tower/src/util/ready.rs
```diff
@@ -62,17 +62,25 @@ where
...
+/// [`ServiceExt::ready`]: crate::util::ServiceExt::ready
+pub struct Ready<'a, T, Request>(ReadyOneshot<&'a mut T, Request>);
+
 /// A future that yields a mutable reference to the service when it is ready to accept a request.
 ///
 /// [`ReadyAnd`] values are produced by [`ServiceExt::ready_and`].
 ///
 /// [`ServiceExt::ready_and`]: crate::util::ServiceExt::ready_and
-pub struct ReadyAnd<'a, T, Request>(ReadyOneshot<&'a mut T, Request>);
+#[deprecated(since = "0.4.6", note = "Please use the Ready future instead")]
```
hawkw, line 77:
```suggestion
#[deprecated(since = "0.4.6", note = "please use the `Ready` future instead")]
```

hawkw (changes requested):
couple of nits --- can we fix the new warnings added in this PR? i think the other ones were already present, so we can fix those in a follow-up, but this adds new deprecation warnings due to internal use of deprecated APIs.

besides that, lgtm!

hawkw (approved):
looks good to me — thanks for fixing the internal deprecation warnings!

PR #569, opened 2021-02-25
util: Add `BoxLayer`

davidpdrsn:
I've run into a use case where I need to return a `Layer` with a complex type from a function. I previously used `impl Layer<S, Service = impl Service<...>>` but that impacted compile times quite significantly. Boxing the `Layer` fixed it. Thought it made sense to upstream.

The `Send + Sync + 'static` bounds on the inner `dyn Layer` were necessary to get it working with hyper.

### DIFF
--- a/tower/src/util/boxed/layer.rs
+++ b/tower/src/util/boxed/layer.rs
@@ -3,14 +3,20 @@ use std::{fmt, sync::Arc};
 use tower_layer::{layer_fn, Layer};
 use tower_service::Service;
 
-/// A boxed `Layer` trait object.
+/// A boxed [`Layer`] trait object.
 ///
-/// [`BoxLayer`] turns a layer into a trait object, allowing the output service to be dynamic.
+/// [`BoxLayer`] turns a layer into a trait object, allowing both the [`Layer`] itself
+/// and the output [`Service`] to be dynamic, while having consistent types.
+///
+/// This [`Layer`] produces [`BoxService`] instances erasing the type of the 
+/// [`Service`] produced by the wrapped [`Layer`].
 ///
 /// # Example
 ///
-/// `BoxLayer` can for example be useful to create layers dynamically that otherwise wouldn't have
-/// the same types.
+/// `BoxLayer` can, for example, be useful to create layers dynamically that otherwise wouldn't have
+/// the same types. In this example, we include a [`Timeout`] layer
+/// only if an environment variable is set. We can use `BoxLayer`
+/// to return a consistent type regardless of runtime configuration:
 ///
 /// ```
 /// use std::time::Duration;
@@ -40,6 +46,11 @@ use tower_service::Service;
 ///     }
 /// }
 /// ```
+///
+/// [`Layer`]: tower_layer::Layer
+/// [`Service`]: tower_service::Service
+/// [`BoxService`]: super::BoxService
+/// [`Timeout`]: crate::timeout
 pub struct BoxLayer<In, T, U, E> {
     boxed: Arc<dyn Layer<In, Service = BoxService<T, U, E>> + Send + Sync + 'static>,
 }
--- a/tower/src/util/boxed/layer.rs
+++ b/tower/src/util/boxed/layer.rs
@@ -1,0 +1,86 @@
+use crate::util::BoxService;
+use std::{fmt, sync::Arc};
+use tower_layer::{layer_fn, Layer};
+use tower_service::Service;
+
+/// A boxed `Layer` trait object.
+///
+/// [`BoxLayer`] turns a layer into a trait object, allowing the output service to be dynamic.
+///
+/// # Example
+///
+/// `BoxLayer` can for example be useful to create layers dynamically that otherwise wouldn't have
+/// the same types.
+///
+/// ```
+/// use std::time::Duration;
+/// use tower::{Service, ServiceBuilder, BoxError, util::BoxLayer};
+///
+/// fn common_layer<S, T>() -> BoxLayer<S, T, S::Response, BoxError>
+/// where
+///     S: Service<T> + Send + 'static,
+///     S::Future: Send + 'static,
+///     S::Error: Into<BoxError> + 'static,
+/// {
+///     let builder = ServiceBuilder::new()
+///         .concurrency_limit(100);
+///
+///     if std::env::var("SET_TIMEOUT").is_ok() {
+///         let layer = builder
+///             .timeout(Duration::from_secs(30))
+///             .into_inner();
+///
+///         BoxLayer::new(layer)
+///     } else {
+///         let layer = builder
+///             .map_err(Into::into)
+///             .into_inner();
+///
+///         BoxLayer::new(layer)
+///     }
+/// }
+/// ```
+pub struct BoxLayer<In, T, U, E> {
+    boxed: Arc<dyn Layer<In, Service = BoxService<T, U, E>> + Send + Sync + 'static>,
+}
+
+impl<In, T, U, E> BoxLayer<In, T, U, E> {
+    /// Create a new [`BoxLayer`].
+    pub fn new<L>(inner_layer: L) -> Self
+    where
+        L: Layer<In> + Send + Sync + 'static,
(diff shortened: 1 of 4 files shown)


tower/src/util/boxed/layer.rs
```diff
@@ -0,0 +1,86 @@
+use crate::util::BoxService;
+use std::{fmt, sync::Arc};
+use tower_layer::{layer_fn, Layer};
+use tower_service::Service;
+
+/// A boxed `Layer` trait object.
```
hawkw, line 6: nit: should `Layer` be a link?
```suggestion
/// A boxed [`Layer`] trait object.
```

tower/src/util/boxed/layer.rs
```diff
@@ -0,0 +1,86 @@
...
+///
+/// [`BoxLayer`] turns a layer into a trait object, allowing the output service to be dynamic.
```
hawkw, line 8: not sure if `BoxLayer` needs to be a link, because this _is_ the docs for `BoxLayer`...I think this should be
```suggestion
/// `BoxLayer` turns a [`Layer`] into a trait object, allowing the output [`Service`] to be dynamic.
```

hawkw: can we maybe explain what "allowing the output service to be dynamic" actually means a little more clearly? This boxes the `Service` as well as the `Layer`, so that the layer can emit more than one kind of `Service`. Maybe something like
```suggestion
/// [`BoxLayer`] turns a layer into a trait object, allowing both the [`Layer`] itself
/// and the output [`Service`] to be dynamic, while having consistent types.
///
/// This [`Layer`] produces [`BoxService`] instances erasing the type of the 
/// [`Service`] produced by the wrapped [`Layer`].
```

tower/src/util/boxed/layer.rs
```diff
@@ -0,0 +1,86 @@
...
+///         let layer = builder
+///             .map_err(Into::into)
+///             .into_inner();
+///
+///         BoxLayer::new(layer)
+///     }
+/// }
+/// ```
+pub struct BoxLayer<In, T, U, E> {
+    boxed: Arc<dyn Layer<In, Service = BoxService<T, U, E>> + Send + Sync + 'static>,
```
hawkw, line 44: kinda ironic that this is called `BoxLayer` but it's actually `ArcLayer` :P

i think this is probably the best name for this, for consistency & because it produces `BoxService`es, but i just thought it was funny...

tower/src/util/boxed/layer.rs
```diff
@@ -0,0 +1,86 @@
...
+    fn clone(&self) -> Self {
+        Self {
+            boxed: Arc::clone(&self.boxed),
+        }
+    }
+}
+
+impl<In, T, U, E> fmt::Debug for BoxLayer<In, T, U, E> {
+    fn fmt(&self, fmt: &mut fmt::Formatter) -> fmt::Result {
+        fmt.debug_struct("BoxLayer").finish()
```
hawkw, line 84: here's a thought; not a blocker for this PR: something we _could_ do, and that we might want to think about doing for `BoxService` as well, is "remembering" the type name of the erased layer, like this:
```rust
pub struct BoxLayer<In, T, U, E> {
    boxed: Arc<dyn Layer<In, Service = BoxService<T, U, E>> + Send + Sync + 'static>,
    name: &'static str,
}

impl<In, T, U, E> BoxLayer<In, T, U, E> {
    /// Create a new [`BoxLayer`].
    pub fn new<L>(inner_layer: L) -> Self
    where
        L: Layer<In> + Send + Sync + 'static,
        L::Service: Service<T, Response = U, Error = E> + Send + 'static,
        <L::Service as Service<T>>::Future: Send + 'static,
    {
        let layer = /* ... */

        Self {
            boxed: Arc::new(layer),
            name: std::any::type_name::<L>(),
        }
    }
}

// ...

impl<In, T, U, E> fmt::Debug for BoxLayer<In, T, U, E> {
    fn fmt(&self, fmt: &mut fmt::Formatter) -> fmt::Result {
        fmt.debug_struct("BoxLayer")
            .field("boxed", &format_args!("Box<{}>", self.name))
            .finish()
     }
}
```

This would let us see what the original, un-erased type was in the `Debug` impl, which could be useful for debugging in some cases.

The main downside to this is that it makes the one-word `BoxLayer` struct suddenly three words long (a `&'static str` is a ptr+len, so two words). This might be meaningful if you're cloning a really large number of `BoxLayer`s. 

Alternatively, we could stick the type name inside the `Arc`, so it's not part of each clone of the struct...but then we'd need some
```rust
struct Inner<In, T, U, E> {
    boxed: Box<dyn Layer<In, Service = BoxService<T, U, E>> + Send + Sync + 'static>,
    name: &'static str,
}
```
which is _also_ not great — now we're dereferencing _two_ heap pointers to call `self.inner.layer`, rather than one.

A third idea is to do the first proposal, but stick `#[cfg(debug_assertions)]` on the `name` field and the code that uses it. That way, type names are included in debug mode, but we remove the extra two words from `BoxLayer` in release mode for performance. IMO, that might be the best choice if we decide to do something like this.

Anyway, just an idea...probably out of scope for this PR and best done in a follow up, if we decide this is worth it.

tower/src/util/boxed/mod.rs
```diff
@@ -32,8 +32,9 @@
 //! [`Service`]: crate::Service
 //! [`Rc`]: std::rc::Rc
 
+mod layer;
 mod sync;
 mod unsync;
 
 #[allow(unreachable_pub)] // https://github.com/rust-lang/rust/issues/57411
-pub use self::{sync::BoxService, unsync::UnsyncBoxService};
+pub use self::{layer::BoxLayer, sync::BoxService, unsync::UnsyncBoxService};
```
hawkw, line 40: do we think there's much of a use-case for an `UnsyncBoxLayer` as well? I think what we _might_ want is a `Layer` that is `Sync` (and still `Arc`ed), but that produces `UnsyncBoxService`s instead of `BoxService`s. 

but, i guess that's just something we can easily add later if it turns out anyone needs it...

tower/src/util/boxed/layer.rs
```diff
@@ -0,0 +1,86 @@
...
+use tower_layer::{layer_fn, Layer};
+use tower_service::Service;
+
+/// A boxed `Layer` trait object.
+///
+/// [`BoxLayer`] turns a layer into a trait object, allowing the output service to be dynamic.
+///
+/// # Example
+///
+/// `BoxLayer` can for example be useful to create layers dynamically that otherwise wouldn't have
```
hawkw, line 12:
```suggestion
/// `BoxLayer` can, for example, be useful to create layers dynamically that otherwise wouldn't have
```

tower/src/util/boxed/layer.rs
```diff
@@ -0,0 +1,86 @@
...
+/// the same types.
```
hawkw, line 13: Maybe explain what the example is doing, here? Like
```suggestion
/// the same types. In this example, we include a [`Timeout`] layer
/// only if an environment variable is set. We can use `BoxLayer`
/// to return a consistent type regardless of runtime configuration:
```

tower/src/util/boxed/layer.rs
```diff
@@ -0,0 +1,86 @@
...
+///         BoxLayer::new(layer)
+///     } else {
+///         let layer = builder
+///             .map_err(Into::into)
+///             .into_inner();
+///
+///         BoxLayer::new(layer)
+///     }
+/// }
+/// ```
```
hawkw, line 42: if you take all my docs suggestions, we'll need to add some new link references:
```suggestion
/// ```
///
/// [`Layer`]: tower_layer::Layer
/// [`Service`]: tower_service::Service
/// [`BoxService`]: super::BoxService
/// [`Timeout`]: crate::timeout
```

hawkw (approved):
The new code looks great pretty much as-is. I had a bunch of documentation suggestions, and one idea for some future work.

hawkw:
Went ahead and made the remaining docs tweaks, gonna merge this pending CI :+1: 

We can look into improving debug output in a follow-up...or, it may not be worth the effort...

PR #572, opened 2021-02-25
chore: debug netlify redirects

davidbarsky:


davidpdrsn (approved):
No idea how this works but go right ahead 😅

PR #574, opened 2021-03-31
Allow reusable concurrency limit via GlobalConcurrencyLimit

jeromegn:
We've been needing this to share the same `Arc<Semaphore>` from multiple listeners.

Concretely, this allows us to have a shared "max connections" limit throughout all our listeners. It also allows us to reuse the same semaphore for all services for the same "app".

I've added a `ToOwnedSemaphore` trait and implemented it for `usize` and `Arc<Semaphore>`.

This is a breaking change for people explicitly creating a `ConcurrencyLimitLayer`. It now requires a type implementing `ToOwnedSemaphore` to be specified.

### DIFF
--- a/tower/src/limit/concurrency/layer.rs
+++ b/tower/src/limit/concurrency/layer.rs
@@ -28,8 +28,10 @@ impl<S> Layer<S> for ConcurrencyLimitLayer {
 
 /// Enforces a limit on the concurrent number of requests the underlying
 /// service can handle.
-/// This variant accepts a owned semaphore (`Arc<Semaphore>`) which can
-/// be reused across services.
+/// Unlike [`ConcurrencyLimitLayer`], which enforces a per-service concurrency
+/// limit, this layer accepts a owned semaphore (`Arc<Semaphore>`) which can be
+/// shared across multiple services.
+/// Cloning it will not create a new semaphore.
 #[derive(Debug, Clone)]
 pub struct GlobalConcurrencyLimitLayer {
     semaphore: Arc<Semaphore>,
@@ -37,7 +39,12 @@ pub struct GlobalConcurrencyLimitLayer {
 
 impl GlobalConcurrencyLimitLayer {
     /// Create a new `GlobalConcurrencyLimitLayer`.
-    pub fn new(semaphore: Arc<Semaphore>) -> Self {
+    pub fn new(max: usize) -> Self {
+        Self::with_semaphore(Arc::new(Semaphore::new(max)))
+    }
+
+    /// Create a new `GlobalConcurrencyLimitLayer` from a `Arc<Semaphore>`
+    pub fn with_semaphore(semaphore: Arc<Semaphore>) -> Self {
         GlobalConcurrencyLimitLayer { semaphore }
     }
 }
@@ -46,6 +53,6 @@ impl<S> Layer<S> for GlobalConcurrencyLimitLayer {
     type Service = ConcurrencyLimit<S>;
 
     fn layer(&self, service: S) -> Self::Service {
-        ConcurrencyLimit::new_shared(service, self.semaphore.clone())
+        ConcurrencyLimit::with_semaphore(service, self.semaphore.clone())
     }
 }
--- a/tower/src/limit/concurrency/layer.rs
+++ b/tower/src/limit/concurrency/layer.rs
@@ -28,10 +28,12 @@ impl<S> Layer<S> for ConcurrencyLimitLayer {
 
 /// Enforces a limit on the concurrent number of requests the underlying
 /// service can handle.
+///
 /// Unlike [`ConcurrencyLimitLayer`], which enforces a per-service concurrency
 /// limit, this layer accepts a owned semaphore (`Arc<Semaphore>`) which can be
 /// shared across multiple services.
-/// Cloning it will not create a new semaphore.
+///
+/// Cloning this layer will not create a new semaphore.
 #[derive(Debug, Clone)]
 pub struct GlobalConcurrencyLimitLayer {
     semaphore: Arc<Semaphore>,
--- a/tower/src/limit/concurrency/layer.rs
+++ b/tower/src/limit/concurrency/layer.rs
@@ -1,4 +1,7 @@
+use std::sync::Arc;
+
 use super::ConcurrencyLimit;
+use tokio::sync::Semaphore;
 use tower_layer::Layer;
 
 /// Enforces a limit on the concurrent number of requests the underlying
@@ -20,5 +23,29 @@ impl<S> Layer<S> for ConcurrencyLimitLayer {
 
     fn layer(&self, service: S) -> Self::Service {
         ConcurrencyLimit::new(service, self.max)
+    }
+}
+
+/// Enforces a limit on the concurrent number of requests the underlying
+/// service can handle.
+/// This variant accepts a owned semaphore (`Arc<Semaphore>`) which can
+/// be reused across services.
+#[derive(Debug, Clone)]
+pub struct GlobalConcurrencyLimitLayer {
+    semaphore: Arc<Semaphore>,
+}
+
+impl GlobalConcurrencyLimitLayer {
+    /// Create a new global concurrency limit layer.
+    pub fn new(semaphore: Arc<Semaphore>) -> Self {
+        GlobalConcurrencyLimitLayer { semaphore }
+    }
+}
+
+impl<S> Layer<S> for GlobalConcurrencyLimitLayer {
+    type Service = ConcurrencyLimit<S>;
+
+    fn layer(&self, service: S) -> Self::Service {
+        ConcurrencyLimit::new_owned(service, self.semaphore.clone())
     }
 }
--- a/tower/src/limit/concurrency/layer.rs
+++ b/tower/src/limit/concurrency/layer.rs
@@ -36,7 +36,7 @@ pub struct GlobalConcurrencyLimitLayer {
 }
 
 impl GlobalConcurrencyLimitLayer {
-    /// Create a new global concurrency limit layer.
+    /// Create a new `GlobalConcurrencyLimitLayer`.
     pub fn new(semaphore: Arc<Semaphore>) -> Self {
         GlobalConcurrencyLimitLayer { semaphore }
     }
@@ -46,6 +46,6 @@ impl<S> Layer<S> for GlobalConcurrencyLimitLayer {
     type Service = ConcurrencyLimit<S>;
 
     fn layer(&self, service: S) -> Self::Service {
-        ConcurrencyLimit::new_owned(service, self.semaphore.clone())
+        ConcurrencyLimit::new_shared(service, self.semaphore.clone())
     }
 }
--- a/tower/src/limit/concurrency/service.rs
+++ b/tower/src/limit/concurrency/service.rs
@@ -26,7 +26,7 @@ pub struct ConcurrencyLimit<T> {
 impl<T> ConcurrencyLimit<T> {
     /// Create a new concurrency limiter.
     pub fn new(inner: T, max: usize) -> Self {
-        Self::new_shared(inner, Arc::new(Semaphore::new(max)))
+        Self::with_semaphore(inner, Arc::new(Semaphore::new(max)))
     }
 
     /// Create a new concurrency limiter with a provided shared semaphore
--- a/tower/src/limit/concurrency/service.rs
+++ b/tower/src/limit/concurrency/service.rs
@@ -26,11 +26,11 @@ pub struct ConcurrencyLimit<T> {
 impl<T> ConcurrencyLimit<T> {
     /// Create a new concurrency limiter.
     pub fn new(inner: T, max: usize) -> Self {
-        Self::new_owned(inner, Arc::new(Semaphore::new(max)))
+        Self::new_shared(inner, Arc::new(Semaphore::new(max)))
     }
 
-    /// Create a new concurrency limiter with a provided owned semaphore
-    pub fn new_owned(inner: T, semaphore: Arc<Semaphore>) -> Self {
+    /// Create a new concurrency limiter with a provided shared semaphore
+    pub fn new_shared(inner: T, semaphore: Arc<Semaphore>) -> Self {
         ConcurrencyLimit {
             inner,
             semaphore: PollSemaphore::new(semaphore),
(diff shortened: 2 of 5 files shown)


davidpdrsn (commented):
I think this looks like a nice addition that makes sense to have. The ability to have a global concurrency limit used across services seems generally useful.

I'm thinking it might be possible to add without a breaking change however. Could we make the `T` on `ConcurrencyLimitLayer<T>` have a default type that would match the current behavior and keep `new` as `fn new(max: usize) -> ConcurrencyLimitLayer`?

We could then add `fn new_with_semaphore(semaphore: T) -> ConcurrencyLimitLayer<T>` for users who need the new feature.

If we did that we could have to keep `ServiceBuilder::concurrency_limit` as it is but doing `.layer(ConcurrencyLimitLayer::new_with_semaphore(semaphore))` isn't too bad.

jeromegn:
ping @davidpdrsn

tower/src/limit/concurrency/layer.rs
```diff
@@ -1,24 +1,67 @@
...
-        ConcurrencyLimitLayer { max }
+        ConcurrencyLimitLayer { semaphore: max }
+    }
+}
+
+impl<T> ConcurrencyLimitLayer<T>
+where
+    T: ToOwnedSemaphore,
+{
+    /// Create a new concurrency limit layer from a type that fulfills `ToOwnedSemaphore`
```
davidpdrsn, line 25:
```suggestion
    /// Create a new concurrency limit layer from a type that implements [`ToOwnedSemaphore`].
```

tower/src/limit/concurrency/layer.rs
```diff
@@ -1,24 +1,67 @@
...
+{
     type Service = ConcurrencyLimit<S>;
 
     fn layer(&self, service: S) -> Self::Service {
-        ConcurrencyLimit::new(service, self.max)
+        ConcurrencyLimit::new(service, &self.semaphore)
+    }
+}
+
+/// Produces a "owned" Semaphore (`Arc<Semaphore>`)
```
davidpdrsn, line 42:
```suggestion
/// Produces an "owned" Semaphore (`Arc<Semaphore>`)
```

tower/src/limit/concurrency/layer.rs
```diff
@@ -1,24 +1,67 @@
...
+pub trait ToOwnedSemaphore {
+    /// Returns a owned Semaphore for the type
```
davidpdrsn, line 44:
```suggestion
    /// Returns aa owned [`Semaphore`] for the type.
```

davidpdrsn (approved):
A few small docs suggestions but otherwise I think it looks good!

I'm gonna ask in Discord for someone else to have a look.

tower/src/limit/concurrency/layer.rs
```diff
@@ -1,24 +1,67 @@
...
+{
     type Service = ConcurrencyLimit<S>;
 
     fn layer(&self, service: S) -> Self::Service {
-        ConcurrencyLimit::new(service, self.max)
+        ConcurrencyLimit::new(service, &self.semaphore)
+    }
+}
+
+/// Produces an owned [`Semaphore`] (`Arc<Semaphore>`)
```
hawkw, line 42: I think if we're going to have this trait be part of the API, it may be worth discussing what it does, and how users would use it to create a global concurrency limit across all instances of a stack?

tower/src/limit/concurrency/layer.rs
```diff
@@ -1,24 +1,67 @@
...
+    }
+}
+
+impl<T> ConcurrencyLimitLayer<T>
+where
+    T: ToOwnedSemaphore,
+{
+    /// Create a new concurrency limit layer from a type that implements [`ToOwnedSemaphore`]
+    pub fn new_with_semaphore(semaphore: T) -> Self {
+        ConcurrencyLimitLayer { semaphore }
```
hawkw, line 27: Not totally convinced it's necessary for this to be totally generic...we could just have a `ConcurrencyLimit::new(usize)` and a `ConcurrencyLimit::new_global(semaphore: Arc<Semaphore>)` or something like that? This would make it clearer to the user that one constructor is used for creating a separate concurrency limit for each service produced by the layer, and the other shares one limit across all such services produced by the layer.

I'm not sure if I can immediately come up with any use-cases for users implementing `ToOwnedSemaphore` for types other than `usize` or `Arc<Semaphore>`...but if there is a reason to, we can always add a more generic constructor later?

tower/src/limit/concurrency/layer.rs
```diff
@@ -1,24 +1,67 @@
...
     fn layer(&self, service: S) -> Self::Service {
-        ConcurrencyLimit::new(service, self.max)
+        ConcurrencyLimit::new(service, &self.semaphore)
+    }
+}
+
+/// Produces an owned [`Semaphore`] (`Arc<Semaphore>`)
+pub trait ToOwnedSemaphore {
+    /// Returns an owned [`Semaphore`] for the type
+    fn to_owned_semaphore(&self) -> Arc<Semaphore>;
```
hawkw, line 45: AFAICT, this trait will basically only ever be implemented for `usize` and `Arc<Semaphore>`, right? It doesn't seem like there's a lot that users could do with implementations for other types? 

Hypothetically, I could maybe see a situation where the concurrency limiting policy depends on the service being wrapped in some way, as a use-case for custom implementations, but the current trait doesn't make that possible.  Perhaps it could take a reference to the inner service when producing the semaphore...but I'm still not sure if something like that would actually be useful.

I would generally try to avoid exposing a trait in the public API if users aren't going to be implementing themselves it or using it as a trait bound, if that's possible. It feels like the user-facing complexity could be avoided if we just had two separate layers.

hawkw (commented):
> Concretely, this allows us to have a shared "max connections" limit throughout all our listeners. It also allows us to reuse the same semaphore for all services for the same "app".

I am definitely in favor of having something like this in `tower` --- we have [our own implementation of a global concurrency limit in `linkerd2-proxy`](https://github.com/linkerd/linkerd2-proxy/blob/07e9249769474f19e11904f9beed29a01dda7fe5/linkerd/concurrency-limit/src/lib.rs), so it would be nice to be able to depend on this from upstream.

I'm somewhat unconvinced about the API changes in this PR. It doesn't seem immediately obvious to the user what the `ToOwnedSemaphore` trait is for or how it would be used to create per-service or global shared concurrency limit middleware. As an alternative, what do we think about just adding a totally separate `GlobalConcurrencyLimit` or `SharedConcurrencyLimit` middleware? This would probably require a lot more code in `tower` than the current approach, but I think it could be easier to understand for users, and would make the distinction between a per-instance concurrency limit and a shared concurrency limit more obvious. What do you think?

jeromegn:
Makes sense!

Perhaps what was in my first commit was closer to mergeable? https://github.com/tower-rs/tower/pull/574/commits/99a7ae384a480c566cf90aef920a991042db51e2

davidpdrsn:
@jeromegn yeah seems a bit closer to what @hawkw is suggesting.

I would probably keep `ConcurrencyLimit` completely as is and make a separate new middleware called `GlobalConcurrencyLimit`.

jeromegn:
I've rewritten this based on my original commit. I think this strikes the right complexity and code-reuse. Shouldn't change the original API at all and allows creating a `ConcurrencyLimit` with a owned semaphore directly or through the new `GlobalConcurrencyLimitLayer`.

tower/src/limit/concurrency/service.rs
```diff
@@ -26,9 +26,14 @@ pub struct ConcurrencyLimit<T> {
 impl<T> ConcurrencyLimit<T> {
     /// Create a new concurrency limiter.
     pub fn new(inner: T, max: usize) -> Self {
+        Self::new_owned(inner, Arc::new(Semaphore::new(max)))
+    }
+
+    /// Create a new concurrency limiter with a provided owned semaphore
+    pub fn new_owned(inner: T, semaphore: Arc<Semaphore>) -> Self {
```
davidpdrsn, line 33:
```suggestion
    /// Create a new concurrency limiter with a provided shared semaphore
    pub fn new_shared(inner: T, semaphore: Arc<Semaphore>) -> Self {
```

I guess technically the `Arc` owns the semaphore 😜

tower/src/limit/concurrency/layer.rs
```diff
@@ -22,3 +25,27 @@ impl<S> Layer<S> for ConcurrencyLimitLayer {
...
+/// service can handle.
+/// This variant accepts a owned semaphore (`Arc<Semaphore>`) which can
+/// be reused across services.
+#[derive(Debug, Clone)]
+pub struct GlobalConcurrencyLimitLayer {
+    semaphore: Arc<Semaphore>,
+}
+
+impl GlobalConcurrencyLimitLayer {
+    /// Create a new global concurrency limit layer.
```
davidpdrsn, line 39:
```suggestion
    /// Create a new `GlobalConcurrencyLimitLayer`.
```

tower/src/limit/concurrency/layer.rs
```diff
@@ -22,3 +25,27 @@ impl<S> Layer<S> for ConcurrencyLimitLayer {
         ConcurrencyLimit::new(service, self.max)
     }
 }
+
+/// Enforces a limit on the concurrent number of requests the underlying
+/// service can handle.
+/// This variant accepts a owned semaphore (`Arc<Semaphore>`) which can
+/// be reused across services.
```
davidpdrsn, line 32: The word "variant" makes a think of an enum but I don't have any better suggestions 🤷

hawkw: how about just
```suggestion
/// Unlike [`ConcurrencyLimitLayer`], which enforces a per-service concurrency
/// limit, this layer accepts a owned semaphore (`Arc<Semaphore>`) which can be
/// shared across multiple services.
```
or something?

tower/src/limit/concurrency/layer.rs
```diff
@@ -22,3 +25,27 @@ impl<S> Layer<S> for ConcurrencyLimitLayer {
...
+#[derive(Debug, Clone)]
+pub struct GlobalConcurrencyLimitLayer {
+    semaphore: Arc<Semaphore>,
+}
+
+impl GlobalConcurrencyLimitLayer {
+    /// Create a new `GlobalConcurrencyLimitLayer`.
+    pub fn new(semaphore: Arc<Semaphore>) -> Self {
+        GlobalConcurrencyLimitLayer { semaphore }
+    }
```
hawkw, line 42: I'm not *totally* sure why this has to take an `Arc<Semaphore>`...couldn't it alternatively just have a constructor that takes the maximum number of concurrent requests, like `ConcurrencyLimitLayer`? Then, instead of cloning the semaphore and constructing a new `ConcurrencyLimitLayer` for each service, you could just clone the concurrency limit layer.

The advantage of this is that it doesn't put the `tokio::sync` type in the public API of `GlobalConcurrencyLimitLayer`. I feel like ideally, the semaphore should be just an implementation detail, not part of the public API. I can't immediately think of a reason we would ever want to change this code not to use `tokio::sync::Semaphore`, but it feels a little better not to expose it publicly, IMO...

tower/src/limit/concurrency/service.rs
```diff
@@ -26,9 +26,14 @@ pub struct ConcurrencyLimit<T> {
 impl<T> ConcurrencyLimit<T> {
     /// Create a new concurrency limiter.
     pub fn new(inner: T, max: usize) -> Self {
+        Self::new_shared(inner, Arc::new(Semaphore::new(max)))
+    }
+
+    /// Create a new concurrency limiter with a provided shared semaphore
+    pub fn new_shared(inner: T, semaphore: Arc<Semaphore>) -> Self {
```
hawkw, line 33: nitpicky, TIOLI: if we're going to have this API, i might prefer naming it something like:
```suggestion
    pub fn with_semaphore(inner: T, semaphore: Arc<Semaphore>) -> Self {
```

hawkw: on one hand, like my suggestion above, not having `tokio::sync`'s semaphore in the public API might be theoretically nicer, hiding the implementation details. on the other hand, this API is the only way we can share a semaphore across`ConcurrencyLimit` services _without_ using the `Layer`, so maybe having this constructor is worth having the semaphore exposed publicly.

if we decide to keep this public, and i think maybe we should, it's fine for the `GlobalConcurrencyLimitLayer` constructor to take a semaphore as it does currently...

hawkw (commented):
thanks, this feels a lot more straightforward to use than the previous implementation with `ToOwnedSemaphore`!

i had some thoughts on whether or not we should expose `tokio::sync::Semaphore` in tower's public API. on one hand, it's probably better to hide the implementation detail, but on the other, if we didn't expose it in constructors, a global concurrency limit could only be shared between services using the `Layer`...i'm open to being convinced either way on this.

i also left a few docs suggestions.

jeromegn:
I think we might have to keep exposing that `Semaphore`, even though it would be cleaner not to.

I don't actually use the `ConcurrencyLimitLayer::with_semaphore` publicly though, but since you're leaning towards keeping it public then it does make sense to allow building a `GlobalConcurrencyLimit` with a `Semaphore`

I made adjustments!

tower/src/limit/concurrency/layer.rs
```diff
@@ -22,3 +25,34 @@ impl<S> Layer<S> for ConcurrencyLimitLayer {
         ConcurrencyLimit::new(service, self.max)
     }
 }
+
+/// Enforces a limit on the concurrent number of requests the underlying
+/// service can handle.
+/// Unlike [`ConcurrencyLimitLayer`], which enforces a per-service concurrency
```
hawkw, line 31: nit:
```suggestion
/// service can handle.
///
/// Unlike [`ConcurrencyLimitLayer`], which enforces a per-service concurrency
```

tower/src/limit/concurrency/layer.rs
```diff
@@ -22,3 +25,34 @@ impl<S> Layer<S> for ConcurrencyLimitLayer {
...
+/// limit, this layer accepts a owned semaphore (`Arc<Semaphore>`) which can be
+/// shared across multiple services.
+/// Cloning it will not create a new semaphore.
```
hawkw, line 34: nit:
```suggestion
/// shared across multiple services.
///
/// Cloning this layer will not create a new semaphore.
```

hawkw (approved):
looks good to me, i commented on a couple tiny docs nits

davidpdrsn:
@jeromegn thanks for working on this! Do you wanna update the [changelog](https://github.com/tower-rs/tower/blob/master/tower/CHANGELOG.md) as well? Then we can merge it.

jeromegn:
@davidpdrsn done!

davidpdrsn (approved):

PR #581, opened 2021-04-26
spawn-ready: Abort background tasks when the service is dropped

olix0r:
The `SpawnReady` service spawns a background task whenever the inner
service is not ready; but when the `SpawnReady` service is dropped, this
task continues to run until it becomes ready (at which point the ready
service will be dropped). This can cause resource leaks when the inner
service never becomes ready.

This change adds a `Drop` implementation for the `SpawnReady` service
that aborts the background task when appropriate.


davidpdrsn (approved):

tower/src/spawn_ready/service.rs
```diff
@@ -73,3 +81,28 @@ where
...
+        }
+    });
+    tokio_test::assert_pending!(task.poll());
+    tokio_test::assert_pending!(handle.poll_request());
+
+    // End the task and ensure that the inner service has been dropped.
+    assert!(drop_tx.send(()).is_ok());
+    tokio_test::assert_ready!(task.poll());
+    assert!(tokio_test::assert_ready!(handle.poll_request()).is_none());
+}
```
hawkw, line 108: Nit/TIOLI: there are integration tests for `spawn_ready` [here](https://github.com/tower-rs/tower/blob/master/tower/tests/spawn_ready/main.rs), would it make more sense to put the test in that file?

hawkw (approved):
lgtm, modulo one note about where the test ought to go.

PR #583, opened 2021-04-28
builder: Add `ServiceBuilder::map_result`

davidpdrsn:
Noticed that `ServiceBuilder` didn't have `map_result`, only `then` which is async.


hawkw (approved):
nice catch, lgtm!

PR #584, opened 2021-04-30
improve flexiblity of retry policy

rcoh:
`retry::Policy` is an effective way to expressing retries, however, there two use cases that
as it stands today, cannot be expressed:
- Altering the final response (eg. to record the fact that you ran out of retries)
- Altering the request (eg. to set a header to the server indicating that this request is a retry)

(Technically the second is possible with `clone_request`, but it's a little unclear _which_ request would actually get sent).

This change implements what I think is pretty close to the minimal update to make this possible, namely, `req`
and `Res` both become mutable references. This enables policies to mutate them during execution & enables both of the
use cases above without complicating the "simple path" callers who don't need this behavior.

**This is a breaking change.** However, the fixes are only a couple of `&mut` and potentially a call to `as_ref()`.

I've also considered (and would be totally happy) with an API that "mapped" over req & res, taking and returning ownership. This may actually even be slightly more flexible. However, I didn't want to leak this detail onto all users.


tower/src/retry/policy.rs
```diff
@@ -58,7 +58,7 @@ pub trait Policy<Req, Res, E>: Sized {
     ///
     /// [`Service::Response`]: crate::Service::Response
     /// [`Service::Error`]: crate::Service::Error
-    fn retry(&self, req: &Req, result: Result<&Res, &E>) -> Option<Self::Future>;
+    fn retry(&self, req: &mut Req, result: &mut Result<Res, E>) -> Option<Self::Future>;
```
davidpdrsn, line 61: I guess getting a `&mut Result<Res, E>` allows users to change `Err(_)` into `Ok(_)` which might be kinda odd. Not sure. Feels like getting `Result<&mut Res, &mut E>` would fit the use-cases as well but not allow changing the `Result` variant.

davidpdrsn (commented):
I have never used this middleware in practice but to me the change looks sensible. Though I'm wondering if someone else with more experience using `Retry` has run into other problems that could be solved, should we decide to make a breaking change. cc @hawkw @olix0r @LucioFranco @seanmonstar 

Had one small suggestion/question.

> I've also considered (and would be totally happy) with an API that "mapped" over req & res, taking and returning ownership. This may actually even be slightly more flexible. However, I didn't want to leak this detail onto all users.

To return a new request/response the future returned by `Policy::retry` would have to change  as well which might be too much of a change.

rcoh:
The entire point is to change the variant (at least for me). Because there
are "success" cases the actually need to be retries and when I run out of
retries I need to change it to failure

davidpdrsn:
Alright cool. That does make sense. 

I kinda feel like if we are going to make a breaking change we should take care to design something that we are happy with. That might be this or it might be bigger changes. I haven't used Retry personally so I don't know which it is. There was some discussion about other potential changes to Retry [here](https://github.com/tower-rs/tower/pull/414).

tower/src/retry/policy.rs
rcoh: comment wasn't threaded because of email:
> The entire point is to change the variant (at least for me). Because there
are "success" cases the actually need to be retries and when I run out of
retries I need to change it to failure

LucioFranco:
@rcoh maybe I am forgetting how this works but can't you return an error in the returned service future?

hawkw:
i actually _just_ ran into a use-case for a mutable request in `clone_request`. i was trying to implement a retry policy where an HTTP request body is buffered lazily as the first request's body is sent, rather than aggregating the whole body prior to sending the initial request.

the approach i was going to take was storing the buffered body on the initial request, and then sending the buffer on a `oneshot` channel when the first request is dropped so that the potential retry can take ownership of it. however, there isn't currently any way to get the receiver side of the oneshot _out_ of the initial request, without introducing a mutex or something for this purpose. making the request mutable in `clone_request` would very make this kind of thing much easier, so i think we should definitely make this change!

rcoh:
@hawkw just to clarify, in this change the request becomes mutable in the `retry` method; I thought it was a little unclear from a user perspective _which_ request was which if you were to mutate it in `clone_request()` (basically, is the cloned request sent first or is the original sent first?)

Which is why I made it mutatable in `retry` because then it seemed more obvious that the request you just mutated would get sent next. I _think_ that should be equivalent with your try_clone use case, but we could also make it mutable in both cases?

hawkw:
> @hawkw just to clarify, in this change the request becomes mutable in the `retry` method; I thought it was a little unclear from a user perspective _which_ request was which if you were to mutate it in `clone_request()` (basically, is the cloned request sent first or is the original sent first?)
> 
> Which is why I made it mutatable in `retry` because then it seemed more obvious that the request you just mutated would get sent next. I _think_ that should be equivalent with your try_clone use case, but we could also make it mutable in both cases?

oh, i misread the PR description, whoops! i'm going to just go ahead and try to implement a lazily buffered request body against this branch and report back? i think it may still be possible without _also_ touching `clone_request`, but we'll see...

another option would be to make the `Policy` itself mutable in both `clone_request()` and `retry()`, which would also give you an avenue for sharing state between the first and second request...

hawkw:
It would be really nice to get this change into 0.5 (which we're currently working on). @rcoh, are you interested in updating this branch to build with the latest `master`, or should I?

rcoh:
I can do the merge tomorrow assuming there's nothing too problematic

LucioFranco (approved):

rcoh:
merge is done—I assume you'll squash merge, but if not let me know and I can cleanup history

tower/src/retry/policy.rs
```diff
@@ -56,6 +56,21 @@ pub trait Policy<Req, Res, E>: Sized {
...
+    /// tracking the retry count in a header.
+    ///
+    /// ## Mutating Results
+    ///
+    /// The policy MAY chose to mutate the result. This can enable the retry policy to convert a failure
+    /// into a success and vice versa. For example, if the policy is used to poll while waiting for a state
+    /// change, you can switch the result to failure such that running out of retries returns a specific failure.
+    ///
+    /// You can also record metadata on the request to include information about the number of retries required
+    /// or to record that a failure failed after exhausting all retries.
```
hawkw, line 72: This is kind of a minor nitpick, but can we wrap these lines at the same column as the other lines in this doc comment? Thanks.

tower/src/retry/policy.rs
```diff
@@ -56,6 +56,21 @@ pub trait Policy<Req, Res, E>: Sized {
...
     ///
+    /// ## Mutating Requests
+    ///
+    /// The policy MAY chose to mutate the `req`: if the request is mutated, the mutated request
+    /// will be sent to the inner service in the next retry. This can be helpful for use cases like
+    /// tracking the retry count in a header.
+    ///
+    /// ## Mutating Results
+    ///
+    /// The policy MAY chose to mutate the result. This can enable the retry policy to convert a failure
```
hawkw, line 67: nit: how about
```suggestion
    /// The policy MAY chose to mutate the result. This enables the retry policy to convert a failure
```

tower/src/retry/policy.rs
```diff
@@ -56,6 +56,21 @@ pub trait Policy<Req, Res, E>: Sized {
...
+    /// into a success and vice versa. For example, if the policy is used to poll while waiting for a state
+    /// change, you can switch the result to failure such that running out of retries returns a specific failure.
```
hawkw, line 69: nit: in general, we prefer to avoid second-person ('you') language in docs:
```suggestion
    /// change, the policy can switch the result to failure such that running out of retries returns a specific failure.
```

tower/src/retry/policy.rs
hawkw, line 72: nit: maybe
```suggestion
    /// Policies can also record metadata on the request. This can include information about the number of retries required,
    /// or to record that a failure failed after exhausting all retries.
```

hawkw (commented):
some minor documentation suggestions, hope they're helpful!

hawkw:
> I assume you'll squash merge, but if not let me know and I can cleanup history

Yeah, that's right, we generally merge PRs by squashing, so you don't have to worry about it! :)

Issue #33, opened 2017-12-11
Document the library


carllerche:
This is a meta issue tracking concepts that should be included as part of the documentation.

* Graceful shutdown (#32).

davidbarsky:
Regarding “Implement Service for `Arc<MyType>` to get around `&mut` requirement“, could this be done using arbitrary self types, or am I misunderstanding something?

davidpdrsn:
Improving the general documentation around Tower is something I would like to drive. I have seen several cases on Twitter and Reddit where people say that Tower seems cool but they don't really understand it. I also remember thinking the same thing when I first came to Tower.

I'm very open to suggestions about how to do this but my current plan is:

1. Make sure all middlewares have examples in the docs. Trying to figure out how to use something from the types alone [can be hard](https://github.com/tower-rs/tower/issues/566).
2. Add examples of how to put all the pieces together and build an HTTP server and client with hyper. Our [current number of examples](https://github.com/tower-rs/tower/tree/master/examples) isn't impressive 😅
3. Write a guide similar to [Tonic's helloworld guide](https://github.com/hyperium/tonic/blob/master/examples/helloworld-tutorial.md) that explains how to build a middleware from scratch.
4. Link to the [Tower deep dive stream](https://www.youtube.com/watch?v=16sU1q8OeeI&t=3052s) I did some months ago. Several people have mentioned this helped them understand the concepts better.

I don't imagine we can get all this done before Tower 0.5 so don't think it should block that release.

PR #585, opened 2021-05-08
docs: Add "Inventing the `Service` trait" guide

davidpdrsn:
Actually didn't expect to spend all day writing this, but here we are 😅

[Rendered](https://github.com/tower-rs/tower/blob/david/inventing-service-guide/guides/inventing-service.md)

This adds the first Tower guide called "Inventing the `Service` trait". It attempts to motivate all the parts to `Service` by walking the user through how they could have invented `Service` themselves, from scratch. It goes into quite a bit of detail but hopefully it paints a somewhat complete picture in the end.

The next guide I want to write is about how to implement a proper `Timeout` middleware using `Layer`, pin-project, and all the bells and whistles.

Ref: https://github.com/tower-rs/tower/issues/33

### DIFF
--- a/guides/README.md
+++ b/guides/README.md
@@ -6,9 +6,9 @@ recommended. If you're brand new to async Rust, we recommend the [Asynchronous
 Programming in Rust][async-book] book or the [Tokio tutorial][tokio-tutorial].
 
 Additionally, some of these guides explain Tower from the perspective of HTTP
-servers and clients. However Tower is useful for anything that follows and async
-request/response pattern. Not just web services even though that is its primary
-use case.
+servers and clients. However, Tower is useful for any network protocol that
+follows an async request/response pattern. HTTP is used here because it is a
+widely known protocol, and one of Tower's more common use-cases.
 
 ## Guides
 
--- a/guides/README.md
+++ b/guides/README.md
@@ -1,0 +1,21 @@
+# Tower guides
+
+These guides are meant to be an introduction to Tower. At least basic Rust
+experience is assumed. Some experience with asynchronous Rust is also
+recommended. If you're brand new to async Rust we recommend the [Asynchronous
+Programming in Rust][async-book] book.
+
+Additionally, some of these guides explains Tower from the perspective of HTTP
+servers and clients. However Tower is useful for anything that follows and async
+request/response pattern. Not just web services even though that is its primary
+use case.
+
+## Guides
+
+- ["Inventing the `Service` trait"][invent] goes through how you could have
+  invented Tower's fundamental `Service` trait starting from scratch. If you
+  have no experience with Tower and want to learn the absolute basics this is
+  where you should start.
+
+[async-book]: https://rust-lang.github.io/async-book/
+[invent]: https://github.com/tower-rs/tower/blob/master/guides/inventing-service.md
--- a/guides/README.md
+++ b/guides/README.md
@@ -3,7 +3,7 @@
 These guides are meant to be an introduction to Tower. At least basic Rust
 experience is assumed. Some experience with asynchronous Rust is also
 recommended. If you're brand new to async Rust we recommend the [Asynchronous
-Programming in Rust][async-book] book.
+Programming in Rust][async-book] book or the [Tokio tutorial][tokio-tutorial].
 
 Additionally, some of these guides explain Tower from the perspective of HTTP
 servers and clients. However Tower is useful for anything that follows and async
@@ -18,4 +18,5 @@ use case.
   where you should start.
 
 [async-book]: https://rust-lang.github.io/async-book/
+[tokio-tutorial]: https://tokio.rs/tokio/tutorial
 [invent]: https://github.com/tower-rs/tower/blob/master/guides/inventing-service.md
--- a/guides/README.md
+++ b/guides/README.md
@@ -5,7 +5,7 @@ experience is assumed. Some experience with asynchronous Rust is also
 recommended. If you're brand new to async Rust we recommend the [Asynchronous
 Programming in Rust][async-book] book.
 
-Additionally, some of these guides explains Tower from the perspective of HTTP
+Additionally, some of these guides explain Tower from the perspective of HTTP
 servers and clients. However Tower is useful for anything that follows and async
 request/response pattern. Not just web services even though that is its primary
 use case.
--- a/guides/README.md
+++ b/guides/README.md
@@ -1,8 +1,8 @@
-# Tower guides
+# Tower Guides
 
 These guides are meant to be an introduction to Tower. At least basic Rust
 experience is assumed. Some experience with asynchronous Rust is also
-recommended. If you're brand new to async Rust we recommend the [Asynchronous
+recommended. If you're brand new to async Rust, we recommend the [Asynchronous
 Programming in Rust][async-book] book or the [Tokio tutorial][tokio-tutorial].
 
 Additionally, some of these guides explain Tower from the perspective of HTTP
@@ -12,9 +12,9 @@ use case.
 
 ## Guides
 
-- ["Inventing the `Service` trait"][invent] goes through how you could have
-  invented Tower's fundamental `Service` trait starting from scratch. If you
-  have no experience with Tower and want to learn the absolute basics this is
+- ["Inventing the `Service` trait"][invent] walks through how you Tower's 
+  fundamental `Service` trait could be designed from scratch. If you
+  have no experience with Tower and want to learn the absolute basics, this is
   where you should start.
 
 [async-book]: https://rust-lang.github.io/async-book/
--- a/guides/inventing-service.md
+++ b/guides/inventing-service.md
@@ -1,10 +1,10 @@
 # Inventing the `Service` trait
 
-At the core of Tower is the `Service` trait. It models an asynchronous function
-that takes a request and produces a response. However, some aspects of its design
-may not be immediately obvious. Rather than starting with the `Service trait as it
-exists in Tower today, let's look at the motivation behind `Service` by imagining
-how you might invent it if you started from scratch.
+At the core of Tower is the [`Service`] trait. A `Service` is an asynchronous
+function that takes a request and produces a response. However, some aspects of
+its design may not be immediately obvious. Rather than starting with the
+`Service` trait as it exists in Tower today, let's look at the motivation behind
+`Service` by imagining how you might invent it if you started from scratch.
 
 Imagine you are building a little HTTP framework in Rust. The framework would
 allow users to implement an HTTP server by supplying code that receives requests
@@ -566,12 +566,13 @@ library on crates.io for other people to use!
 
 Our little `Handler` trait is working quite nicely, but currently it only
 supports our `Request` and `Response` types. It would be nice if those where
-generic so users could use whatever type they want.
+generic, so users could use whatever type they want.
 
 We make `Request` a generic type parameter of the trait so that a given service
-can accept many different types of requests. We make `Response` an associated type
-because for any _given_ `Request` type, there can only be one (associated) `Response`
-type: the one the corresponding call returns!
+can accept many different types of requests. That allows to define handlers that
+can be use for different protocols. We make `Response` an associated type
+because for any _given_ `Request` type, there can only be one (associated)
+`Response` type: the one the corresponding call returns!
 
 ```rust
 trait Handler<Request> {
@@ -898,7 +899,7 @@ pub trait Service<Request> {
 ```
 
 Many middleware don't add their own backpressure, and simply delegate to the
-wrapped service's `poll_ready_. However, backpressure in middleware does enable
+wrapped service's `poll_read_`. However, backpressure in middleware does enable
 some interesting use cases, such as various kinds of rate limiting, load balancing,
 and auto scaling.
 
@@ -938,6 +939,7 @@ a feature called [generic associated types][gat] which will let us get around th
 Generic associated types will allow us to define the response future as
 `type Future<'a>`, and `call` as `fn call<'a>(&'a mut self, ...) -> Self::Future<'a>`.
 
+[`Service`]: https://docs.rs/tower/latest/tower/trait.Service.html
 [async-trait]: https://crates.io/crates/async-trait
 [dropping]: https://doc.rust-lang.org/stable/std/ops/trait.Drop.html
 [`futures::future::poll_fn`]: https://docs.rs/futures/0.3.14/futures/future/fn.poll_fn.html
--- a/guides/inventing-service.md
+++ b/guides/inventing-service.md
@@ -276,7 +276,7 @@ We still have to require that `Handler::Future` implements `Future` where the
 output type is `Result<Response, Error>` as that is what `Server::run` requires.
 
 Having `call` take `&mut self` is useful because it allows handlers to update
-their internal state if necessary.
+their internal state if necessary<sup>[1](#pin)</sup>.
 
 Lets convert our original `handle_request` function into an implementation of
 this trait:
@@ -374,7 +374,7 @@ block. That means the lifetime of our future is tied to the lifetime of `&mut
 self`. This doesn't work for us since we might want to run our response futures
 on multiple threads to get better performance or produce multiple response
 futures and run them all in parallel. That isn't possible if a reference to the
-handler lives inside the futures<sup>[1](#gats)</sup>.
+handler lives inside the futures<sup>[2](#gats)</sup>.
 
 Instead we need convert the `&mut self` into an owned `self`. That is exactly
 what `Clone` does:
@@ -897,7 +897,12 @@ let response = service
 
 ## Footnotes
 
-<a name="gats">1</a>: To be a bit more precise, the reason this requires the
+<a name="pin">1</a>: There has been some discussion around whether `call` should
+take `Pin<&mut Self>` or not but so far we've decided to go with a plain `&mut
+self` which means handlers (ahem services) must be `Unpin`. In practice that is
+rarely an issue. More details [here](https://github.com/tower-rs/tower/issues/319).
+
+<a name="gats">2</a>: To be a bit more precise, the reason this requires the
 response future to be `'static` is that writing `Box<dyn Future>` actually
 becomes `Box<dyn Future + 'static>` which the anonymous lifetime in `fn call(&'_
 mut self, ...)` doesn't satisfy. In the future we're hoping to use [Generic
--- a/guides/inventing-service.md
+++ b/guides/inventing-service.md
@@ -373,7 +373,7 @@ block. That means the lifetime of our future is tied to the lifetime of `&mut
 self`. This doesn't work for us since we might want to run our response futures
 on multiple threads to get better performance or produce multiple response
 futures and run them all in parallel. That isn't possible if a reference to the
-handler lives inside the futures.
+handler lives inside the futures<sup>[1](#gats)</sup>.
 
 Instead we need convert the `&mut self` into an owned `self`. That is exactly
 what `Clone` does:
@@ -545,19 +545,19 @@ library on crates.io for other people to use!
 
 Our little `Handler` trait is working quite nicely but currently it only
 supports our `Request` and `Response` types. It would be nice if those where
-generic so users could use whatever type they want:
+generic so users could use whatever type they want.
+
+We make `Request` a generic type parameter of the trait so that a given service
+can accept many different types of requests. We make `Response` an associated type
+because for any given `Request` type, there can only be one (associated) `Response`
+type: the one the corresponding call returns!
 
 ```rust
-// Lets make the request type a generic type parameter.
-// This allows users to implement `Handler` multiple times for
-// the same type. Its not something we need but our users might.
 trait Handler<Request> {
-    // Similarly to `Future` we add an associated type for the
-    // response
     type Response;
 
-    // Might as well add one for error as well. No reason for that
-    // to be a hardcoded type
+    // Error should also be an associated type. No reason for that to be a
+    // hardcoded type
     type Error;
 
     // Our future type from before but now its output must use
@@ -565,7 +565,7 @@ trait Handler<Request> {
     type Future: Future<Output = Result<Self::Response, Self::Error>>;
 
     // `call` is unchanged but note that `Request` here is our generic
-    // `Request` type paramter and not the `Request` type we've used
+    // `Request` type parameter and not the `Request` type we've used
     // until now
     fn call(&mut self, request: Request) -> Self::Future;
 }
@@ -772,12 +772,11 @@ impl<R, T> Service<R> for ConcurrencyLimit<T> {
 ```
 
 If there is no capacity left we have to wait and somehow get notified when
-capacity becomes available. So if the service is receiving a lot of requests we
-risk spending a lot of time waiting. It would be nice if we had a way to tell
-the caller that there was no more capacity left. Maybe they wanted to react by
-calling another service or somehow creating a new service to handle part of the
-load. We also risk running out of memory if we keep accepting requests faster
-than we can process them.
+capacity becomes available. Additionally we have to keep the request in memory
+while we're waiting (also called buffering). It would be robust to only allocate
+space for the request when we are sure the service has capacity to handle it.
+Otherwise we risk using a lot of memory buffering requests while we wait for our
+service to become ready.
 
 In a way it would be nice if `Service` had a method like this:
 
@@ -819,21 +818,30 @@ trait Service<R> {
 ```
 
 This means if the service is at capacity `poll_ready` will return
-`Poll::Pending` and somehow ensure to be woken up using `Context` when capacity
-becomes available. At that point `poll_ready` can be called again and if it
-returns `Poll::Ready(())` then capacity would be reserved and `call` can be
-called.
+`Poll::Pending` and somehow notify the caller using the waker from the `Context`
+when capacity becomes available. At that point `poll_ready` can be called again
+and if it returns `Poll::Ready(())` then capacity would be reserved and `call`
+can be called.
+
+`poll_ready` not returning a `Future` also means we're able to quickly check if
+a service is ready without being forced to wait for it to become ready. If we
+call `poll_ready` and get back `Poll::Pending` we can simply decide to do
+something else instead of waiting. Among other things, this allows you to build
+load balancers that estimates the load of services by how often they return
+`Poll::Pending` and sends requests to the service with the least load.
 
 It would still be possible get a `Future` that resolves when capacity is
 available using something like [`futures::future::poll_fn`] (or
 [`tower::ServiceExt::ready`]).
 
 This concept of services communicating with their callers about their capacity
-is called "backpressure propagation". The fundamental idea is that you shouldn't
-send a request to a service that doesn't have the capacity to handle it. Instead
-you should wait, call another service, or somehow scale up the number of
-services available. You can learn more about the general concept of backpressure
-[here][backpressure].
+is called "backpressure propagation". Think of it as services pushing back on
+their callers telling them to slow down if they're going too fast. The
+fundamental idea is that you shouldn't send a request to a service that doesn't
+have the capacity to handle it. Instead you should wait (buffering), drop the
+request (load shedding), or handle the lack of capacity in some other way. You
+can learn more about the general concept of backpressure [here][backpressure]
+and [here][backpressure2].
 
 Finally, it might also be possible for some error to happen while reserving
 capacity so `poll_ready` probably should return `Poll<Result<(), Self::Error>>`.
@@ -854,7 +862,7 @@ pub trait Service<Request> {
 }
 ```
 
-Middlewares that care about backpressure is pretty rare but it does enable some
+Middlewares that care about backpressure are pretty rare but it does enable some
 interesting use cases such as various kinds of rate limiting, load balancing,
 and auto scaling.
 
@@ -877,8 +885,22 @@ let response = service
     .call(request).await?;
 ```
 
+---
+
+## Footnotes
+
+<a name="gats">1</a>: To be a bit more precise, the reason this requires the
+response future to be `'static` is that writing `Box<dyn Future>` actually
+becomes `Box<dyn Future + 'static>` which the anonymous lifetime in
+`fn call(&'_ mut self, ...)` doesn't satisfy. In the future we're hoping to
+use [Generic associated types][gat] to expression this a bit more cleanly.
+That will allow us to define the response future as `type Future<'a>` and `call`
+as `fn call<'a>(&'a mut self, ...) -> Self::Future<'a>`.
+
 [async-trait]: https://crates.io/crates/async-trait
 [dropping]: https://doc.rust-lang.org/stable/std/ops/trait.Drop.html
 [`futures::future::poll_fn`]: https://docs.rs/futures/0.3.14/futures/future/fn.poll_fn.html
 [`tower::ServiceExt::ready`]: https://docs.rs/tower/0.4.7/tower/trait.ServiceExt.html#method.ready
 [backpressure]: https://medium.com/@jayphelps/backpressure-explained-the-flow-of-data-through-software-2350b3e77ce7
+[backpressure2]: https://aws.amazon.com/builders-library/using-load-shedding-to-avoid-overload/
+[gat]: https://github.com/rust-lang/rust/issues/44265
--- a/guides/inventing-service.md
+++ b/guides/inventing-service.md
@@ -150,7 +150,7 @@ duration `handler` is allowed to take. If it doesn't produce a response within
 that amount of time an error is returned.
 
 Your first idea might be to modify `Server` to support being configured with a
-timeout which it will somehow apply to call of the handler. However it turns out
+timeout which it will somehow apply to the call of the handler. However it turns out
 you can actually add a timeout without modifying `Server`. Using
 `tokio::time::timeout` we can make a new handler function that calls our
 previous `handle_request` but with a timeout of 30 seconds:
@@ -188,7 +188,7 @@ async fn handler_with_timout_and_content_type(
 }
 ```
 
-We now have handler that will process an HTTP request, take no longer than 30
+We now have a handler that will process an HTTP request, take no longer than 30
 seconds, and always have the right `Content-Type` header all without modifying
 our original `handle_request` function or `Server` struct.
 
@@ -805,7 +805,7 @@ avaiable using something like [`futures::future::poll_fn`] (or
 This concept of services communicating with their callers about their capacity
 is called "backpressure propagation". The fundamental idea is that you shouldn't
 send a request to a service that doesn't have the capacity to handle it. Instead
-you should wait, call another service, or somehow scale up the number of service
+you should wait, call another service, or somehow scale up the number of services
 available.
 
 Finally, it might also be possible for some error to happen while reserving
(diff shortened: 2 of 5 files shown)


guides/README.md
```diff
@@ -0,0 +1,21 @@
+# Tower guides
+
+These guides are meant to be an introduction to Tower. At least basic Rust
+experience is assumed. Some experience with asynchronous Rust is also
+recommended. If you're brand new to async Rust we recommend the [Asynchronous
+Programming in Rust][async-book] book.
```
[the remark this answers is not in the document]
davidpdrsn: Good point! Fixed in 6e4872f30ceb2

guides/inventing-service.md
```diff
@@ -0,0 +1,856 @@
...
+This provides a quite nice separation of concerns. We were able to add a timeout
+without changing any of our existing code.
+
+Lets add one more feature in this way. Imagine we're building a JSON API and
+would therefore like a `Content-Type: application/json` header on all responses.
+We can wrap `handler_with_timeout` in a similar way and modify the response like
+so:
+
+```rust
+async fn handler_with_timout_and_content_type(
```
ibraheemdev, line 182:
```suggestion
async fn handler_with_timeout_and_content_type(
```

guides/inventing-service.md
```diff
@@ -0,0 +1,856 @@
...
+    request: Request,
+) -> Result<Response, Error> {
+    let mut response = handler_with_timout(request).await?;
```
ibraheemdev, line 185:
```suggestion
    let mut response = handler_with_timeout(request).await?;
```

guides/inventing-service.md
```diff
@@ -0,0 +1,856 @@
...
+Designing libraries that can be extended in this way is very powerful since it
+allows users to extend things without having to wait for the library maintainers
+to add support for it.
+
+However there is one problem. Our current setup doesn't scale very well. Imagine
+we have many `handle_with_*` functions that each add a little bit of behavior.
+Having to hard code the chain of which intermediate handler calls which isn't
+great. Our current chain is
+
+1. `handler_with_timout_and_content_type` which calls
```
ibraheemdev, line 204:
```suggestion
1. `handler_with_timeout_and_content_type` which calls
```

guides/inventing-service.md
```diff
@@ -0,0 +1,856 @@
...
+2. `handler_with_timout` which calls
```
ibraheemdev, line 205:
```suggestion
2. `handler_with_timeout` which calls
```

guides/inventing-service.md
```diff
@@ -0,0 +1,856 @@
...
+that amount of time an error is returned.
+
+Your first idea might be to modify `Server` to support being configured with a
+timeout which it will somehow apply to the call of the handler. However it turns out
+you can actually add a timeout without modifying `Server`. Using
+`tokio::time::timeout` we can make a new handler function that calls our
+previous `handle_request` but with a timeout of 30 seconds:
+
+```rust
+async fn handler_with_timout(request: Request) -> Result<Response, Error> {
```
ibraheemdev, line 159:
```suggestion
async fn handler_with_timeout(request: Request) -> Result<Response, Error> {
```

ibraheemdev (commented):
Looks really nice! A couple spelling mistakes I noticed while skimming it:

guides/inventing-service.md
```diff
@@ -0,0 +1,856 @@
...
+        //    again.
+    }
+}
+```
+
+If there is no capacity left we just have to wait a bit and then check again. So
+if the service is receiving a lot of requests you can imagine waiting for a long
+time.  It would also be nice if we had a way to tell the caller that there was
+no more capacity left. Maybe they wanted to react to that by calling another
+service instead.
```
davidpdrsn, line 759: I'm not quite sure this argument is strong enough. Rather than "waiting for some random amount of time" one could await on an async semaphore. However that merges "driving a service readiness" and "calling the service" so `poll_ready` is still better. I'm just not quite sure this section drives that point home.

guides/inventing-service.md
```diff
@@ -0,0 +1,856 @@
...
+
+Lets go with option 2 as its the most flexible. Users who have a concrete future
+type can use that without the cost of a `Box` and users who don't care can still
+use `Pin<Box<...>>`.
+
+```rust
+trait Handler {
+    type Future: Future<Output = Result<Response, Error>>;
+
+    fn call(&mut self, request: Request) -> Self::Future;
```
davidpdrsn, line 256: I would also like to add a paragraph about why `call` taking `&mut self` is necessary. As I don't think thats entirely obvious.

guides/inventing-service.md
```diff
@@ -0,0 +1,856 @@
...
+
+While still being able to run our handler like before:
+
+```rust
+server.run(final_handler).await?;
+```
+
+You could imagine `with_content_type` and `with_timeout` being functions that
+took an argument of type `F: Fn(Request) -> Future<Output = Response>` and
+returned a closure a la `impl Fn(Request) -> Future<Output = Result<Response,
```
Folyd, line 223: Hey, does the word `la` is a typo?

guides/inventing-service.md
```diff
@@ -0,0 +1,856 @@
...
+trait Service<R> {
+    async fn ready(&mut self);
+}
+```
+
+`ready` would be an async function that completes when the service has capacity
+enough to receive one new request. We would then require users to first call
+`service.ready().await` before doing `service.call(request).await`.
+
+This way way `ConcurrencyLimit` could track capacity inside `ready` and not
```
Folyd, line 773: Duplicated word `way`?

guides/inventing-service.md
```diff
@@ -0,0 +1,856 @@
...
+
+Types like `Timeout` and `JsonContentType` are typically called "middlewares"
+since they wrap another `Service` and transform the request or response in some
+way. Whereas types like `RequestHandler` is typically called "leaf services"
+since it sits at the leaves of a tree of nested services. The actual responses
+are normally produced in leaf services and modified by middlewares.
+
+The only thing left to talk about is "backpressure" and `poll_ready`.
+
+## Backpressure
```
Folyd, line 732: I'd recommend adding a `backpressure` explanation article here. Such as this wonder one: https://medium.com/@jayphelps/backpressure-explained-the-flow-of-data-through-software-2350b3e77ce7.

This would help a lot to those new to `backpressure`. What do you think?

guides/inventing-service.md
davidpdrsn: Nope:

> Borrowed from French, à la means "according to" or "in the manner of,"

https://www.dictionary.com/e/slang/a-la/

I have changed it however to "like" as that is probably more understandable.

guides/inventing-service.md
davidpdrsn: Yes! Good catch.

guides/inventing-service.md
davidpdrsn: Good idea! I added a link in 7461ae492ddaa

guides/inventing-service.md
davidpdrsn: Hopefully 8283ecaec7446 makes things a bit more clear.

guides/inventing-service.md
```diff
@@ -0,0 +1,866 @@
...
+155 | |             }
+156 | |         })
+    | |_________^ ...is captured here, requiring it to live as long as `'static`
+```
+
+The issue is that we're capturing a `&mut self` and moving it into an async
+block. That means the lifetime of our future is tied to the lifetime of `&mut
+self`. This doesn't work for us since we might want to run our response futures
+on multiple threads to get better performance or produce multiple response
+futures and run them all in parallel. That isn't possible if a reference to the
```
LucioFranco, line 361: This isn't actually the reason for this error, basically, TLDR we don't have GAT. All associated types must return a `'static` reference and since `'static` > `'a`  from `&mut 'a self` it doesn't like it. Async-trait gets around this above as I mentioned in that comment, plus the fact that it is a trait object so a concrete type vs a generic like we have here.

guides/inventing-service.md
```diff
@@ -0,0 +1,866 @@
...
+handler lives inside the futures.
+
+Instead we need convert the `&mut self` into an owned `self`. That is exactly
```
LucioFranco, line 364: another option is because the inner `Handle::Future` _has_ to be `'static` you can:

```
let fut = self.inner.call(...);
```

and that `fut` has to be a `'static` future.

LucioFranco: The clone method also works though :D

guides/inventing-service.md
```diff
@@ -0,0 +1,866 @@
...
+152 | |                 this.inner_handler.call(request),
+...   |
+159 | |             }
+160 | |         })
+    | |__________^ ...so that the type `impl Future` will meet its required lifetime bounds
+```
+
+The problem now is that `T` can be any type whatsoever. It can even be a type
+that contains references like `Vec<&'a str>`. However that wont work for the
+same reason as before. We need the response future to have a `'static` lifetime
```
LucioFranco, line 432: This is because again our associated type is `Pin<Box<dyn Future + 'static>>` (if you don't include a lifetime its auto a `'static`. And again the same problem as above.

guides/inventing-service.md
```diff
@@ -0,0 +1,866 @@
...
+
+Our little `Handler` trait is working quite nicely but currently it only
+supports our `Request` and `Response` types. It would be nice if those where
+generic so users could use whatever type they want:
+
+```rust
+// Lets make the request type a generic type parameter.
+// This allows users to implement `Handler` multiple times for
+// the same type. Its not something we need but our users might.
+trait Handler<Request> {
```
LucioFranco, line 537: Might want to add a note why the `Request` is generic but `Response` is an associated type. This is because generally for traits if its a param for method the generic is much easier while a return type associated types work much better with how you compose things.

guides/inventing-service.md
```diff
@@ -0,0 +1,866 @@
...
+
+If we put all those things together we get
+
+```rust
+// `Timeout` accepts any request of type `R` as long as `T`
+// accepts the same type of request
+impl<R, T> Handler<R> for Timeout<T>
+where
+    // The actual type of request must not contain
+    // references. The compiler would tell us to add
```
LucioFranco, line 587: bless Dr compiler 🙌

guides/inventing-service.md
```diff
@@ -0,0 +1,866 @@
...
+similar to what we've just did.
+
+Some example services provided by Tower:
+
+- `Timeout` - This is pretty much identical to the timeout we have built.
+- `Retry` - To automatically retry failed requests.
+- `RateLimit` - To limit the number of requests a service can handle over a
+  period of time.
+
+Types like `Timeout` and `JsonContentType` are typically called "middlewares"
```
LucioFranco, line 727:
```suggestion
Types like `Timeout` and `JsonContentType` are typically called "middleware"
```

I don't think this needs to be plural when its in quotes like this?

guides/inventing-service.md
```diff
@@ -0,0 +1,866 @@
...
+since they wrap another `Service` and transform the request or response in some
+way. Whereas types like `RequestHandler` is typically called "leaf services"
+since it sits at the leaves of a tree of nested services. The actual responses
+are normally produced in leaf services and modified by middlewares.
```
LucioFranco, line 731: I might mention something about how they are usually the things interacting with io. We use the term "leaf" futures for the futures that interact direction with `TcpStream`, etc.

guides/inventing-service.md
```diff
@@ -0,0 +1,866 @@
...
+
+The only thing left to talk about is "backpressure" and `poll_ready`.
+
+## Backpressure
+
+Imagine you wanted to write a rate limit middleware that wraps a `Service` and
+puts a limit on the maximum number of concurrent requests the underlying service
+can will receive. This would be useful if you had some service that had a hard
```
LucioFranco, line 739: can will? lol

guides/inventing-service.md
```diff
@@ -0,0 +1,866 @@
...
+    }
+}
+```
+
+If there is no capacity left we have to wait and somehow get notified when
+capacity becomes available. So if the service is receiving a lot of requests we
+risk spending a lot of time waiting. It would be nice if we had a way to tell
+the caller that there was no more capacity left. Maybe they wanted to react to
+that by calling another service or somehow creating another service to handle
+part of the load.
```
LucioFranco, line 762: or worst case, we keep accepting requests and that results in unbounded memory growth.

guides/inventing-service.md
```diff
@@ -0,0 +1,866 @@
...
+since they wrap another `Service` and transform the request or response in some
+way. Whereas types like `RequestHandler` is typically called "leaf services"
+since it sits at the leaves of a tree of nested services. The actual responses
+are normally produced in leaf services and modified by middlewares.
+
+The only thing left to talk about is "backpressure" and `poll_ready`.
+
+## Backpressure
+
+Imagine you wanted to write a rate limit middleware that wraps a `Service` and
```
LucioFranco, line 737: Might wanna link https://aws.amazon.com/builders-library/using-load-shedding-to-avoid-overload/ and there are a few other articles in there that are REALLY good and connect the real world with these concepts. Maybe for later in this section.

guides/inventing-service.md
```diff
@@ -0,0 +1,866 @@
...
+```
+
+`ready` would be an async function that completes when the service has capacity
+enough to receive one new request. We would then require users to first call
+`service.ready().await` before doing `service.call(request).await`.
+
+Separating "calling the service" from "reserving capacity" also unlocks new use
+cases such as being able to maintain a set of "ready services" that we keep up
+to date in the background such that when a request arrives we already have a
+ready service to send it to and don't have to first wait for it to become ready.
```
LucioFranco, line 779: or if its required to do a lot of work to say allocate the request you only need to do it when you can ensure it will get processed. (Think like reading a big image off the wire etc)

guides/inventing-service.md
```diff
@@ -0,0 +1,866 @@
...
+
+`Service`s that don't care about capacity can just return immediately from
+`ready` or delegate to some inner `Service` they're wrapping.
+
+However we still cannot define async functions inside traits. We could add
+another associated type to `Service` called `ReadyFuture` but having to return a
+`Future` will give us the same lifetime issues we've run into before. It would
+be nice if there was someway around that.
+
+Instead we can take some inspiration from the `Future` trait and define a method
```
LucioFranco, line 792: actually technically the inspiration came from `Sink` since it originally had a `start_send` but I think @carllerche wrote that too?

guides/inventing-service.md
```diff
@@ -0,0 +1,866 @@
...
+    fn poll_ready(
+        &mut self,
+        cx: &mut Context<'_>,
+    ) -> Poll<Result<(), Self::Error>>;
+
+    fn call(&mut self, req: Request) -> Self::Future;
+}
+```
+
+Middlewares that care about backpressure is pretty rare but it does enable some
```
LucioFranco, line 839: buffering is the other big one! This is what like TCP does with its kernel buffer.

LucioFranco (commented):
This is fantastic!! Ive had a lot of trouble trying to figure out how to write about tower but I think this does a really good job!

guides/inventing-service.md
davidpdrsn: 🙇‍♂️

guides/inventing-service.md
```diff
@@ -0,0 +1,866 @@
...
+}
+```
+
+We now have a handler that will process an HTTP request, take no longer than 30
+seconds, and always have the right `Content-Type` header all without modifying
+our original `handle_request` function or `Server` struct.
+
+Designing libraries that can be extended in this way is very powerful since it
+allows users to extend things without having to wait for the library maintainers
+to add support for it.
```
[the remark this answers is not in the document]
davidpdrsn: I added a paragraph just below this one.

guides/inventing-service.md
davidpdrsn: Yes thats totally correct. Again I'm not sure how much lifetime trouble to actually go into. I'm a bit worried that if we say `Pin<Box<dyn Future>>` needs a `+ 'a` from `&'a mut self` then some users might wonder "well easy, just define service as `trait Service<'a>`". For example this does compile:

```rust
use std::future::Future;

trait Service<'a, Request> {
    type Response;
    type Error;
    type Future: Future<Output = Result<Self::Response, Self::Error>> + 'a;

    fn call(&'a mut self, request: Request) -> Self::Future;
}

struct HttpRequest;
struct Error;

async fn call_all<S>(mut svc: S, requests: Vec<HttpRequest>) -> Result<(), Error>
where
    for<'a> S: Service<'a, HttpRequest, Error = Error>,
{
    for request in requests {
        svc.call(request).await?;
    }
    
    Ok(())
}
```

But we can't `tokio::spawn` the response as that requires the future to be `'static` which then makes `self` `&'static mut self` which clearly is quite limiting.

So yeah you're totally correct but I'm trying to say something that will make users think "yeah that kinda makes sense" without opening a can of worms 😅

What do you think?

guides/inventing-service.md
davidpdrsn: I can will good English.

guides/inventing-service.md
davidpdrsn: Good point 👍 Added a note about that.

guides/inventing-service.md
davidpdrsn: I'm not familiar with that. Would you mention that, and if so how would you phrase it?

guides/inventing-service.md
```diff
@@ -0,0 +1,866 @@
...
+}
+```
+
+Having a trait like this allows us to write concrete types that implement it so
+we don't have to deal with `Fn`s all the time.
+
+However Rust currently doesn't support async traits so we have two options:
+
+1. Make `call` return a boxed future like `Pin<Box<dyn Future<Output =
+   Result<Response, Error>>>`. This is what the [async-trait] crate does.
```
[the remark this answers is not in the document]
LucioFranco: Maybe its worth adding a footnote? I think for some users it could connect the dots better than others. Unfortunately with stuff like this you will encounter lifetimes (as you show later on) so maybe its worth it to bring it up?

guides/inventing-service.md
LucioFranco: I think you're right there is a balance to be achieved here 😄

guides/inventing-service.md
LucioFranco: Yeah, there are a few articles in that builder library that are great reads and connect some real world concepts. Load shedding tends to be the root behavior where you reject requests when the buffer is full and that rejection stops requests from coming in etc. What ever you feel is correct :)

guides/inventing-service.md
LucioFranco: Yeah totally fair, just thought I'd mention it xD

guides/inventing-service.md
jonhoo: I think it's worth leaving a note here saying that `clone` isn't the only way to do this, and that in the future we're hoping to be able to express this in an even cleaner way (link to GAT issue). That's probably enough for the casual reader.

guides/inventing-service.md
jonhoo: Here's how I'd phrase this:

> We make `Request` a generic type parameter of the trait so that a given service can accept many different types of requests. We make `Response` an associated type because for any _given_ `Request` type, there can only be one (associated) `Response` type: the one the corresponding `call` returns!

guides/inventing-service.md
```diff
@@ -0,0 +1,884 @@
...
+        // 2. If there is capacity left send the request to `T`
+        //    and increment the counter.
+        // 3. If not somehow wait until capacity becomes available.
+    }
+}
+```
+
+If there is no capacity left we have to wait and somehow get notified when
+capacity becomes available. So if the service is receiving a lot of requests we
+risk spending a lot of time waiting. It would be nice if we had a way to tell
```
jonhoo, line 776: "spending a lot of time waiting" struck me as a somewhat odd way of expressing the problem. After all, the future will just be asleep, so what's the issue? You later get onto what I think is the _real_ problem, namely that you have to already have the request in memory in order to invoke `call`, and therefore you can only exert backpressure after you've already accepted a request, which leads to buffering up a lot of requests internally in the server. I think it's probably worth hoisting that point up here directly.

jonhoo (approved):
I _love_ this — thanks for taking the time @davidpdrsn! I left some minor comments.

guides/inventing-service.md
davidpdrsn: Adding it as a footnote is a good idea!

guides/inventing-service.md
davidpdrsn: Love it!

guides/inventing-service.md
davidpdrsn: I actually thinking linking to both is fine 😊

guides/inventing-service.md
davidpdrsn: Thats a very good point! I'll fix that :)

LucioFranco (approved):
LGTM! The only one thing I might add is that we don't have `Pin<&mut Self>` in the `poll_ready` api and it might make sense to add a foot note that links to the issue where we talked about it.

guides/inventing-service.md
```diff
@@ -0,0 +1,919 @@
...
+
+## Making `Handler` more flexible
+
+Our little `Handler` trait is working quite nicely but currently it only
+supports our `Request` and `Response` types. It would be nice if those where
+generic so users could use whatever type they want.
+
+We make `Request` a generic type parameter of the trait so that a given service
+can accept many different types of requests. We make `Response` an associated type
+because for any given `Request` type, there can only be one (associated) `Response`
```
jonhoo, line 555:
```suggestion
because for any _given_ `Request` type, there can only be one (associated) `Response`
```

guides/inventing-service.md
```diff
@@ -0,0 +1,919 @@
...
+use std::task::{Context, Poll};
+
+trait Service<R> {
+    fn poll_ready(&mut self, cx: &mut Context<'_>) -> Poll<()>;
+}
+```
+
+This means if the service is at capacity `poll_ready` will return
+`Poll::Pending` and somehow notify the caller, using the waker from the
+`Context`, when capacity becomes available. At that point `poll_ready` can be
```
jonhoo, line 825:
```suggestion
`Poll::Pending` and notify the caller using the waker from the
`Context` when capacity becomes available. At that point `poll_ready` can be
```

guides/inventing-service.md
```diff
@@ -0,0 +1,919 @@
...
+called again and if it returns `Poll::Ready(())` then capacity would be reserved
+and `call` can be called.
+
+Note that there is technically nothing that prevents users from calling `call`
+without first making sure the service is ready. Doing so however is considered a
+violation of the `Service` contract and implementations are allowed to `panic`
+if `call` is called on a service that isn't ready!
```
jonhoo, line 832: Should we mention here that getting `Ready` means a slot is "reserved" for us, and that we should then use it or we're holding up capacity in the system? Plus maybe a footnote with a link to the `disarm` stuff?

jonhoo (approved):
I think this is great as-is, but left a few minor follow-ups.

guides/inventing-service.md
```diff
@@ -0,0 +1,919 @@
...
+- `Timeout` - This is pretty much identical to the timeout we have built.
+- `Retry` - To automatically retry failed requests.
+- `RateLimit` - To limit the number of requests a service can handle over a
+  period of time.
+
+Types like `Timeout` and `JsonContentType` are typically called "middleware"
+since they wrap another `Service` and transform the request or response in some
+way. Whereas types like `RequestHandler` is typically called a "leaf service"
+since it sits at the leaves of a tree of nested services. The actual responses
+are normally produced in leaf services and modified by middlewares.
```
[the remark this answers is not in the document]
hawkw: maybe also reference https://docs.rs/tower/0.4.7/tower/trait.Service.html#middleware--layer here?

guides/inventing-service.md
```diff
@@ -0,0 +1,919 @@
...
+The only thing left to talk about is "backpressure" and `poll_ready`.
+
+## Backpressure
+
+Imagine you wanted to write a rate limit middleware that wraps a `Service` and
+puts a limit on the maximum number of concurrent requests the underlying service
+will receive. This would be useful if you had some service that had a hard upper
+limit on the amount of load it could handle.
+
+With our current `Service` trait we don't really have a good way to implement
```
hawkw, line 761:
```suggestion
With our current `Service` trait, we don't really have a good way to implement
```

guides/inventing-service.md
```diff
@@ -0,0 +1,919 @@
...
+    }
+}
+```
+
+If there is no capacity left we have to wait and somehow get notified when
+capacity becomes available. Additionally we have to keep the request in memory
+while we're waiting (also called buffering). It would be more robust to only
+allocate space for the request when we are sure the service has capacity to
+handle it. Otherwise we risk using a lot of memory buffering requests while we
+wait for our service to become ready.
```
hawkw, line 782:
```suggestion
If there is no capacity left, we have to wait and somehow get notified when
capacity becomes available. Additionally, we have to keep the request in memory
while we're waiting (also called _buffering_). This means that the more requests
that are waiting for capacity, the more memory our program would use --- if more 
requests are produced than our service can handle, we might run out of memory! 
It would be more robust to only allocate space for the request when we are sure
the service has capacity to handle it. Otherwise, we risk using a lot of memory 
buffering requests while we wait for our service to become ready.
```

guides/inventing-service.md
```diff
@@ -0,0 +1,919 @@
...
+
+In a way it would be nice if `Service` had a method like this:
```
hawkw, line 784:
```suggestion
It would be nice if `Service` had a method like this:
```

guides/inventing-service.md
```diff
@@ -0,0 +1,919 @@
...
+    async fn ready(&mut self);
+}
+```
+
+`ready` would be an async function that completes when the service has capacity
+enough to receive one new request. We would then require users to first call
+`service.ready().await` before doing `service.call(request).await`.
+
+Separating "calling the service" from "reserving capacity" also unlocks new use
+cases such as being able to maintain a set of "ready services" that we keep up
```
hawkw, line 797:
```suggestion
cases, such as being able to maintain a set of "ready services" that we keep up
```

guides/inventing-service.md
```diff
@@ -0,0 +1,919 @@
...
+to date in the background such that when a request arrives we already have a
```
hawkw, line 798:
```suggestion
to date in the background, so that when a request arrives we already have a
```

guides/inventing-service.md
```diff
@@ -0,0 +1,919 @@
...
+ready service to send it to and don't have to first wait for it to become ready.
+
+With this setup `ConcurrencyLimit` could track capacity inside `ready` and not
```
hawkw, line 801:
```suggestion
With this design, `ConcurrencyLimit` could track capacity inside `ready` and not
```

guides/inventing-service.md
```diff
@@ -0,0 +1,919 @@
...
+allow users to call `call` until there is sufficient capacity.
+
+`Service`s that don't care about capacity can just return immediately from
+`ready` or delegate to some inner `Service` they're wrapping.
```
hawkw, line 805:
```suggestion
`ready`, or if they wrap some inner `Service`, they could delegate to its
`ready` method.
```

guides/inventing-service.md
```diff
@@ -0,0 +1,919 @@
...
+
+However we still cannot define async functions inside traits. We could add
+another associated type to `Service` called `ReadyFuture` but having to return a
+`Future` will give us the same lifetime issues we've run into before. It would
+be nice if there was someway around that.
```
hawkw, line 810:
```suggestion
However, we still cannot define async functions inside traits. We could add
another associated type to `Service` called `ReadyFuture`, but having to return a
`Future` will give us the same lifetime issues we've run into before. It would
be nice if there was some way around that.
```

guides/inventing-service.md
```diff
@@ -0,0 +1,919 @@
...
+
+Instead we can take some inspiration from the `Future` trait and define a method
```
hawkw, line 812:
```suggestion
Instead, we can take some inspiration from the `Future` trait and define a method
```

guides/inventing-service.md
```diff
@@ -0,0 +1,919 @@
...
+trait Service<R> {
+    fn poll_ready(&mut self, cx: &mut Context<'_>) -> Poll<()>;
+}
+```
+
+This means if the service is at capacity `poll_ready` will return
+`Poll::Pending` and notify the caller using the waker from the
+`Context` when capacity becomes available. At that point `poll_ready` can be
+called again and if it returns `Poll::Ready(())` then capacity would be reserved
+and `call` can be called.
```
hawkw, line 827:
```suggestion
This means if the service is at capacity, `poll_ready` will return
`Poll::Pending` and notify the caller using the waker from the
`Context` when capacity becomes available. At that point, `poll_ready` can be
called again, and if it returns `Poll::Ready(())` then capacity is  reserved and
`call` can be called.
```

guides/inventing-service.md
```diff
@@ -0,0 +1,919 @@
...
+
+Note that there is technically nothing that prevents users from calling `call`
+without first making sure the service is ready. Doing so however is considered a
```
hawkw, line 830:
```suggestion
without first making sure the service is ready. However, doing so is considered a
```

guides/inventing-service.md
```diff
@@ -0,0 +1,919 @@
...
+violation of the `Service` contract and implementations are allowed to `panic`
```
hawkw, line 831:
```suggestion
violation of the `Service` API contract, and implementations are allowed to `panic`
```

guides/inventing-service.md
```diff
@@ -0,0 +1,919 @@
...
+if `call` is called on a service that isn't ready!
+
+`poll_ready` not returning a `Future` also means we're able to quickly check if
+a service is ready without being forced to wait for it to become ready. If we
+call `poll_ready` and get back `Poll::Pending` we can simply decide to do
```
hawkw, line 836:
```suggestion
call `poll_ready` and get back `Poll::Pending`, we can simply decide to do
```

guides/inventing-service.md
```diff
@@ -0,0 +1,919 @@
...
+something else instead of waiting. Among other things, this allows you to build
+load balancers that estimates the load of services by how often they return
```
hawkw, line 838:
```suggestion
load balancers that estimate the load of services by how often they return
```

guides/inventing-service.md
```diff
@@ -0,0 +1,919 @@
...
+`Poll::Pending` and sends requests to the service with the least load.
```
hawkw, line 839:
```suggestion
`Poll::Pending`, and send requests to the service with the least load.
```

guides/inventing-service.md
```diff
@@ -0,0 +1,919 @@
...
+
+It would still be possible get a `Future` that resolves when capacity is
+available using something like [`futures::future::poll_fn`] (or
+[`tower::ServiceExt::ready`]).
+
+This concept of services communicating with their callers about their capacity
+is called "backpressure propagation". Think of it as services pushing back on
```
hawkw, line 846: maybe:
```suggestion
is called "backpressure propagation". You can think of it as services pushing back on
```

guides/inventing-service.md
```diff
@@ -0,0 +1,919 @@
...
+their callers telling them to slow down if they're going too fast. The
```
hawkw, line 847:
```suggestion
their callers, and telling them to slow down if they're producing requests too fast. The
```

guides/inventing-service.md
```diff
@@ -0,0 +1,919 @@
...
+fundamental idea is that you shouldn't send a request to a service that doesn't
+have the capacity to handle it. Instead you should wait (buffering), drop the
+request (load shedding), or handle the lack of capacity in some other way. You
+can learn more about the general concept of backpressure [here][backpressure]
+and [here][backpressure2].
+
+Finally, it might also be possible for some error to happen while reserving
+capacity so `poll_ready` probably should return `Poll<Result<(), Self::Error>>`.
```
hawkw, line 855:
```suggestion
capacity, so `poll_ready` probably should return `Poll<Result<(), Self::Error>>`.
```

guides/inventing-service.md
```diff
@@ -0,0 +1,919 @@
...
+        cx: &mut Context<'_>,
+    ) -> Poll<Result<(), Self::Error>>;
+
+    fn call(&mut self, req: Request) -> Self::Future;
+}
+```
+
+Middlewares that care about backpressure are pretty rare but it does enable some
+interesting use cases such as various kinds of rate limiting, load balancing,
+and auto scaling.
```
hawkw, line 875: maybe
```suggestion
Many middleware don't add their own backpressure, and simply delegate to the
wrapped service's `poll_ready_. However, backpressure in middleware does enable
some interesting use cases, such as various kinds of rate limiting, load balancing,
and auto scaling.
```

guides/inventing-service.md
```diff
@@ -0,0 +1,919 @@
...
+
+Since you never know exactly which middlewares a `Service` might consist of it
```
hawkw, line 877:
```suggestion
Since you never know exactly which middleware a `Service` might consist of, it
```

guides/inventing-service.md
```diff
@@ -0,0 +1,919 @@
...
+    // send the request
+    .call(request).await?;
+```
+
+---
+
+## Footnotes
+
+<a name="pin">1</a>: There has been some discussion around whether `call` should
+take `Pin<&mut Self>` or not but so far we've decided to go with a plain `&mut
```
hawkw, line 901:
```suggestion
take `Pin<&mut Self>` or not, but so far we've decided to go with a plain `&mut
```

guides/inventing-service.md
```diff
@@ -0,0 +1,919 @@
...
+self` which means handlers (ahem services) must be `Unpin`. In practice that is
```
hawkw, line 902:
```suggestion
self` which means handlers (ahem, _services_) must be `Unpin`. In practice that is
```

guides/inventing-service.md
```diff
@@ -0,0 +1,919 @@
...
+rarely an issue. More details [here](https://github.com/tower-rs/tower/issues/319).
+
+<a name="gats">2</a>: To be a bit more precise, the reason this requires the
+response future to be `'static` is that writing `Box<dyn Future>` actually
+becomes `Box<dyn Future + 'static>` which the anonymous lifetime in `fn call(&'_
```
hawkw, line 907:
```suggestion
becomes `Box<dyn Future + 'static>`, which the anonymous lifetime in `fn call(&'_
```

guides/inventing-service.md
```diff
@@ -0,0 +1,919 @@
...
+mut self, ...)` doesn't satisfy. In the future we're hoping to use [Generic
+associated types][gat] get around this. That will allow us to define the
+response future as `type Future<'a>` and `call` as `fn call<'a>(&'a mut self,
+...) -> Self::Future<'a>`.
```
hawkw, line 911:
```suggestion
mut self, ...)` doesn't satisfy. In the future, the Rust compiler team plans to
a feature called [generic associated types][gat] which will let us get around this.
Generic associated types will allow us to define the response future as
`type Future<'a>`, and `call` as `fn call<'a>(&'a mut self, ...) -> Self::Future<'a>`.
```

hawkw (approved):
this is a great way to introduce the `Service` trait's design, and i think the informal tone works well for this kind of tutorials.

i left a bunch of minor copyedits, mostly on grammar tweaks and fixing some awkward wording. hope that's helpful!

PR #586, opened 2021-05-14
Fix a couple of typos in the guide

davidpdrsn:


Folyd (changes requested):
Hey, I found another small typo: 
```
It would then apply that timeout every time it calls `handler.` However, it turns out
```

Line 171: the "`handler.`" should be "`handler`."

LucioFranco (approved):

PR #590, opened 2021-05-22
docs: add "Building a middleware from scratch" guide

davidpdrsn:
This adds a guide that explains how to implement a middleware from scratch without taking any shortcuts. It walks through implementing `Timeout` as it exists in Tower today.

The hope is that once users have read [the previous guide](https://tokio.rs/blog/2021-05-14-inventing-the-service-trait) followed by this one they should be fully equipped to implement their own middleware.

[🖼 Rendered](https://github.com/tower-rs/tower/blob/david/timeout-from-scratch-guide/guides/building-a-middleware-from-scratch.md)


guides/building-a-middleware-from-scratch.md
```diff
@@ -0,0 +1,693 @@
...
+```rust
+enum TimeoutError<Error> {
+    // Variant used if we hit the timeout
+    Timeout(InnerTimeoutError),
+    // Variant used if the inner service produced an error
+    Service(Error),
+}
+```
+
+While this seems ideal on the surface as we're not loosing any type information
```
memoryruins, line 400:
```suggestion
While this seems ideal on the surface as we're not losing any type information
```

guides/building-a-middleware-from-scratch.md
```diff
@@ -0,0 +1,693 @@
...
+So to make `ResponseFuture<F>` implement `Unpin` we need `F` and `Sleep` to
+implement `Unpin`. However `tokio::time::Sleep` is not `Unpin`. So even if we
+made `F` somehow implement `Unpin`, `ResponseFuture<F>` still wouldn't be
+`Unpin`.
+
+Its possible to make any type `Unpin` using `Box::pin`. But that requires
+allocating which is what we're trying to avoid. The proper solution is to use
+"pin projection".
+
+Pin projection sounds a lot more fancy than it really is. It basically means to
```
memoryruins, line 260:
```suggestion
Pin projection sounds a lot fancier than it really is. It basically means to
```

rrichardson:
The guide looks great. I think the topic is perfect.   "Merging" types can/should often only be done in the returned Future.  This was the last major stumbling block I ran into when learning Tower. 
 
I have one very minor question:  
Do we expect that readers will be using  `tower`  or `tower_service` ?  
 
 The previous guide did `use tower`  for its deps.

davidpdrsn:
> Do we expect that readers will be using tower or tower_service ?

Good point. I think depending on `tower` is fine and probably what users will do. I've updated it.

guides/building-a-middleware-from-scratch.md
```diff
@@ -0,0 +1,693 @@
...
+
+Writing a robust middleware requires working with async Rust at a slightly lower
+level than you might be used to. The goal of this guide is to demystify the
+concepts and patterns so you can start writing your own middleware and maybe
+even contribute back to the Tower ecosystem!
+
+## Getting started
+
+The middleware we're going to build is [`tower::timeout::Timeout`]. It will set
+a limit on the maximum duration its inner `Service` is allowed to take. If it
```
LucioFranco, line 17: I would maybe say it in this way, the maximum duration its inner service can dispatch for? Since its not the duration of the service but the response returned from `call`?

guides/building-a-middleware-from-scratch.md
```diff
@@ -0,0 +1,693 @@
...
+```rust
+use std::time::Duration;
+
+struct Timeout<T> {
+    inner: T,
+    timeout: Duration,
+}
+```
+
+As we learned in ["Inventing the `Service` trait"][invent] its important for
```
LucioFranco, line 34: Hmm Im not sure if this is important it is just one method. The key point is that there are sometimes things you need to own for the response, ie the duration of the timeout but I believe even in timeout you can create the timeout future then move that in. So why would the service need to be clone (besides that if it can impl clone it should to forward). So I guess my point is that I think this circumvents why clone is important for service middleware rather than using it for moving the data it should be there to forward (and sometimes you do need to move self but not always).

guides/building-a-middleware-from-scratch.md
```diff
@@ -0,0 +1,693 @@
...
+
+```rust
+impl<S> Timeout<S> {
+    pub fn new(inner: S, timeout: Duration) -> Self {
+        Timeout { inner, timeout }
+    }
+}
+```
+
+Note that we don't have to add any trait bounds on `S` even though we expect `S`
```
LucioFranco, line 58: Good call out! 👍

guides/building-a-middleware-from-scratch.md
```diff
@@ -0,0 +1,693 @@
...
+    // We don't have to clone `self.duration` as it implements the `Copy` trait.
+    let sleep = tokio::time::sleep(self.timeout);
+
+    // what to write here?
+}
+```
+
+One possible return type is `Pin<Box<dyn Future<...>>>`. However we want our
+`Timeout` to add as little overhead as possible, so we would like to find a way
+to avoid allocating a `Box`. Imagine we have a large stack, with dozens of
```
LucioFranco, line 121: Would be good to call out that `type_alias_impl_trait` fixes this issue since we can return an unnamed type.

guides/building-a-middleware-from-scratch.md
```diff
@@ -0,0 +1,693 @@
...
+
+What Rust is trying to tell us is that we need a `Pin<&mut F>` to be able to
+call `poll`. Accessing `F` through `self.response_future` when `self` is a
+`Pin<&mut Self>` doesn't work.
+
+One solution is to make `ResponseFuture<F>` implement [`Unpin`]. If a type
+implements `Unpin` it means we can ignore `Pin` and treat a `Pin<&mut Self>` as
+just a `&mut self`, as if the `Pin` wasn't there. As `Unpin` is an auto trait
+the compiler will automatically implement it if all types contained are also
+`Unpin`, similarly to `Send` and `Sync`.
```
LucioFranco, line 249: another interesting call out is that implementing `Unpin` manually is safe while doing `Send`/`Sync` is `unsafe impl`. This is because `Unpin` is safe to `impl` on a struct as long as you don't use any other unsafe code to access stuff. Its not unsafe but that is where the unsaftey comes up. So if you can impl `Unpin` on your struct and not use any `unsafe` to get any of the pins out then its totally safe. Its a bit different from Send/Sync in this case.

guides/building-a-middleware-from-scratch.md
```diff
@@ -0,0 +1,693 @@
...
+
+Which option you prefer is a matter of personal preference. Both have their
+advantages and disadvantages. However the pattern that we've decided to use in
+Tower is boxed trait objects. You can find the original discussion
+[here](https://github.com/tower-rs/tower/issues/131).
+
+For our `Timeout` middleware that means we need to create a struct that
+implements `std::error::Error` such that we can convert it into a `Box<dyn
+std::error::Error + Send + Sync>`. We also have to require that the inner
+service's error type implements `Into<Box<dyn std::error::Error + Send +
```
LucioFranco, line 440: And one note here is that we must always use `Into` if some middleware use `From` and other use `Into` it will break the stack in horrible to understand ways.

LucioFranco (approved):
LGTM! I left some inline questions/comments.

guides/building-a-middleware-from-scratch.md
```diff
@@ -0,0 +1,693 @@
...
+    inner: T,
+    timeout: Duration,
+}
+```
+
+As we learned in ["Inventing the `Service` trait"][invent] its important for
+services to implement `Clone` such that you can convert the `&mut self` given to
+`Service::call` into an owned `self` that can be moved into the response future,
+if necessary. We should therefore add `#[derive(Clone)]` to our struct. We
+should also drive `Debug` while we're at it:
```
jonhoo, line 38:
```suggestion
should also derive `Debug` while we're at it:
```

guides/building-a-middleware-from-scratch.md
```diff
@@ -0,0 +1,693 @@
...
+
+```rust
+#[derive(Debug, Clone)]
```
jonhoo, line 41: I think this is common enough in the ecosystem that I probably wouldn't even include text on why we're choosing to do it. See also [the Rust API guidelines' recommendation](https://rust-lang.github.io/api-guidelines/interoperability.html#types-eagerly-implement-common-traits-c-common-traits).

guides/building-a-middleware-from-scratch.md
jonhoo: I actually wonder if we should make this even briefer:
> Note that we omit bounds on `S` even though we expect it to implement `Service`, as the [Rust API guidelines recommend](https://rust-lang.github.io/api-guidelines/future-proofing.html#c-struct-bounds).

guides/building-a-middleware-from-scratch.md
```diff
@@ -0,0 +1,693 @@
...
+
+To avoid using `Box` lets instead write our own `Future` implementation. We
+start by creating a struct called `ResponseFuture`. It has to be generic over
+the inner service's response future type. This is analogous to wrapping services
+in other services, but this time we're wrapping futures in other futures.
+
+```rust
+use tokio::time::Sleep;
+
+pub struct ResponseFuture<F> {
```
jonhoo, line 136:
```suggestion
pub struct FutureWithTimeout<F> {
```

guides/building-a-middleware-from-scratch.md
```diff
@@ -0,0 +1,693 @@
...
+
+The real issue has to do with [`Pin`]. The full details of pinning is outside
+the scope of this guide. If you're brand new to `Pin` we recommend ["The Why,
+What, and How of Pinning in Rust"][pin] by Jon Gjengset.
+
+What Rust is trying to tell us is that we need a `Pin<&mut F>` to be able to
+call `poll`. Accessing `F` through `self.response_future` when `self` is a
+`Pin<&mut Self>` doesn't work.
+
+One solution is to make `ResponseFuture<F>` implement [`Unpin`]. If a type
```
jonhoo, line 245: This feels like a level of detail that isn't necessary for this article. I would jump straight to `pin_project` without first going through this strawman.

guides/building-a-middleware-from-scratch.md
```diff
@@ -0,0 +1,693 @@
...
+        f.pad("request timed out")
+    }
+}
+
+impl std::error::Error for TimeoutError {}
+```
+
+We add a private field to `TimeoutError` such that users outside of Tower cannot
+construct their own `TimeoutError`. They can only be obtained through our
+middleware.
```
jonhoo, line 463: I think the "modern" way of doing this is with `#[non_exhaustive]`.

guides/building-a-middleware-from-scratch.md
```diff
@@ -0,0 +1,693 @@
...
+        BoxError,
+    >;
+
+    fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+        let this = self.project();
+
+        match this.response_future.poll(cx) {
+            Poll::Ready(result) => {
+                // Use `map_err` to convert the error type.
+                let result = result.map_err(Into::into);
```
jonhoo, line 494: I _think_ you can just use `?` directly on `Poll<Result<_, _>>` now?

jonhoo (approved):
Nice! Left some notes, but none are blocking.

guides/building-a-middleware-from-scratch.md
davidpdrsn: Yes, `Timeout` doesn't require its service to be Clone but I still think the general recommendation of making services Clone, so services can be moved into response futures, is important. So I think I'm going to leave this as is.

guides/building-a-middleware-from-scratch.md
davidpdrsn: Yeah I like that 👍

guides/building-a-middleware-from-scratch.md
davidpdrsn: I've added a footnote about it. Generally not a fan of mentioning features that have been "coming soon" for years 😜

guides/building-a-middleware-from-scratch.md
davidpdrsn: Yep! That is something Jon covers in his video so don't think I'll mention it here.

guides/building-a-middleware-from-scratch.md
davidpdrsn: Good point! std mentions that. Gonna add a link to it.

guides/building-a-middleware-from-scratch.md
davidpdrsn: Hm yeah thats true. I think I'm going to leave it though since its extra important for services so users can move the service into the response future.

guides/building-a-middleware-from-scratch.md
davidpdrsn: That doesn't seem to work for empty structs https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=2ccc7ef1027273fdb59abe545aa20512

guides/building-a-middleware-from-scratch.md
davidpdrsn: Yeah I think you're right. I'm going to leave it as is because I think the additional verbosity might help those who haven't written futures before.

guides/building-a-middleware-from-scratch.md
davidpdrsn: Hm yeah thats probably true. I'm going to edit things down.

guides/building-a-middleware-from-scratch.md
davidpdrsn: I actually think I'm going to stick with `ResponseFuture` as that is what its called in Tower. I quite like that people can read this guide and when they're done be familiar with the code in Tower.

jonhoo: Yeah, I suppose it's okay in Tower it's okay as the type lives in its own module.

guides/building-a-middleware-from-scratch.md
jonhoo: `#[non_exhaustive]` is only enforced across crate boundaries :)

davidpdrsn: Ah yes of course! Cool. Didn't know 😊

PR #593, opened 2021-06-08
Add `HandleError`

davidpdrsn:
Today in Discord someone asked how to write middleware that maps all
errors into responses, to ensure their service would never return an
error. This is actually something I've had to do recently as well so
thought it made sense to upstream what I had into Tower.

This adds `HandleError` to address that. `HandleError` is a new
middleware that takes a closure used to map errors to responses and
changes the error type to `Infallible`.

Example usage:

```rust
let infallible_service = ServiceBuilder::new()
    .layer(HandleErrorLayer::new(|err: BoxError| {
        make_response(format!("error handled: {}", err))
    }))
    .service(fallible_service);
```

I initially thought this was possible with `MapResult` however that
requires `NewError: From<S::Error>` to map errors in `poll_ready`.
However that doesn't work if `NewError` is `Infallible` as you can never
create an `Infallible` from anything. I suppose you could use
`unreachable!()` but that doesn't feel very clean.

`HandleError` gets around this by, if `poll_ready` fails, storing the
error and immediately returning it in `call`. Sorta similar to
`LoadShed`.

Let me know what you think :) I'll write docs if we decide to move
forward with this.

### DIFF
--- a/tower/src/handle_error/mod.rs
+++ b/tower/src/handle_error/mod.rs
@@ -1,0 +1,274 @@
+// TODO(david): write docs if we go forward with this
+#![allow(missing_docs)]
+
+use futures_core::ready;
+use pin_project::pin_project;
+use std::{
+    convert::Infallible,
+    fmt,
+    future::Future,
+    marker::PhantomData,
+    pin::Pin,
+    task::{Context, Poll},
+};
+use tower_layer::Layer;
+use tower_service::Service;
+
+pub struct HandleErrorLayer<F, R> {
+    f: F,
+    _request: PhantomData<fn() -> R>,
+}
+
+impl<F, R> Clone for HandleErrorLayer<F, R>
+where
+    F: Clone,
+{
+    fn clone(&self) -> Self {
+        Self {
+            f: self.f.clone(),
+            _request: PhantomData,
+        }
+    }
+}
+
+impl<F, R> fmt::Debug for HandleErrorLayer<F, R> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("HandleErrorLayer")
+            .field("f", &format_args!("{}", std::any::type_name::<F>()))
+            .finish()
+    }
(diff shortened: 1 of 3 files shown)


tower/src/handle_error/mod.rs
```diff
@@ -0,0 +1,274 @@
...
+where
+    S: Service<R>,
+    F: FnOnce(S::Error) -> S::Response + Clone,
+{
+    type Response = S::Response;
+    type Error = Infallible;
+    type Future = ResponseFuture<S::Future, F, S::Error>;
+
+    fn poll_ready(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
+        if let Err(err) = ready!(self.inner.poll_ready(cx)) {
```
LucioFranco, line 115: Doesn't this break the service contract?

> If Poll::Ready(Err(_)) is returned, the service is no longer able to service requests and the caller should discard the service instance.

davidpdrsn: Hm yeah I suppose it does :(

hawkw: Yeah, I think errors from `poll_ready` should be propagated --- errors from `call`, on the other hand, _don't_ indicate that the service is unusable, and should be handled.

davidpdrsn: I suppose that means this general strategy of changing the error type to `Infallible` and always producing a response isn't possible 🤔

hawkw: You could maybe have a service whose error type is `Infallible` if it _also_ included a `MakeService` and some mechanism for making new services and driving them for readiness when the inner service fails. You would probably also need to include a backoff strategy for when making a new service fails.

Potentially, there could be a helper to combine a `MakeService`/reconnect strategy with a `HandleError` middleware, and then map all errors to `Infallible`, since we would then know that both `poll_ready` _and_ `call` errors would be handled nicely in that case.

PR #600, opened 2021-09-03
Implement `Layer` for `ServiceBuilder`

davidpdrsn:
@jonhoo from discord:

> I may feel silly for asking this, but why doesn't ServiceBuilder implement Layer?
> Like, it has into_inner, but why is the indirection required?

I have wondered the same thing. It compiles and seems sensible to me.


jonhoo (approved):
👀

PR #612, opened 2021-11-04
chore: replace `Never` with `Infallible`

Folyd:
I'm quite sure the `Never` enums in `retry` and `concurrency` are useless, 
and we can replace another `Never` in `discover` with `Infallible` 
since the tower's MSRV is 1.46.


davidpdrsn:
That would also be breaking if someone has implemented a trait for both `Never` and `Infallible`. Probably unlikely though 🤔

hawkw:
> That would also be breaking if someone has implemented a trait for both `Never` and `Infallible`. Probably unlikely though thinking

Oh, yeah, that's a good point. In that case, we probably can't do this until we're preparing to release v0.5, regardless.

PR #613, opened 2021-11-04
docs: improve docs of `retry budget `

Folyd:


davidpdrsn:
Wanna add a reason to why you believe this needs to be deleted? The module is public so we can't just delete it.

Folyd:
> Wanna add a reason to why you believe this needs to be deleted? The module is public so we can't just delete it.

Yeah, the `Budget` is used by other crates, such as `linkerd2-proxy`. https://github.com/Panther0087/linkerd2-proxy/blob/74762fe16f1a43d5622f8400e851e67092f95ebe/linkerd/retry/src/lib.rs#L9

However, I think we should have more docs about how to use `Budget`. I just found the `Budget` was inspired by finagle's `RetryBudget` according to #98.

Also see: https://finagle.github.io/blog/2016/02/08/retry-budgets/

Folyd:
After diving into the `Linkerd` project, I found this `budget` module just a `linkerd-proxy` retry feature. Those three parameters of [Budget::new()](https://docs.rs/tower/0.4.10/tower/retry/budget/struct.Budget.html#method.new) is perfect mirror to `Linkerd`'s [retry budget](https://linkerd.io/2.11/tasks/configuring-retries/#retry-budgets).

```rust
/// Create a [`Budget`] that allows for a certain percent of the total
/// requests to be retried.
///
/// - The `ttl` is the duration of how long a single `deposit` should be
///   considered. Must be between 1 and 60 seconds.
/// - The `min_per_sec` is the minimum rate of retries allowed to accomodate
///   clients that have just started issuing requests, or clients that do
///   not issue many requests per window.
/// - The `retry_percent` is the percentage of calls to `deposit` that can
///   be retried. This is in addition to any retries allowed for via
///   `min_per_sec`. Must be between 0 and 1000.
///
///   As an example, if `0.1` is used, then for every 10 calls to `deposit`,
///   1 retry will be allowed. If `2.0` is used, then every `deposit`
///   allows for 2 retries.
pub fn new(ttl: Duration, min_per_sec: u32, retry_percent: f32) -> Self {
}
```
```yml
spec:
  retryBudget:
    retryRatio: 0.2
    minRetriesPerSecond: 10
    ttl: 10s
```

Maybe #98 ported the `fiangle`'s `RetryBudget` to `tower` mainly for `Linkerd`? Since I can't find any other public usage of `tower`'s `Budget` except `linkerd-proxy` by [searching Github](https://github.com/search?l=&q=tower+budget+language%3ARust&type=code).  Also, the `Budget`  isn't a general feature fit to most applications, it seems dedicated to `Linkerd`. 

So, I think we should move the `budget.rs` to `linkerd-proxy`. 
Also, ping @hawkw. Any thoughts?

seanmonstar:
A retry budget is a very good idea for anyone using retries. It was built for linkerd, but I think it's important enough for everyone to use it, that it should stay in Tower.

seanmonstar:
That said, I think we could improve the documentation to better show how to use it, and include an example in the `tower::retry` docs encouraging people to use one for their `Policy`.

Folyd:
> That said, I think we could improve the documentation to better show how to use it, and include an example in the `tower::retry` docs encouraging people to use one for their `Policy`.

That cool! We really need good docs and examples about `retry budget`.

olix0r:
I agree with @seanmonstar. It's generally dangerous to implement retries without some limiting factor. Budgets are that limit. Let's definitely improve the docs/examples if there is ambiguity in how these should be used, but this is a generally useful feature of the retry middleware.

Folyd:
Thanks, @carllerche, and @olix0r. I just found a great explanation of why the budget is matters from the `Linkerd` blog.

https://linkerd.io/2019/02/22/how-we-designed-retries-in-linkerd-2-2/#why-budgets-and-not-max-retries

Folyd:
I just improved the docs of `Budget`. Most of the docs are from Linkerd's blog post. The example is inspired by `linkerd-proxy`'s `RetryPolicy`.

seanmonstar (approved):
This looks great to me!

PR #615, opened 2021-11-09
util: add `CloneBoxService`

davidpdrsn:
This upstreams a little utility I'm using a bunch in axum. Its often
useful to erase the type of a service while still being able to clone
it.

`BoxService` isn't `Clone` so previously you had to combine it with
`Buffer` but doing that a lot (which we did in axum) had measurable
impact on performance.

Note this doesn't add a `!Send` variant. I personally have no use
for that but if someone thinks it would be nice I can do a follow up PR.

### DIFF
--- a/tower/CHANGELOG.md
+++ b/tower/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 # Unreleased
 
+- **util**: Add `CloneService` which is a `Clone + Send` boxed `Service`.
+
 ### Fixed
 
 - **balance**: Remove redundant `Req: Clone` bound from `Clone` impls
--- a/tower/CHANGELOG.md
+++ b/tower/CHANGELOG.md
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 # Unreleased
 
-- **util**: Add `CloneService` which is a `Clone + Send` boxed `Service`.
+- **util**: Add `CloneBoxService` which is a `Clone + Send` boxed `Service`.
 
 ### Fixed
 
--- a/tower/src/util/clone_boxed.rs
+++ b/tower/src/util/clone_boxed.rs
@@ -1,0 +1,138 @@
+use super::ServiceExt;
+use futures_util::future::BoxFuture;
+use std::{
+    fmt,
+    task::{Context, Poll},
+};
+use tower_layer::{layer_fn, LayerFn};
+use tower_service::Service;
+
+/// A `Clone + Send` boxed `Service`.
+///
+/// [`CloneBoxService`] turns a service into a trait object, allowing the
+/// response future type to be dynamic, and allowing the service to be cloned.
+///
+/// # Example
+///
+/// ```
+/// use tower::{Service, ServiceBuilder, BoxError, util::CloneBoxService};
+/// use std::time::Duration;
+/// #
+/// # struct Request;
+/// # struct Response;
+/// # impl Response {
+/// #     fn new() -> Self { Self }
+/// # }
+///
+/// // This service has a complex type that is hard to name
+/// let service = ServiceBuilder::new()
+///     .map_request(|req| {
+///         println!("received request");
+///         req
+///     })
+///     .map_response(|res| {
+///         println!("response produced");
+///         res
+///     })
+///     .load_shed()
+///     .concurrency_limit(64)
+///     .timeout(Duration::from_secs(10))
+///     .service_fn(|req: Request| async {
+///         Ok::<_, BoxError>(Response::new())
+///     });
+/// # let service = assert_service(service);
+///
+/// // `CloneService` will erase the type so its nameable
+/// let service: CloneBoxService<Request, Response, BoxError> = CloneBoxService::new(service);
+/// # let service = assert_service(service);
+///
+/// // And we can still clone the service
+/// let cloned_service = service.clone();
+/// #
+/// # fn assert_service<S, R>(svc: S) -> S
+/// # where S: Service<R> { svc }
+/// ```
+pub struct CloneBoxService<T, U, E>(
+    Box<
+        dyn CloneService<T, Response = U, Error = E, Future = BoxFuture<'static, Result<U, E>>>
+            + Send,
+    >,
+);
+
+impl<T, U, E> CloneBoxService<T, U, E> {
+    /// Create a new `CloneBoxService`.
+    pub fn new<S>(inner: S) -> Self
+    where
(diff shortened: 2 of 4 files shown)


tower/src/util/clone_boxed.rs
```diff
@@ -0,0 +1,138 @@
+use super::ServiceExt;
+use futures_util::future::BoxFuture;
+use std::{
+    fmt,
+    task::{Context, Poll},
+};
+use tower_layer::{layer_fn, LayerFn};
+use tower_service::Service;
+
+/// A `Clone + Send` boxed `Service`.
```
hawkw, line 10: nit, TIOLI: should we maybe link to the docs for the `Clone` and `Service` traits?

tower/src/util/clone_boxed.rs
```diff
@@ -0,0 +1,138 @@
...
+///
+/// [`CloneBoxService`] turns a service into a trait object, allowing the
+/// response future type to be dynamic, and allowing the service to be cloned.
```
hawkw, line 13: it would be nice if the docs for this included a comparison with the non-`Clone` `BoxService` type? something like "This is similar to the [`BoxService`] type, but ..."

tower/src/util/clone_boxed.rs
```diff
@@ -0,0 +1,138 @@
...
+///     })
+///     .load_shed()
+///     .concurrency_limit(64)
+///     .timeout(Duration::from_secs(10))
+///     .service_fn(|req: Request| async {
+///         Ok::<_, BoxError>(Response::new())
+///     });
+/// # let service = assert_service(service);
+///
+/// // `CloneService` will erase the type so its nameable
```
hawkw, line 45: minor nit: when used as a contraction for "it is", you want "it's", rather than "its":
```suggestion
/// // `CloneService` will erase the type so it's nameable
```

tower/src/util/clone_boxed.rs
```diff
@@ -0,0 +1,138 @@
...
+        Box::new(self.clone())
+    }
+}
+
+impl<T, U, E> fmt::Debug for CloneBoxService<T, U, E>
+where
+    T: fmt::Debug,
+    U: fmt::Debug,
+    E: fmt::Debug,
+{
```
hawkw, line 134: the debug output doesn't include any of these type parameters, so I believe these can all be removed, and the type can always be `Debug`:
```suggestion
impl<T, U, E> fmt::Debug for CloneBoxService<T, U, E> {
```

tower/src/util/clone_boxed.rs
```diff
@@ -0,0 +1,138 @@
...
+    fn fmt(&self, fmt: &mut fmt::Formatter) -> fmt::Result {
+        fmt.debug_struct("CloneBoxService").finish()
```
hawkw, line 136: take it or leave it: we _may_ want to consider including the request and response type _names_ here, so different instances of `CloneBoxService` can be distinguished? Maybe something like
```suggestion
        write!(
            fmt, 
            "CloneBoxService<Request = {}, Response = {}, Error = {}>",
            std::any::type_name::<T>(),
            std::any::type_name::<U>(),
            std::any::type_name::<E>(),
        )
```
but, on the other hand, this could be quite long, and we might want to correctly handle multi-line formatting in alt mode (`{:#}`)...so, maybe that's something to hold off on and add later if it's needed...

hawkw: also, `BoxService` doesn't do this, so it's probably better left for a separate PR if we _do_ want to add something like that...

hawkw (changes requested):
Overall, this looks great to me --- we may also want to add methods to `ServiceExt` and/or `ServiceBuilder` for wrapping a service in `CloneBoxService`, but we can do that in a follow up PR.

I commented on a few minor things. The only one that I think is actually important to address before merging is the unnecessary `fmt::Debug` bounds on `CloneBoxService`'s type parameters: since those types are never formatted in the `Debug` impl, it shouldn't be necessary for them to implement `Debug`.

Beyond that, this looks good to me and I'll be happy to approve it once that's addressed!

tower/src/util/clone_boxed.rs
davidpdrsn: One day I'll learn this 😜

tower/src/util/clone_boxed.rs
davidpdrsn: I was actually thinking the type params are there such that this could be improved in the future, without a breaking change. I think I'll leave it as it is now and make an issue about improving the formatting, for `BoxService` as well.

tower/CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 # Unreleased
 
+- **util**: Add `CloneService` which is a `Clone + Send` boxed `Service`.
```
olix0r, line 10: Here and elsewhere in the comments you refer to `CloneService`, but I think you mean `CloneBoxService`?
```suggestion
- **util**: Add `CloneBoxService` which is a `Clone + Send` boxed `Service`.
```

tower/src/util/clone_boxed.rs
olix0r:
```suggestion
/// // `CloneBoxService` will erase the type so it's nameable.
```

tower/CHANGELOG.md
davidpdrsn: Ah yes good catch! I was a bit too quick on the auto complete.

tower/src/util/clone_boxed.rs
hawkw: > I was actually thinking the type params are there such that this could be improved in the future,

I don't think we want the type params to be `Debug` in any case; we can't actually include values of those types in the formatted output, since they're not part of the struct. I think the where clause won't ever be useful for future improvements to the debug format, since we'll never have instances of those types to format when formatting a `CloneBoxService` instance. We can use `type_name` to get the _names_ of those types without requiring those types to be `Debug`.

tower/src/util/clone_boxed.rs
davidpdrsn: I fixed the typos in https://github.com/tower-rs/tower/pull/615/commits/28915aa9d1cab86aebb38a8fd6f928f0bd1c58d4

davidpdrsn:
@hawkw I think adding more convenience methods to `ServiceExt` and `ServiceBuilder` for this is a good idea. I'll do that in a follow up PR.

tower/src/util/clone_boxed.rs
davidpdrsn: Oh yeah doh of course! I've removed the bounds.

hawkw (approved):
okay, this looks good to me! thanks!

PR #616, opened 2021-11-11
builder,util: add convenience methods for boxing services

davidpdrsn:
This adds a couple of new methods to `ServiceBuilder` and `ServiceExt`:

- `ServiceBuilder::boxed`
- `ServiceExt::boxed`
- `ServiceBuilder::clone_boxed`
- `ServiceExt::clone_boxed`

### DIFF
--- a/tower/CHANGELOG.md
+++ b/tower/CHANGELOG.md
@@ -8,6 +8,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 # Unreleased
 
 - **util**: Add `CloneBoxService` which is a `Clone + Send` boxed `Service`.
+- **util:** Add `ServiceExt::boxed` for applying a `BoxService` middleware.
+- **util:** Add `ServiceExt::clone_boxed` for applying a `CloneBoxService` middleware.
+- **builder:** Add `ServiceExtBuilder::boxed` for applying a `BoxService` layer.
+- **builder:** Add `ServiceExtBuilder::clone_boxed` for applying a `CloneBoxService` layer.
 
 ### Fixed
 
--- a/tower/src/builder/mod.rs
+++ b/tower/src/builder/mod.rs
@@ -688,6 +688,117 @@ impl<L> ServiceBuilder<L> {
     {
         self
     }
+
+    /// Convert the resulting service into a [`Service`] + [`Send`] trait object.
+    ///
+    /// This applies the [`BoxService:layer()`] layer. See that method for more details.
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// use tower::{Service, ServiceBuilder, BoxError, util::BoxService};
+    /// use std::time::Duration;
+    /// #
+    /// # struct Request;
+    /// # struct Response;
+    /// # impl Response {
+    /// #     fn new() -> Self { Self }
+    /// # }
+    ///
+    /// let service: BoxService<Request, Response, BoxError> = ServiceBuilder::new()
+    ///     .boxed()
+    ///     .load_shed()
+    ///     .concurrency_limit(64)
+    ///     .timeout(Duration::from_secs(10))
+    ///     .service_fn(|req: Request| async {
+    ///         Ok::<_, BoxError>(Response::new())
+    ///     });
+    /// # let service = assert_service(service);
+    /// # fn assert_service<S, R>(svc: S) -> S
+    /// # where S: Service<R> { svc }
+    /// ```
+    ///
+    /// [`BoxService:layer()`]: crate::util::BoxService::layer()
+    pub fn boxed<S, R>(
+        self,
+    ) -> ServiceBuilder<
+        Stack<
+            tower_layer::LayerFn<
+                fn(
+                    L::Service,
+                ) -> crate::util::BoxService<
+                    R,
+                    <L::Service as Service<R>>::Response,
+                    <L::Service as Service<R>>::Error,
+                >,
+            >,
+            L,
+        >,
+    >
+    where
+        L: Layer<S>,
+        L::Service: Service<R> + Send + 'static,
+        <L::Service as Service<R>>::Future: Send + 'static,
+    {
+        self.layer(crate::util::BoxService::layer())
+    }
+
+    /// Convert the resulting service into a [`Service`] + [`Clone`] + [`Send`] trait object.
+    ///
+    /// This applies the [`CloneBoxService:layer()`] layer. See that method for more details.
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// use tower::{Service, ServiceBuilder, BoxError, util::CloneBoxService};
+    /// use std::time::Duration;
+    /// #
+    /// # struct Request;
+    /// # struct Response;
+    /// # impl Response {
+    /// #     fn new() -> Self { Self }
+    /// # }
+    ///
+    /// let service: CloneBoxService<Request, Response, BoxError> = ServiceBuilder::new()
+    ///     .clone_boxed()
+    ///     .load_shed()
+    ///     .concurrency_limit(64)
+    ///     .timeout(Duration::from_secs(10))
+    ///     .service_fn(|req: Request| async {
+    ///         Ok::<_, BoxError>(Response::new())
+    ///     });
+    /// # let service = assert_service(service);
+    ///
+    /// // The boxed service can still be cloned.
+    /// service.clone();
+    /// # fn assert_service<S, R>(svc: S) -> S
+    /// # where S: Service<R> { svc }
+    /// ```
+    ///
+    /// [`CloneBoxService:layer()`]: crate::util::CloneBoxService::layer()
+    pub fn clone_boxed<S, R>(
+        self,
+    ) -> ServiceBuilder<
+        Stack<
+            tower_layer::LayerFn<
+                fn(
+                    L::Service,
+                ) -> crate::util::CloneBoxService<
+                    R,
+                    <L::Service as Service<R>>::Response,
+                    <L::Service as Service<R>>::Error,
+                >,
+            >,
+            L,
+        >,
+    >
+    where
+        L: Layer<S>,
+        L::Service: Service<R> + Clone + Send + 'static,
+        <L::Service as Service<R>>::Future: Send + 'static,
+    {
+        self.layer(crate::util::CloneBoxService::layer())
+    }
 }
 
 impl<L: fmt::Debug> fmt::Debug for ServiceBuilder<L> {
--- a/tower/src/util/mod.rs
+++ b/tower/src/util/mod.rs
@@ -953,6 +953,91 @@ pub trait ServiceExt<Request>: tower_service::Service<Request> {
     {
         MapFuture::new(self, f)
     }
+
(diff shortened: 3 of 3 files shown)


tower/src/builder/mod.rs
```diff
@@ -688,6 +688,121 @@ impl<L> ServiceBuilder<L> {
...
+    /// // The boxed service can still be cloned.
+    /// service.clone();
+    /// # fn assert_service<S, R>(svc: S) -> S
+    /// # where S: Service<R> { svc }
+    /// ```
+    ///
+    /// [`CloneBoxService:layer()`]: crate::util::CloneBoxService::layer()
+    #[cfg(feature = "util")]
+    #[cfg_attr(docsrs, doc(cfg(feature = "util")))]
+    pub fn clone_boxed<S, R>(
```
hawkw, line 783: TIOLI but I might have called this `boxed_clone`? although on the other hand, this isn't consistent with the name of the return type...idk.

tower/src/util/mod.rs
```diff
@@ -953,6 +953,95 @@ pub trait ServiceExt<Request>: tower_service::Service<Request> {
...
+    fn boxed(self) -> BoxService<Request, Self::Response, Self::Error>
+    where
+        Self: Sized + Send + 'static,
+        Self::Future: Send + 'static,
+    {
+        BoxService::new(self)
+    }
+
+    /// Convert the service into a [`Service`] + [`Clone`] + [`Send`] trait object.
+    ///
```
hawkw, line 1001: might be nice to have a line like 
```suggestion
    /// Convert the service into a [`Service`] + [`Clone`] + [`Send`] trait object.
    ///
    /// This is similar to the [`boxed`] method, but it requires that `Self` implement
    /// [`Clone`], and the returned boxed service implements [`Clone`].
```

tower/src/util/mod.rs
```diff
@@ -953,6 +953,95 @@ pub trait ServiceExt<Request>: tower_service::Service<Request> {
     {
         MapFuture::new(self, f)
     }
+
+    /// Convert the service into a [`Service`] + [`Send`] trait object.
+    ///
+    /// See [`BoxService`] for more details.
+    ///
```
hawkw, line 960: maybe worth adding
```suggestion
    ///
    /// If `Self` implements the [`Clone`] trait, the [`boxed_clone`] method
    /// can be used instead, to produce a boxed service which will also
    /// implement [`Clone`].
    ///
```

tower/src/builder/mod.rs
```diff
@@ -688,6 +688,121 @@ impl<L> ServiceBuilder<L> {
     {
         self
     }
+
+    /// Convert the resulting service into a [`Service`] + [`Send`] trait object.
+    ///
+    /// This applies the [`BoxService:layer()`] layer. See that method for more details.
```
hawkw, line 694: nit: wrong link
```suggestion
    /// This applies the [`BoxService::layer()`] layer. See that method for more details.
```

tower/src/builder/mod.rs
```diff
@@ -688,6 +688,121 @@ impl<L> ServiceBuilder<L> {
...
+    ///     .timeout(Duration::from_secs(10))
+    ///     .service_fn(|req: Request| async {
+    ///         Ok::<_, BoxError>(Response::new())
+    ///     });
+    /// # let service = assert_service(service);
+    /// # fn assert_service<S, R>(svc: S) -> S
+    /// # where S: Service<R> { svc }
+    /// ```
+    ///
+    /// [`BoxService:layer()`]: crate::util::BoxService::layer()
```
hawkw, line 721:
```suggestion
    /// [`BoxService::layer()`]: crate::util::BoxService::layer()
```

tower/src/builder/mod.rs
```diff
@@ -688,6 +688,121 @@ impl<L> ServiceBuilder<L> {
...
+        L: Layer<S>,
+        L::Service: Service<R> + Send + 'static,
+        <L::Service as Service<R>>::Future: Send + 'static,
+    {
+        self.layer(crate::util::BoxService::layer())
+    }
+
+    /// Convert the resulting service into a [`Service`] + [`Clone`] + [`Send`] trait object.
+    ///
+    /// This applies the [`CloneBoxService:layer()`] layer. See that method for more details.
```
hawkw, line 750:
```suggestion
    /// This applies the [`CloneBoxService::layer()`] layer. See that method for more details.
```

tower/src/builder/mod.rs
```diff
@@ -688,6 +688,121 @@ impl<L> ServiceBuilder<L> {
...
+    ///     });
+    /// # let service = assert_service(service);
+    ///
+    /// // The boxed service can still be cloned.
+    /// service.clone();
+    /// # fn assert_service<S, R>(svc: S) -> S
+    /// # where S: Service<R> { svc }
+    /// ```
+    ///
+    /// [`CloneBoxService:layer()`]: crate::util::CloneBoxService::layer()
```
hawkw, line 780:
```suggestion
    /// [`CloneBoxService::layer()`]: crate::util::CloneBoxService::layer()
```

tower/src/builder/mod.rs
```diff
@@ -688,6 +688,121 @@ impl<L> ServiceBuilder<L> {
...
+    where
+        L: Layer<S>,
+        L::Service: Service<R> + Send + 'static,
+        <L::Service as Service<R>>::Future: Send + 'static,
+    {
+        self.layer(crate::util::BoxService::layer())
+    }
+
+    /// Convert the resulting service into a [`Service`] + [`Clone`] + [`Send`] trait object.
+    ///
```
hawkw, line 749: nit, take it or leave it: maybe worth adding a note comparing this with `box_service`, like I mentioned on the `ServiceExt` methods? not a blocker...

tower/src/builder/mod.rs
hawkw: also, TIOLI, but most of the other `ServiceBuilder` methods use a slightly different phrasing "This wraps ...":
https://github.com/tower-rs/tower/blob/ef9066b567af0191007dc94c2c04573ead5fbadd/tower/src/builder/mod.rs#L193-L194
so, maybe:
```suggestion
    /// This wraps the inner service with the [`Layer`] returned by [`BoxService::layer()`]. 
    /// See that method for more details.
```

tower/src/builder/mod.rs
hawkw: also, maybe:
```suggestion
    /// This wraps the inner service with the [`Layer`] returned by [`CloneBoxService::layer()`]. 
    /// See that method for more details.
```

tower/src/builder/mod.rs
```diff
@@ -688,6 +688,121 @@ impl<L> ServiceBuilder<L> {
...
+            tower_layer::LayerFn<
+                fn(
+                    L::Service,
+                ) -> crate::util::CloneBoxService<
+                    R,
+                    <L::Service as Service<R>>::Response,
+                    <L::Service as Service<R>>::Error,
+                >,
+            >,
+            L,
```
hawkw, line 796: gosh... since the `CloneBoxService::layer()` method hasn't been released yet, we _could_ potentially replace the `LayerFn` with a named `Layer` type. IDK if it's worth it just to get rid of this type signature, but...wow.

tower/CHANGELOG.md
```diff
@@ -8,6 +8,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 # Unreleased
 
 - **util**: Add `CloneBoxService` which is a `Clone + Send` boxed `Service`.
+- **util:** Add `ServiceExt::boxed` for applying a `BoxService` middleware.
+- **util:** Add `ServiceExt::clone_boxed` for applying a `CloneBoxService` middleware.
+- **builder:** Add `ServiceExtBuilder::boxed` for applying a `BoxService` layer.
+- **builder:** Add `ServiceExtBuilder::clone_boxed` for applying a `CloneBoxService` layer.
```
hawkw, line 14: take it or leave it: i think it would probably be fine to combine some of these changelog entries, like
```suggestion
- **util:** Add `ServiceExt::boxed` and `ServiceExt::clone_boxed` for applying the
   `BoxService` and `CloneBoxService` middleware.
- **builder:** Add `ServiceBuilder::boxed` and `ServiceBuilder::clone_boxed` for applying 
   `BoxService` and `CloneBoxService` layers.
```

hawkw: also, i noticed these changelog entries refer to `ServiceExtBuilder`, which we should fix regardless of whether we apply my suggestion to combine the entries:
```suggestion
- **util:** Add `ServiceExt::boxed` for applying a `BoxService` middleware.
- **util:** Add `ServiceExt::clone_boxed` for applying a `CloneBoxService` middleware.
- **builder:** Add `ServiceBuilder::boxed` for applying a `BoxService` layer.
- **builder:** Add `ServiceBuilder::clone_boxed` for applying a `CloneBoxService` layer.
```

hawkw (approved):
I commented on some docs nits and stuff, but I'm mostly good with this. It would be nice to fix the wrong syntax in links before merging, though.

tower/src/builder/mod.rs
```diff
@@ -688,6 +688,121 @@ impl<L> ServiceBuilder<L> {
...
+        Stack<
+            tower_layer::LayerFn<
+                fn(
+                    L::Service,
+                ) -> crate::util::BoxService<
+                    R,
+                    <L::Service as Service<R>>::Response,
+                    <L::Service as Service<R>>::Error,
+                >,
+            >,
```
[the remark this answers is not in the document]
davidpdrsn: I went down that route initially actually but you end up having to have the request type as a type param on the layer:

```rust
// have to have `T` here otherwise its unconstrained in the layer impl
pub struct CloneBoxServiceLayer<T> {
    _request: PhantomData<T>,
}

impl<S, T> Layer<S> for CloneBoxServiceLayer<T>
where
    S: Service<T> + Clone + Send + 'static,
    S::Future: Send + 'static,
{
    type Service = CloneBoxService<T, S::Response, S::Error>;

    fn layer(&self, inner: S) -> Self::Service {
        CloneBoxService::new(inner)
    }
}
```

That could be workable to leads to inference issues when actually using the builder method:

```
---- src/builder/mod.rs - builder::ServiceBuilder<L>::clone_boxed (line 754) stdout ----
error[E0284]: type annotations needed
  --> src/builder/mod.rs:765:6
   |
14 |     .clone_boxed()
   |      ^^^^^^^^^^^ cannot infer type for type parameter `S`
   |
   = note: cannot satisfy `<_ as Service<Request>>::Future == _`
```

tower/src/builder/mod.rs
davidpdrsn: Right I went with this name because it matches the service name, which we can rename since its not released. `CloneBoxService` or `BoxCloneService` seem kinda same same to me though.

tower/src/builder/mod.rs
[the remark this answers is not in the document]
hawkw: ugh, never mind. it's probably better to avoid introducing phantomdata anyway, as they will result in rustc having to typecheck a larger type, potentially impacting build time... this is fine, then.

hawkw (approved):

PR #620, opened 2021-11-19
tower: update hdrhistogram to 7.0 and remove useless dependencies

koushiro:
Signed-off-by: koushiro <koushiro.cqx@gmail.com>

* update `hdrhistogram` to 7.0
* remove useless dependency 'quickcheck'
* add `path` for `tower-service` dependency.


davidpdrsn (approved):
LGTM!

hdrhistogram doesn't seem to be public dependency so updating it should be fine.

PR #621, opened 2021-11-24
Assert that no unsafe code is used in tower

olix0r:
The tower crates do not include any `unsafe` code, but tools like
[`cargo-geiger`][cg] can't necessarily detect that. This change adds a
stronger assertion at the top of each crate with a
`#![forbid(unsafe_code)]` directive to assert that no unsafe code is
used in the crate. This also serves as a more obvious obstacle to
introducing unsafe code in future changes.

[cg]: https://github.com/rust-secure-code/cargo-geiger

### DIFF
--- a/tower-layer/src/lib.rs
+++ b/tower-layer/src/lib.rs
@@ -5,6 +5,7 @@
     rust_2018_idioms,
     unreachable_pub
 )]
+#![forbid(unsafe_code)]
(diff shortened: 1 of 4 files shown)


davidpdrsn (approved):
👌

hawkw (approved):
maybe worth a README/docs note to the affect that all Tower crates have `#![forbid(unsafe_code)]`? not a blocker.

PR #623, opened 2021-12-29
Import tracing without the attributes feature

paolobarbolini:
tracing-attributes depends on syn and proc-macro2, which are slow to compile


hawkw (approved):
looks good to me, given that nothing in `tower` uses the attribute macros

davidpdrsn (approved):

Issue #550, opened 2021-01-27
Consider making Either uniform with futures::Either


olix0r:
The `tower::Either` behaviors similarly to `futures::Either`, but they've diverged over time. Namely, tower's is:

```rust
enum Either<A, B> {
    A(A),
    B(B),
}
```

but future's is now:

```rust
enum Either<L, R> {
    Left(L),
    Right(R),
}
```

* Should we rename our either variants to match future's?
* Can we simply re-export the `futures::Either` type and provide `Service` impls for it?

davidpdrsn:
> Should we rename our either variants to match future's?

Yeah I think thats a good idea.

> Can we simply re-export the futures::Either type and provide Service impls for it?

There is a bit more to tower's `impl Future for Either` and futures's so not sure thats possible. It uses `#[pin_project]` and has bounds the variants containing `Result`s.

olix0r:
> There is a bit more to tower's impl Future for Either and futures's so not sure thats possible. It uses #[pin_project] and has bounds the variants containing Results.

Does it need to? I.e. if we inherit the Future impl, can it be used for our needs to satisfy `Service<Req> for futures::Either<L,R>`?

olix0r:
As discussed in #629, we should probably update the `Either` service's return type to be an opaque future so we can drop the dependency on `pin-project` in tower. We should also require the inner service's to have a uniform error type (rather than boxing the error, potentially erasing infallibility).

PR #629, opened 2022-01-17
Remove last dependency on pin-project

jplatte:
Followup to / depends on #603.

Was controversial and thus split out into its own PR (see https://github.com/tower-rs/tower/pull/603#issuecomment-1014778924 and following comments). The reason this can't be done using `pin-project-lite` is now documented as a code comment.

With a breaking-change coming up soon-ish, I think another alternative would be to replace this type with the one from the `either` crate and have them provide `EitherProj` plus the necessary unsafe code That would imply moving the `Layer` and `Service` implementations for `Either` out to `tower-service` & `tower-layer` but I think that should be fine. It would also imply a bunch of extra impls on `Either` that may or may not be nice to have, as well as an extra API change since `either::Either` has `Left` and `Right` variants rather than `A` and `B`.

cc @olix0r @davidpdrsn

### DIFF
--- a/tower-layer/Cargo.toml
+++ b/tower-layer/Cargo.toml
@@ -1,7 +1,6 @@
 [package]
 name = "tower-layer"
 # When releasing to crates.io:
-# - Remove path dependencies
 # - Update doc url
 #   - Cargo.toml
 #   - README.md
--- a/tower-layer/Cargo.toml
+++ b/tower-layer/Cargo.toml
@@ -2,7 +2,6 @@
 name = "tower-layer"
 # When releasing to crates.io:
 # - Remove path dependencies
-# - Update html_root_url.
 # - Update doc url
 #   - Cargo.toml
 #   - README.md
--- a/tower-layer/src/lib.rs
+++ b/tower-layer/src/lib.rs
@@ -1,4 +1,3 @@
-#![doc(html_root_url = "https://docs.rs/tower-layer/0.3.1")]
 #![warn(
     missing_debug_implementations,
     missing_docs,
--- a/tower-service/Cargo.toml
+++ b/tower-service/Cargo.toml
@@ -2,7 +2,6 @@
 name = "tower-service"
 # When releasing to crates.io:
 # - Remove path dependencies
-# - Update html_root_url.
 # - Update doc url
 #   - Cargo.toml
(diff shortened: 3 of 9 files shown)


olix0r:
I'm sympathetic to `pin-project` being a heavyweight dependency for such a simple use case.

On the other hand, I'm also conTcerned that this change will make tools like [`cargo-geiger`](https://github.com/rust-secure-code/cargo-geiger) flag `tower`.

Removing `pin-project` is probably worth it...

While we're here though: why does `Either` need to hold pinned values at all? Is this is just to implement `Future`? Doesn't the `futures` crate already ship a version of `Either` that implements `Future`? Perhaps we can remove this functionality completely in the next minor release?

olix0r:
Also note that https://github.com/tower-rs/tower/issues/550 tracks updating the `Either` type to have `Left` and `Right` variants, planned for the next minor release.

davidpdrsn:
My 2 cents: I think we should leave `Either` as it is currently and eat the dependency on `pin-project`. However for the next breaking version of tower I think we should change `Either` to not implement `Future` and change its response future to be a struct with a private inner enum that uses `pin-project-lite`.

While we're at it I also think we should change `Either` to not use `BoxError` as the error type but instead require the two inner services to have the same error type and use that. I've found it odd that combining two infallible services and gives you one that is fallible, since [axum requires all services to be infallible](https://docs.rs/axum/latest/axum/error_handling/index.html).

jplatte:
> While we're here though: why does `Either` need to hold pinned values at all? Is this is just to implement `Future`? Doesn't the `futures` crate already ship a version of `Either` that implements `Future`? Perhaps we can remove this functionality completely in the next minor release?

It's part of futures-util: https://docs.rs/futures-util/latest/futures_util/future/enum.Either.html

What I wrote about `either::Either` could of course be done with that one too. The impls in tower-service and tower-layer could even be added in a patch release, then other crates could build things off of that sooner.

> While we're at it I also think we should change `Either` to not use `BoxError` as the error type but instead require the two inner services to have the same error type and use that. I've found it odd that combining two infallible services and gives you one that is fallible, since axum requires all services to be infallible.

👍 I've had this issue as well and had to roll my own `option_layer` because of it, including another separate `Either` type.

PR #635, opened 2022-02-01
rewrite `buffer` to use bounded MPSC

hawkw:
## Motivation

Currently, `tower::buffer` uses `tokio::sync::mpsc`'s _unbounded_
channel plus a `tokio::sync::Semaphore`, in order to re-implement a
bounded channel. This was necessary because when this code was updated
to the latest version of `tokio`, there was no way to reserve a
non-borrowed send permit from a `Sender`. Thus, it was necessary to use
the `Semaphore` for the future that is polled in `poll_ready` to acquire
send capacity, since a `Permit` from the `Sender` could not be stored in
the struct until it's consumed in `call`.

This code is Not Ideal. Reimplementing the bounded channel makes the
implementation more complicated, and means that there is a bunch of
extra stuff we have to do to e.g. propagate cancellations/service errors
to tasks waiting on `poll_ready`. The bounded MPSC would solve most of
this for us. It might also be a bit more efficient, since we would only
have a single reference-counted heap allocation (the `Sender`), rather
than two (the `Sender` _and_ the `Arc<Semaphore>`).

Now that `tokio::sync::mpsc::Sender` has a `reserve_owned` method, we no
longer need to use the semaphore, and can simplify the buffer
implementation a bit.

## Solution

This branch changes the buffer to use only a single bounded MPSC, rather
than an unbounded MPSC and a semaphore. The bounded MPSC internally
manages its semaphore, so we can now remove a lot of complexity in the
current implementation.

I had to change some of the integration tests slightly as part of this
change. This is because the buffer implementation using semaphore
permits is _very subtly_ different from one using a bounded channel. In
the `Semaphore`-based implementation, a semaphore permit is stored in
the `Message` struct sent over the channel. This is so that the capacity
is used as long as the message is in flight. However, when the worker
task is processing a message that's been recieved from the channel, the
permit is still not dropped. Essentially, the one message actively held
by the worker task _also_ occupies one "slot" of capacity, so the actual
channel capacity is one less than the value passed to the constructor,
_once the first request has been sent to the worker_. The bounded MPSC
changed this behavior so that capacity is only occupied while a request
is actually in the channel, which broke some tests that relied on the
old (and technically wrong) behavior.

## Notes

There is one sort of significant issue with this change, which is why
it's currently a draft. The issue is that this unfortunately requires
adding `Send` and `'static` bounds to the `T::Future` and `Request`
types. This is because they occur within the `Message` type that's sent
over the channel, and a MPSC's `OwnedPermit<T>` for a message of type
`T` is only `Send` or `'static` when `T` is. The `ReusableBoxFuture`
that the `reserve_owned` future is stored in requires the future be
`Send + 'static`, which it is not when `OwnedPermit` isn't.

I don't believe that it's actually _necessary_ for the `OwnedPermit<T>`
type to require `T: Send` in order to be `Send`, or `T: 'static` in
order to be valid for the `'static` lifetime. An `OwnedPermit` never
actually contains an instance of type `T`, it just represents the
_capacity_ to send that type to the channel. The channel itself will
actually contain the values of type `T`. Therefore, it's possible this
could be changed upstream in Tokio, although I haven't looked into it
yet.

In practice, these bounds shouldn't really be a problem, but adding them
is a breaking change. We could alternatively just wait for `tower` 0.5
to merge this PR.


tower/src/buffer/service.rs
```diff
@@ -105,45 +117,79 @@ where
...
+        }
+    }
+}
+
 impl<T, Request> Service<Request> for Buffer<T, Request>
 where
     T: Service<Request>,
     T::Error: Into<crate::BoxError>,
+    T::Future: Send + 'static,
+    Request: Send + 'static,
```
hawkw, line 146: this is the unfortunate new bounds i mentioned in the PR description

tower/tests/buffer/main.rs
```diff
@@ -243,6 +243,14 @@ async fn wakes_pending_waiters_on_close() {
     assert_pending!(worker.poll());
     let mut response = task::spawn(service1.call("hello"));
 
+    assert!(worker.is_woken(), "worker task should be woken by request");
+    assert_pending!(worker.poll());
+
+    // fill the channel so all subsequent requests will wait for capacity
+    let service1 = assert_ready_ok!(task::spawn(service.ready()).poll());
+    assert_pending!(worker.poll());
+    let mut response2 = task::spawn(service1.call("world"));
```
hawkw, line 252: this test required adding an additional message to the channel for the service to be "full", I described this in the PR description

tower/tests/buffer/main.rs
```diff
@@ -301,6 +316,14 @@ async fn wakes_pending_waiters_on_failure() {
...
```
hawkw, line 325: this test required adding an additional message to the channel for the buffer to be "full", I described this in the PR description

tower/src/buffer/service.rs
```diff
@@ -19,26 +24,39 @@ pub struct Buffer<T, Request>
...
-    // `async fn ready`, which borrows the sender. Therefore, we implement our
-    // own bounded MPSC on top of the unbounded channel, using a semaphore to
-    // limit how many items are in the channel.
-    semaphore: PollSemaphore,
-    // The current semaphore permit, if one has been acquired.
-    //
-    // This is acquired in `poll_ready` and taken in `call`.
-    permit: Option<OwnedSemaphorePermit>,
+    tx: mpsc::Sender<Message<Request, T::Future>>,
+    reserve: ReusableBoxFuture<Result<Permit<Request, T::Future>, SendError<()>>>,
```
olix0r, line 28: I think I understand what this is doing, but worth a docstring?

olix0r (approved):
This makes sense to me!

hawkw:
Now that `tokio-util` 0.7 is released, this code can be rewritten to just use the `PollSender` type. This is because the semantics of `PollSender` was changed to have a `poll_reserve` rather than `poll_send_done`, which is the required behavior for `Buffer`. The implementation of `PollSender` now looks more or less identical to the code I had written using `ReusableBoxFuture` and `reserve_owned`. Therefore, we can now just use the `tokio_util::sync::PollSender` type, and the buffer implementation is now much simpler.

LucioFranco:
@hawkw is this still blocked on the breaking change?

hawkw:
> @hawkw is this still blocked on the breaking change?

yes, this introduces [trait bounds](https://github.com/tower-rs/tower/pull/635/files#diff-35398bbf39b455b121231a971e231cfc669556eda1c98705f367aa774f3da86cR92-R93) that were previously not required. It's _possible_ an upstream change in `tokio-util` could make them unnecessary, but I'm not really sure how easy that would be yet (haven't looked too closely at it).

LucioFranco:
I vote that we go ahead with a breaking change for this?

LucioFranco:
cc @davidpdrsn

hawkw:
> I vote that we go ahead with a breaking change for this?

Yeah, I'm hoping to prepare an 0.5 in the next couple days. we should ship another 0.4 release first though, which is why it's a draft.

I do also want to look at `tokio-util` to double check if it's possible to not introduce those bounds, because it is kind of a bummer that `buffer` now cannot be used with `!Sync` services...

davidpdrsn (approved):
I'm cool with moving ahead with a 0.5

Issue #594, opened 2021-07-21
Switch from pin-project to pin-project-lite


seanmonstar:
I doubt we actually need the extra features of the former, and changing to to "lite" will be a compilation speed improvement.

hawkw:
Looks like `tower::buffer` currently uses `PinnedDrop`: https://github.com/tower-rs/tower/blob/e1760d385d06c2f55dbd0b0b8054f176cff8c505/tower/src/buffer/worker.rs#L23-L24 which `pin-project-lite` doesn't support. 

I think it may be possible to refactor the implementation somewhat so that the drop impl is not needed. We could probably add a wrapper struct around just the semaphore with a `Drop` impl, instead.

Michael-J-Ward:
Started the grind-work [part of this migration](https://github.com/Michael-J-Ward/tower/tree/migrate-to-pin-project-lite).

Two limitations of `pin-project-lite` that I ran into:

1) [does not support tuple structs / variants - which is labelled as won't fix.](https://github.com/taiki-e/pin-project-lite/blob/416be96f7777862c68b567c92a91887f69a8c2b3/src/lib.rs#L401-L408)
2) [Cannot interoperate with other field attributes](https://github.com/taiki-e/pin-project-lite/issues/3) including doc comments

This means each `StateEnum` needs a small refactor [like this one](https://github.com/Michael-J-Ward/tower/commit/dc42c2567d515a111f8d6699ebaecbcb49e48624).

taiki-e:
> which `pin-project-lite` doesn't support.

No. I recently added support for it: https://github.com/taiki-e/pin-project-lite/releases/tag/v0.2.7

Michael-J-Ward:
And [implemented for buffer::Worker](https://github.com/tower-rs/tower/pull/595/commits/ade48a255311b124163f32cf8242e0fba70fc12a)

@taiki-e is there anything preventing trait bounds on the `PinnedDrop` impl

taiki-e:
https://github.com/tower-rs/tower/pull/595/commits/ade48a255311b124163f32cf8242e0fba70fc12a

> However, it does not support generic trait bounds on the PinnedDrop impl.
>
> To workaround this, I removed the T::Error bound from the Worker struct definition,
> and moved `close_semaphore` to a a new impl without that trait bound.

Hmm, that's odd... I'll take a look later.

PR #637, opened 2022-02-08
util: rework `Either`

davidpdrsn:
In practice I've found `Either` be hard to use since it changes the error type to `BoxError`. That means if you combine two infallible services you get a service that, to the type system, is fallible. That doesn't work well with [axum's error handling](https://docs.rs/axum/latest/axum/error_handling/index.html) model which requires all services to be infallible and thus always return a response. So you end up having to add boilerplate just to please the type system.

Additionally, the fact that `Either` implements `Future` also means we cannot fully remove the dependency on `pin-project` since `pin-project-lite` doesn't support tuple enum variants, only named fields.

This PR reworks `Either` to address these:

- It now requires the two services to have the same error type so no type information is lost. I did consider doing something like `where B::Error: From<A::Error>` but I hope this simpler model will lead to better compiler errors.
- Changes the response future to be a struct with a private enum using `pin-project-lite`
- Removes the `Future` impl so we can remove the dependency on `pin-project`

Goes without saying that this is a breaking change so we have to wait until tower 0.5 to ship this.

cc @jplatte

Fixes https://github.com/tower-rs/tower/issues/594
Fixes https://github.com/tower-rs/tower/issues/550

### DIFF
--- a/tower/src/util/either.rs
+++ b/tower/src/util/either.rs
@@ -2,8 +2,7 @@
 //!
 //! See [`Either`] documentation for more details.
 
-use futures_core::ready;
-use pin_project::pin_project;
+use pin_project_lite::pin_project;
 use std::{
     future::Future,
     pin::Pin,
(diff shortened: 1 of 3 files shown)


jplatte:
Have you considered implementing `Layer` and `Service` for [futures_util::future::Either](https://docs.rs/futures-util/latest/futures_util/future/enum.Either.html) instead and using that type in the relevant places? That approach seemed a little bit cleaner to me when going over the possible solutions last time.

jplatte:
Additionally, IIRC there was some talk about changing `A` and `B` to `Left` and `Right`, which should maybe be part of the same PR?

davidpdrsn:
> Have you considered implementing Layer and Service for [futures_util::future::Either](https://docs.rs/futures-util/latest/futures_util/future/enum.Either.html) instead and using that type in the relevant places? That approach seemed a little bit cleaner to me when going over the possible solutions last time.

That would mean tower gets a public dependency on `futures_util` which I don't think we have today 🤔

> Additionally, IIRC there was some talk about changing `A` and `B` to `Left` and `Right`, which should maybe be part of the same PR?

Yeah that seems fine to me. I'll do that.

jplatte (approved):

tower/src/util/either.rs
```diff
@@ -17,58 +16,73 @@ use tower_service::Service;
...
+    }
+}
+
+pin_project! {
+    #[project = KindProj]
+    enum Kind<A, B> {
+        Left { #[pin] inner: A },
+        Right { #[pin] inner: B },
+    }
+}
```
hawkw, line 73: nit, not a big deal: can these be combined into a single `pin_project!` block?
```suggestion
pin_project! {
    /// Response future for [`Either`].
    pub struct EitherResponseFuture<A, B> {
        #[pin]
        kind: Kind<A, B>
    }

    #[project = KindProj]
    enum Kind<A, B> {
        Left { #[pin] inner: A },
        Right { #[pin] inner: B },
    }
}
```

tower/src/util/either.rs
```diff
@@ -17,58 +16,73 @@ use tower_service::Service;
 /// Both services must be of the same request, response, and error types.
```
hawkw, line 16: lol, this comment was previously incorrect, but it's correct now :P

hawkw:
> > Have you considered implementing Layer and Service for [futures_util::future::Either](https://docs.rs/futures-util/latest/futures_util/future/enum.Either.html) instead and using that type in the relevant places? That approach seemed a little bit cleaner to me when going over the possible solutions last time.
> 
> That would mean tower gets a public dependency on `futures_util` which I don't think we have today

FWIW, `tower` [does depend on `futures_util`](https://github.com/tower-rs/tower/blob/2f50b49f5af54dced258b2d781894848cc79fdfb/tower/Cargo.toml#L72), but it's not exposed in any public APIs as far as I know.

davidpdrsn:
Actually implementing `Service` and `Layer` directly for `futures_util::future::Either` would require putting those impls in `tower-service` and `tower-layer` and probably get in the way of making them 1.0, at least `futures_util` until 1.0. Dunno what the timeline is on that but I doubt it'll happen any time soon.

hawkw:
> Actually implementing `Service` and `Layer` directly for `futures_util::future::Either` would require putting those impls in `tower-service` and `tower-layer` and probably get in the way of making them 1.0, at least until `futures_util` itself is 1.0. Dunno what the timeline is on that but I doubt it'll happen any time soon.

Oh, you're right, I didn't think through the orphan rules here. I don't think we'll be able to do that, then.

We _could_ use `futures_util::future::Either` instead of the custom response future type, but I don't think it's worth making it public API for that purpose. The implementation is trivial enough that we don't _need_ to use it from `futures_util`, I think the only motivation would be compatibility, and compatibility would require putting it in `tower-service`, which I don't think we should do.

tower/src/util/either.rs
```diff
@@ -17,58 +16,73 @@ use tower_service::Service;
...
+                    inner: service.call(request),
+                },
+            },
         }
     }
 }
 
-impl<A, B, T, AE, BE> Future for Either<A, B>
+pin_project! {
+    /// Response future for [`Either`].
```
hawkw, line 60: take it or leave it:
```suggestion
    /// Response future returned by the [`Service`] implementation for [`Either`].
```
but on the other hand, maybe that's just needlessly wordy?

hawkw (approved):
seems good to me (in 0.5, of course)

tower/src/util/either.rs
davidpdrsn: Oh cool I didn't realize you could do this!

tower/src/util/either.rs
davidpdrsn: Its was eventually consistent 🤓

tower/src/util/either.rs
davidpdrsn: Yeah I think I'll just leave it as is.

olix0r:
I'm generally +1 on this. we should do 0.5 sooner than later...

The CI failure seems legit, though.

tower/src/util/either.rs
hawkw: i think i was actually wrong and you can't do this, this seems to have broken CI. my bad, sorry --- we should back this change out.

PR #639, opened 2022-02-11
tower: fix annoying clippy lints

hawkw:
This fixes a bunch of minor clippy lints. None of them were particularly
major, but I was getting tired of the warnings showing up in vscode.

The one lint that had to be ignored rather than fixed is the
`clippy::bool_assert_comparison` lint, which triggers on the
`tower_test::assert_request_eq!` macro. The lint triggers when writing
code like `assert_eq!(whatever, true)` rather than simply
`assert!(whatever)`. In this case, this occurs because the macro makes
an assertion about a request value, and in _some_ tests, the request
type is `bool`. We can't change this to use `assert!`, because in most
cases, when the request is not `bool`, we actually do need `assert_eq!`,
so I ignored that warning.

### DIFF
--- a/tower-test/src/macros.rs
+++ b/tower-test/src/macros.rs
@@ -34,8 +34,13 @@ macro_rules! assert_request_eq {
             Some(r) => r,
             None => panic!("expected a request but none was received."),
         };
-
-        assert_eq!(actual, $expect, $($arg)*);
+        // In some cases, this may be used with `bool` as the `Request` type, in
+        // which case, clippy emits a warning. However, this can't be changed to
+        // `assert!`, because the request type may *not* be `bool`...
+        #[allow(clippy::bool_assert_comparison)]
+        {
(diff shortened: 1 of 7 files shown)


hawkw:
The build failure appears to be due to a compiler regression (https://github.com/rust-lang/rust/issues/93821). I believe there's a PR fixing it (https://github.com/rust-lang/rust/pull/93893), but it's not in yesterday's nightly.

Because our CI is currently configured with fail-fast enabled for the cross-version test jobs, we don't get to see the results of the builds on versions other than nightly, so it's not obvious that this is a nightly-specific issue. I opened #640 to change the CI configuration to not do that.

I think we may also want to change the GitHub settings to _not_ require the nightly build in order to merge a PR, so that compiler regressions like this one don't block merging?

LucioFranco (approved):

PR #643, opened 2022-02-16
ci: automatically publish release notes to GitHub Releases

hawkw:
This branch adds a GitHub Actions workflow to automatically publish
release notes to GitHub Releases when a tag is pushed on the `master`
branch that corresponds to a release.

The release notes are parsed from the changelog using
taiki-e/create-release-action. The workflow will only run when a tag
matching `tower-[0-9].+` or `tower-[a-z]+-[0-9].+` is pushed to the
`master` branch on the origin (`tower-rs/tower`) repo.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,24 @@
+name: create github release
+
+on:
+  push:
+    tags:
+      - tower-v[0-9]+.*
+      - tower-[a-z]+-v[0-9]+.*
+
+jobs:
+  create-release:
+    name: Create GitHub release
+    # only publish from the origin repository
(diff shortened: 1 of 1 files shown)


davidpdrsn (approved):
Usually not a fan of automated release flows like this but up for giving it a shot.

hawkw:
@davidpdrsn for the record, _all_ this does is create a GitHub Release from a tag; my experience is that it's easy to forget to do that, so it's nice to automate that step.

hawkw:
(and, if it fails for any reason, the GitHub release can still be created manually)

davidpdrsn:
Ah yes true. All good!

Issue #636, opened 2022-02-02
path to tower-service 1.0


hawkw:
## Motivation

Tower has been used in production for over three years. Since the `Service` trait is intended to provide a common abstraction and compatibility interface across libraries, the definition of the `Service` trait is factored out into the `tower-service` crate so that it can remain stable even as other parts of the `Tower ecosystem are changed. A significant number of major crates [depend on `tower-service`](https://lib.rs/crates/tower-service/rev), including `hyper`, `tonic`, `warp`, and `axum`. In most cases, these are _public API dependencies_. This means that these crates cannot currently release 1.0 versions while they expose `tower-service` in their public APIs.

The last time a breaking change was made to the `Service` trait was [November 2019](https://github.com/tower-rs/tower/blob/master/tower-service/CHANGELOG.md#031-november-29-2019); this was in order to transition from `futures` 0.1 to `std::future`. While there have been a number of proposed improvements to the `Service` trait (e.g.  #110, #136, #137, #319, #408, #412, #626), most of these have not ended up being practical, or require future language features that are not currently available on stable Rust.

Therefore, I think it's time to have a serious conversation about stabilizing the `Service` trait.

## Proposal

In order to stabilize `tower-service`, I think we should do the following:
1. **Set an approximate time-frame for a stable release**. If possible, coordinate with other libraries that have public API dependencies on `tower-service` and wish to release 1.0 versions (e.g. `hyper`).
2. **Thoroughly investigate proposed `Service` trait changes**. Before stabilizing `Service` *exactly* as it exists right now, we should look into all the potential breaking changes that have been suggested, investigate whether they are possible, and assess the potential benefits of making them. This includes ideas like [`Service::disarm`](#408), [tokens](https://github.com/tower-rs/tower/pull/631), [splitting `Service` into `PollReady` and `Call` traits](https://github.com/tower-rs/tower/issues/626#issuecomment-1009256748), etc. I think the best way to assess these ideas is to actually implement them in a branch. To really determine the benefits and drawbacks of these changes, it would probably also be good to fork a crate that _uses_ `tower-service`, and try to update it to use the modified version. If this isn't practical, we should at least try to update a non-trivial example project; at the very least, we should be able to test out `tower::balance` and get the [load-balancing example](https://github.com/tower-rs/tower/blob/master/tower/examples/tower-balance.rs) to work.
[...]

PR #650, opened 2022-03-09
service: improve unsized types' support

cppforliving:
While `Service` impl for `Box<S>` covers unsized `S` case, the impl for `&mut S` is not, this PR is an attempt to fix that.

Also there are a few places in `Service` docs where the `poll_ready` call seems to be missing.


davidpdrsn:
I'm curious what your use case for this is 🤔

I've personally considered removing this blanket impl and instead have a blanket impl for async functions so we wouldn't need `service_fn`.

cppforliving:
Not that I particularly need this, I just spotted https://github.com/tower-rs/tower/issues/636 so I looked at tower-service API and noticed this minor asymmetry.

I really like the idea of adding blanket impl for async functions, however it seems that removing existing blanket impls would require reworking `tower::util::Ready` as its current impl relies on the `&mut Service` blanket impl. Maybe a bit unrelated example, but `std::future::Future` has a blanket impls for `&mut Future` and `Box<Future>` which to me seems quite useful, so maybe it would be worthy to keep similar blanket impls for `tower::Service` as well, just thinking...

cppforliving:
Any thoughts on this?

davidpdrsn:
I filed a proposal for removing the old blanket impls https://github.com/tower-rs/tower/pull/657

ripytide (approved):
I've read through this and #657. From my understanding this PR makes the blanket impl for `&mut S where S: Service` more generic. I think this is a good idea. Blanket trait impls should be as generic as they can be, part of the robustness principle.

I don't why #657 and adding a blanket `Service` trait impl for `Fn()` types would mean we wouldn't want to offer these reference type impls too. Again part of the "trait impls should be as generic as possible" principle.

The doc example changes make sense to me too. We should definitely not be showing examples that breach the "wait for ready before calling" API contract of `Service`.

jplatte (approved):

PR #654, opened 2022-03-11
buffer: change `Buffer` to be generic over inner service's `Future`

### LINKED ISSUE #641: `buffer::Service` should probably erase the inner service type
Currently, `tower::buffer::Service` is parameterized over the type of the inner `Service` inside the buffer worker. However, the `Service` struct does not actually need to know the type of the inner service, only the types of the inner service's `Future` and `Error` associated types. 

In many cases, the `Service` type will likely be a significantly longer type than the `Future` and `Error` types --- when the service is composed of a number of chained middleware services, its type may be quite complex. Although futures will also often be composed of many nested future types, these will generally not be quite as large as `Service` types, for a few reasons:
-  Some middleware `Service`s may not wrap the inner service's future, and simply perform logic in `call` and/or `poll_ready`. These `Service`s will be generic over the inner `Service` type, but will not wrap the `Future` type
- Some middleware may return boxed futures, erasing the inner `Future` type while still being generic over the inner `Service` type

Therefore, we may wish to consider changing the definition of the `buffer::Service` type to be generic over only the `Future` and `Error` types of the inner service. This could reduce type complexity that may impact compilation time (as well as binary size, if type names are ever displayed in the program). Of course, because this changes the type signature of `buffer::Worker`, this will have to wait until the next breaking change.

cc @olix0r, who initially suggested this.

hawkw:
This commit changes the `Buffer` service to be generic over the inner
service's `Future` type, rather than over the `Service` type itself. The
`Worker` type is still generic over the inner `Service` (and, it must
be, as it stores an instance of that service). This should reduce type
complexity of buffers a bit, as they will erase nested service types.

Unfortunately, `Buffer` is still generic over the inner service's
`Future`, which may also be nested, but it is likely to be less complex
than the `Service`. It would be nice if we could erase the `Future` type
as well, but I don't believe this is possible without either boxing the
futures or changing them to always be spawned on a background task,
neither of which seems desirable here.

Closes #641

### DIFF
--- a/tower/src/buffer/service.rs
+++ b/tower/src/buffer/service.rs
@@ -4,7 +4,10 @@ use super::{
     worker::{Handle, Worker},
 };
 
-use std::task::{Context, Poll};
+use std::{
+    future::Future,
+    task::{Context, Poll},
+};
 use tokio::sync::{mpsc, oneshot};
 use tokio_util::sync::PollSender;
 use tower_service::Service;
@@ -13,18 +16,14 @@ use tower_service::Service;
 ///
 /// See the module documentation for more details.
 #[derive(Debug)]
-pub struct Buffer<T, Request>
-where
-    T: Service<Request>,
-{
-    tx: PollSender<Message<Request, T::Future>>,
+pub struct Buffer<Request, F> {
+    tx: PollSender<Message<Request, F>>,
     handle: Handle,
 }
 
-impl<T, Request> Buffer<T, Request>
+impl<Request, F> Buffer<Request, F>
 where
-    T: Service<Request>,
-    T::Error: Into<crate::BoxError>,
+    F: 'static,
 {
     /// Creates a new [`Buffer`] wrapping `service`.
     ///
@@ -47,11 +46,11 @@ where
     /// [`Poll::Ready`]: std::task::Poll::Ready
     /// [`call`]: crate::Service::call
     /// [`poll_ready`]: crate::Service::poll_ready
-    pub fn new(service: T, bound: usize) -> Self
+    pub fn new<T>(service: T, bound: usize) -> Self
     where
-        T: Send + 'static,
-        T::Future: Send,
-        T::Error: Send + Sync,
+        T: Service<Request, Future = F> + Send + 'static,
+        F: Send,
+        T::Error: Into<crate::BoxError> + Send + Sync,
         Request: Send + 'static,
     {
         let (service, worker) = Self::pair(service, bound);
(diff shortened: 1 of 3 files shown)


tower/src/buffer/service.rs
```diff
@@ -47,11 +46,11 @@ where
...
     /// [`poll_ready`]: crate::Service::poll_ready
-    pub fn new(service: T, bound: usize) -> Self
+    pub fn new<T>(service: T, bound: usize) -> Self
     where
-        T: Send + 'static,
-        T::Future: Send,
-        T::Error: Send + Sync,
+        T: Service<Request, Future = F> + Send + 'static,
+        F: Send,
+        T::Error: Into<crate::BoxError> + Send + Sync,
```
olix0r, line 53: I wonder if it's more or less onerous to take a `thiserror` dependency and explicitly handle the error cases than requiring boxing. Note, for instance, that `anyhow::Error` does not impl `Into<BoxError>`; and we don't really have any requirements internally (i.e. on the `Service trait` that `T::Error` impl `std::error::Error`.

tower/src/buffer/service.rs
```diff
@@ -47,11 +46,11 @@ where
     /// [`Poll::Ready`]: std::task::Poll::Ready
     /// [`call`]: crate::Service::call
     /// [`poll_ready`]: crate::Service::poll_ready
-    pub fn new(service: T, bound: usize) -> Self
+    pub fn new<T>(service: T, bound: usize) -> Self
```
olix0r, line 49: Controversial opinion: but I generally prefer using `S` for service generics.

tower/src/buffer/service.rs
```diff
@@ -85,16 +84,15 @@ where
     }
 }
 
-impl<T, Request> Service<Request> for Buffer<T, Request>
+impl<Request, F, Resp, Error> Service<Request> for Buffer<Request, F>
```
olix0r, line 87: I know this is pre-existing, but I greatly prefer abbreviations for generics: `Req` & `Rsp`. Definitely more a personal taste thing than a blocker; but any time I see `Request` I assume there's an actual concrete type we're handling.

olix0r (commented):
Generally +1 on the change; some small readability quibbles

tower/src/buffer/service.rs
hawkw: sure, happy to change that --- i was just going with the existing naming. i agree that `Request` feels like a concrete name to me.

tower/src/buffer/service.rs
hawkw: could be worth considering but I think we should address that in a separate PR?

hawkw: opened #655 for that!

hawkw:
I think this is probably ready to merge, if anyone wants to give it another look?

davidpdrsn (approved):
Noise!

PR #656, opened 2022-03-14
util: Fix call_all hang when stream is pending

leoyvens:
Currently `call_all` will hang in a busy loop if called when the input stream is pending.

### DIFF
--- a/tower/src/util/call_all/common.rs
+++ b/tower/src/util/call_all/common.rs
@@ -111,6 +111,7 @@ where
                 Poll::Pending => {
                     // TODO: We probably want to "release" the slot we reserved in Svc here.
                     // It may be a while until we get around to actually using it.
+                    return Poll::Pending;
                 }
             }
         }
--- a/tower/src/util/call_all/common.rs
+++ b/tower/src/util/call_all/common.rs
@@ -97,21 +97,18 @@ where
                 .expect("Using CallAll after extracing inner Service");
             ready!(svc.poll_ready(cx))?;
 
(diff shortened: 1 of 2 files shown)


tower/src/util/call_all/common.rs
```diff
@@ -111,6 +111,7 @@ where
                 Poll::Pending => {
                     // TODO: We probably want to "release" the slot we reserved in Svc here.
                     // It may be a while until we get around to actually using it.
+                    return Poll::Pending;
```
hawkw, line 114: nit, take it or leave it: now that we return `Poll::Pending` in this case, we could simplify the nested `match` to just a single `match` by using `ready!` instead:
```rust

            // If it is, gather the next request (if there is one), or return `Pending` if the
            // stream is not ready.
            // TODO: We probably want to "release" the slot we reserved in Svc if the
            // stream returns `Pending`. It may be a while until we get around to actually
            // using it.
            match ready!(this.stream.as_mut().poll_next(cx)) {
                Some(req) => {
                    this.queue.push(svc.call(req));
                }
                None => {
                    // We're all done once any outstanding requests have completed
                    *this.eof = true;
                }
            }
 ```

leoyvens: Thank you for the review, I've done this refactor.

hawkw (approved):
looks great, thank you!

leoyvens:
@hawkw Thanks for the review! I forced push so this still needs a merge.

PR #660, opened 2022-04-13
log: don't enable this feature by default

Noah-Kennedy:
Unfortunately, tracing/log is a non-additive feature (https://github.com/tokio-rs/tracing/issues/1793), so enabling it is a bit radioactive.

### DIFF
--- a/tower/Cargo.toml
+++ b/tower/Cargo.toml
@@ -23,7 +23,6 @@ edition = "2018"
 rust-version = "1.49.0"
 
 [features]
-default = ["log"]
 
 # Internal
 __common = ["futures-core", "pin-project-lite"]


davidpdrsn (approved):
LGTM!

@hawkw what do you think?

hawkw (approved):
I think that _most_ people using Tower are probably also using `tracing` at this point, so I'm fine with removing the `log` support.

However, there's definitely an argument that this should be considered a breaking change. Although it shouldn't cause compilation to break, it _does_ mean that `log` records from `tower` may suddenly disappear for any users who are (a) using `log` and (b) not enabling the `tracing/log` feature themselves. On the other hand...I don't actually care very much. I just thought this was worth bringing up.

PR #661, opened 2022-05-27
load_shed: make constructor for Overloaded error public

mattklein123:
This allows for mocking. This also matches what is done for
the timeout Elapsed error.

Signed-off-by: Matt Klein <mklein@lyft.com>

### DIFF
--- a/tower/src/load_shed/error.rs
+++ b/tower/src/load_shed/error.rs
@@ -7,6 +7,7 @@ use std::fmt;
 /// called.
 ///
 /// [`LoadShed`]: crate::load_shed::LoadShed
+#[derive(Default)]
 pub struct Overloaded {
     _p: (),
 }
--- a/tower/src/load_shed/error.rs
+++ b/tower/src/load_shed/error.rs
@@ -12,7 +12,8 @@ pub struct Overloaded {
 }
 
 impl Overloaded {
-    pub(crate) fn new() -> Self {
+    /// Construct a new overloaded error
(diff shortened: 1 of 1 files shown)


mattklein123:
Any chance of getting this reviewed/merged? Thank you!

LucioFranco (approved):
Sorry for the delay on this! I was out on vacation :D 

Generally, we don't want to expose these constructors but we already do this for the `Elapsed` error in the timeout layer so I am fine with merging this.

tower/src/load_shed/error.rs
```diff
@@ -12,7 +12,8 @@ pub struct Overloaded {
 }
 
 impl Overloaded {
-    pub(crate) fn new() -> Self {
+    /// Construct a new overloaded error
+    pub fn new() -> Self {
```
hawkw, line 16: tiny style nit: if we're going to have a public 0-argument `new`, we may as well also implement `Default`...

hawkw (approved):
This seems fine to me.

In general, we try to avoid exposing error type constructors when they're not necessary, as it poses a forward-compatibilty issue (we may want to add additional data to the error in the future), but in this case, it seems pretty unlikely, and if it's useful for testing, we may as well make it public.

tower/src/load_shed/error.rs
mattklein123: Updated, thanks for reviewing.

mattklein123:
Awesome thank you for the merge!

PR #662, opened 2022-05-27
service: clarify docs around shared resource consumption in poll_ready()

mattklein123:
Signed-off-by: Matt Klein <mklein@lyft.com>


hawkw (approved):
this looks good to me, thanks!

PR #667, opened 2022-06-16
ready_cache: just use pin_project for `Pending`

hawkw:
This gets rid of the `Unpin` impl with the weird comment on it.

Alternatively, we could just put a `S: Unpin` bound on `Pending`, but
this changes the public API to require that the service type is `Unpin`.
In practice, it will be, but we could also just avoid the trait bound.


LucioFranco:
I don't understand the motivation for this change? Is it to just remove the unsafe impl?

PR #670, opened 2022-06-17
ci: run MSRV checks with minimal dep versions

hawkw:
In many cases, new releases of a dependency can break compatibility with
Tower's minimum supported Rust version (MSRV). It shouldn't be necessary
for Tower to bump its MSRV when a dependency does, as users on older
Rust versions should be able to depend on older versions of that crate.
Instead, we should probably just run our MSRV checks with minimal
dependency versions.

This branch changes Tower's CI jobs to do that.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -6,24 +6,40 @@ on:
     - master
   pull_request: {}
 
+env:
+  MSRV: 1.40.0
+
 jobs:
-  check:
+  check-stable:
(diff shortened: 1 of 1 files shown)


olix0r (approved):

hawkw:
Oh. Nevermind. We can't do this, because `-Z minimal-versions` is an unstable cargo feature and requires a nightly toolchain (https://github.com/taiki-e/cargo-minimal-versions#prerequisites). Hmm...

PR #676, opened 2022-07-02
util: add `Wrap`

lilyball:
This is middleware that wraps a service with both pre and post functions. It offers two flavors, one which is purely for passing shared state from the pre function to the post, and the other of which allows for asynchronous pre/post functions, mapping the request type, and short-circuiting so the inner service isn't called.

This is something I've wanted for a while, but it took a lot of experimentation to figure out how to actually express this in a way that was both generic and compiled. This is something that I always felt was missing functionality, because it turns out to be fairly common in my experience to want to take some service and wrap it in arbitrary middleware that needs to share state across the call to the inner service.

Some common uses of this are to copy data from the request to the response or to stick a cache in front of a service.

I do think this could probably stand to have some examples added to demonstrate using it, but I didn't want to make the documentation too complex. I'm also open to the idea of flipping the names of `decorate` and `wrap`, I just figured that `wrap` was the more obvious name to go for and so that should be the one with simpler usage.

### DIFF
--- a/tower/CHANGELOG.md
+++ b/tower/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 # Unreleased
 
+### Added
+
+- **util**: Add `Wrap` for wrapping a `Service` with pre/post functions
+
 ### Changed
 
 - **util**: Removed deprecated `ServiceExt::ready_and` method and `ReadyAnd`
--- a/tower/src/builder/mod.rs
+++ b/tower/src/builder/mod.rs
@@ -826,9 +826,13 @@ impl<L> ServiceBuilder<L> {
     /// returned by the inner service, `T` is the shared state provided by `pre`, and `Response`
     /// and `Err` match the types used in `pre`. The returned [`Result`] is the overall result from
     /// the wrapped service.
+    /// 
+    /// See the documentation for the [`decorate` combinator][] for examples.
     ///
     /// See also [`wrap`](Self::wrap) for when you just need to share state across the inner
     /// service and don't need asynchronous functions or short-circuiting.
+    /// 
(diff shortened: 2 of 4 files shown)


lilyball:
More succinctly, this handles the problem of wrapping a service with `service_fn()` not handling backpressure from the wrapped service.

lilyball:
FWIW I just prototyped an alternative solution that looks like

```rust
fn service_fn2<T, CallF, PollF>(state: T, poll_ready: PollF, call: CallF) -> ServiceFn2<T, CallF, PollF> { … }

impl<T, CallF, PollF, Fut, Request, R, E> Service<Request> for ServiceFn2<T, CallF, PollF>
where
    CallF: FnMut(&mut T, Request) -> Fut,
    PollF: FnMut(&mut T, &mut Context<'_>) -> Poll<Result<(), E>>,
    Fut: Future<Output = Result<R, E>>,
{ … }
```

and while it's technically doable, we get really awkward errors with the compiler and have to write out type signatures everywhere when using it. I assume this is at least partially due to not being able to specify bounds in the `service_fn2` signature, but adding bounds there constrains the function (in particular, it would prevent using this to return a service that is generic over its request type, which may be a problem if you're trying to wrap an inner service that is generic).

lilyball:
I pushed examples for `wrap` and `decorate` (along with some other fixes). `wrap` is a nice simple example, but I'm not sure how compelling the `decorate` example is. I just know that trying to add a cache in front of a service was something I really struggled with before (and ended up writing a custom service for before writing `decorate`) and it feels like something that should be possible without writing custom types.

lilyball:
I was reminded about this today. I still think it'd be nice to land this, though I haven't been using `tower` myself in a while. I just skimmed through it though and the thing that jumps out at me is that `decorate` should probably be tweaked to use `ControlFlow` instead of the outer `Result`.

PR #677, opened 2022-07-07
feat(timeout): add a timeout which can also timeout on poll_ready

bnjjj:
Following a discussion on [Discord](https://discord.com/channels/500028886025895936/627651185406509067/992057868289310761) I mentioned it was impossible to use `Timeout` with `RateLimit` because `RateLimit` is currently waiting in the `poll_ready`  but the current `Timeout` implementation doesn't care at all about `poll_ready` call to inner service. The main goal of this PR is to add the ability to set a timeout for both call to `poll_ready` and `call` to be sure we can in case of a wait in `poll_ready` be able to interrupt/timeout the current request.

I'm not good to find good names so if you have better suggestions than just `GlobalTimeout` feel free :)

### DIFF
--- a/tower/CHANGELOG.md
+++ b/tower/CHANGELOG.md
@@ -11,9 +11,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - **util**: Removed deprecated `ServiceExt::ready_and` method and `ReadyAnd`
   future ([#652])
+- **retry**: **Breaking Change** `retry::Policy::retry` now accepts `&mut Req` and `&mut Res` instead of the previous mutable versions. This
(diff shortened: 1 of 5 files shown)


tobz:
I realize this is a two-year old PR, but the sake of reviewing it and providing some feedback...

I can definitely understand the goal here, although this strikes me as kind of a footgun, since a lot of `Service` implementations treat `poll_ready` returning an error as a terminal state for the service (i.e. they fuse and can no longer be called). Wrapping the overall call to the service in a timeout would accomplish the same thing, since it can be wrapped around both the waiting for the service to become ready and the call itself.

I _think_ it could potentially be cleaner if this was a helper method on `ServiceExt` that provided the timeout-wrapped-over-the-poll-ready-call behavior, so it could be named explicitly and make it clear we're waiting specifically for the service to become ready.

PR #681, opened 2022-08-03
retry: Change `Policy` to accept `&mut self`

LucioFranco:
This changes the `Policy` trait in the `retry` layer to accept `&mut
self` instead of `&self` and changes the output type of the returned
future to `()`. The motivation for this change is to simplify
the trait a bit. By the trait methods having mutable references it means
that for each retry request session you can mutate the local policy.
This is because the policy is cloned for each individual request that
arrives into the retry middleware. In addition, this allows the `Policy`
trait to be object safe.

cc @rcoh

### DIFF
--- a/tower/src/retry/future.rs
+++ b/tower/src/retry/future.rs
@@ -64,7 +64,7 @@ where
 impl<P, S, Request> Future for ResponseFuture<P, S, Request>
 where
     P: Policy<Request, S::Response, S::Error>,
-    S: Service<Request> + Clone,
+    S: Service<Request>,
 {
     type Output = Result<S::Response, S::Error>;
 
--- a/tower/src/retry/future.rs
+++ b/tower/src/retry/future.rs
@@ -63,7 +63,7 @@ where
 
 impl<P, S, Request> Future for ResponseFuture<P, S, Request>
 where
-    P: Policy<Request, S::Response, S::Error> + Clone,
+    P: Policy<Request, S::Response, S::Error> + Clone + Unpin,
     S: Service<Request> + Clone,
 {
     type Output = Result<S::Response, S::Error>;
@@ -76,7 +76,7 @@ where
                 StateProj::Called { future } => {
                     let mut result = ready!(future.poll(cx));
                     if let Some(req) = &mut this.request {
-                        match this.retry.policy.retry(req, &mut result) {
+                        match this.retry.as_mut().project().policy.retry(req, &mut result) {
                             Some(checking) => {
                                 this.state.set(State::Checking { checking });
                             }
@@ -88,11 +88,8 @@ where
                     }
                 }
                 StateProj::Checking { checking } => {
-                    this.retry
-                        .as_mut()
-                        .project()
-                        .policy
-                        .set(ready!(checking.poll(cx)));
+                    ready!(checking.poll(cx));
+
                     this.state.set(State::Retrying);
                 }
                 StateProj::Retrying => {
--- a/tower/src/retry/future.rs
+++ b/tower/src/retry/future.rs
@@ -63,7 +63,7 @@ where
 
 impl<P, S, Request> Future for ResponseFuture<P, S, Request>
 where
-    P: Policy<Request, S::Response, S::Error> + Clone,
+    P: Policy<Request, S::Response, S::Error>,
     S: Service<Request> + Clone,
 {
     type Output = Result<S::Response, S::Error>;
--- a/tower/src/retry/future.rs
+++ b/tower/src/retry/future.rs
@@ -34,11 +34,11 @@ pin_project! {
             future: F
         },
         // Polling the future from [`Policy::retry`]
-        Checking {
+        Waiting {
             #[pin]
-            checking: P
+            waiting: P
         },
-        // Polling [`Service::poll_ready`] after [`Checking`] was OK.
+        // Polling [`Service::poll_ready`] after [`Waiting`] was OK.
         Retrying,
     }
 }
@@ -77,8 +77,8 @@ where
                     let mut result = ready!(future.poll(cx));
                     if let Some(req) = &mut this.request {
                         match this.retry.policy.retry(req, &mut result) {
-                            Some(checking) => {
-                                this.state.set(State::Checking { checking });
+                            Some(waiting) => {
+                                this.state.set(State::Waiting { waiting });
                             }
                             None => return Poll::Ready(result),
                         }
@@ -87,8 +87,8 @@ where
                         return Poll::Ready(result);
                     }
                 }
-                StateProj::Checking { checking } => {
-                    ready!(checking.poll(cx));
+                StateProj::Waiting { waiting } => {
+                    ready!(waiting.poll(cx));
 
                     this.state.set(State::Retrying);
                 }
--- a/tower/src/retry/future.rs
+++ b/tower/src/retry/future.rs
@@ -63,7 +63,7 @@ where
 
 impl<P, S, Request> Future for ResponseFuture<P, S, Request>
 where
-    P: Policy<Request, S::Response, S::Error> + Clone + Unpin,
+    P: Policy<Request, S::Response, S::Error> + Clone,
     S: Service<Request> + Clone,
 {
     type Output = Result<S::Response, S::Error>;
@@ -76,7 +76,7 @@ where
                 StateProj::Called { future } => {
                     let mut result = ready!(future.poll(cx));
                     if let Some(req) = &mut this.request {
-                        match this.retry.as_mut().project().policy.retry(req, &mut result) {
+                        match this.retry.policy.retry(req, &mut result) {
                             Some(checking) => {
                                 this.state.set(State::Checking { checking });
                             }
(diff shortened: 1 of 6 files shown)


LucioFranco:
cc @olix0r

tower/src/retry/future.rs
```diff
@@ -63,7 +63,7 @@ where
 
 impl<P, S, Request> Future for ResponseFuture<P, S, Request>
 where
-    P: Policy<Request, S::Response, S::Error> + Clone,
+    P: Policy<Request, S::Response, S::Error> + Clone + Unpin,
```
seanmonstar, line 66: Not that I assume `Policy`s will be require to be pinned normally, but I'm curious why this change. Isn't it fine to just project like we did before?

seanmonstar (commented):
I think these changes are likely the right way to go. I'd prefer to get `r+`s from @olix0r and @rcoh, since they have impls that would need to be changed.

tower/src/retry/future.rs
LucioFranco: Removed in https://github.com/tower-rs/tower/pull/681/commits/416e323024b0f7a21ca545ccee1915cf3c7bd034

tower/src/retry/policy.rs
```diff
@@ -37,14 +38,14 @@ use std::future::Future;
...
 /// }
 /// ```
-pub trait Policy<Req, Res, E>: Sized {
+pub trait Policy<Req, Res, E> {
     /// The [`Future`] type returned by [`Policy::retry`].
-    type Future: Future<Output = Self>;
+    type Future: Future<Output = ()>;
 
     /// Check the policy if a certain request should be retried.
     ///
```
rcoh, line 51: some docs here need to be updated

rcoh (approved):
I think the docs will need a close pass to make sure they all get updated, otherwise LGTM! the one thing I _might_ consider is making the returned type `T=()` I'm thinking about token bucket use cases where the policy might want to take and then return a token or something...but it's not immediately obvious to me how that would get wired

olix0r:
I'm a little unclear on this:

> By the trait methods having mutable references it means
that for each retry request session you can mutate the local policy.
This is because the policy is cloned for each individual request that
arrives into the retry middleware

So, we clone the policy for each request AND it is mutable? How does that work? If it's cloned for each request, we'd need an `Arc<Mutex<..>>` for state to be held across all clones? Or is the mutability only intended to support mutating a single cloned replica?

LucioFranco:
@olix0r By each request I mean request session, maybe that is a better way to word it. Aka we only clone the policy once iirc. So if you want data to be shared across all instances of the retry middleware then you'd need to `Arc<Mutex<..>`.

tower/src/retry/policy.rs
LucioFranco: Docs have been updated

tower/src/retry/mod.rs
```diff
@@ -17,9 +17,20 @@ pin_project! {
...
     ///
     /// A [`Policy`] classifies what is a "failed" response.
+    ///
+    /// # Clone
+    ///
+    /// This middleware requires that the `Service` is `Clone` due to requirements
+    /// of `'static` futures. To easily add `Clone` to your service you can
+    /// use the `Buffer` middleware.
+    ///
+    /// The `Policy` is also required to implement `Clone`. This middleware will
```
LucioFranco, line 27: Hopefully this and the removal of the `Clone` and the relaxation of the bounds on `ResponseFuture` make this more clear.

LucioFranco:
> the one thing I might consider is making the returned type T=() I'm thinking about token bucket use cases where the policy might want to take and then return a token or something...but it's not immediately obvious to me how that would get wired

@rcoh I would imagine you can do anything with` &mut self`? and `Arc` if you need it shared.

LucioFranco:
Ok this is ready for another round of reviews, CI is only failing due to a rustc beta/nightly bug.

tower/src/retry/mod.rs
```diff
@@ -17,9 +17,20 @@ pin_project! {
...
+    /// clone the policy for each _request session_. This means one instance
+    /// of the policy will exist for one request and its subsequent retries only.
+    /// This means the mutable reference is only for that session and if you want
+    /// to share data between request sessions then you will need to `Arc<Mutex<...>`.
```
hawkw, line 31: how about:
```suggestion
    /// clone the policy for each _request session_. This means a new clone
    /// of the policy will be created for each initial request and any subsequent 
    /// retries of that request. Therefore, any state stored in the `Policy` instance
    /// is for that request session only. In order to share data across request 
    /// sessions, that shared state may be stored in an [`Arc`], so that all clones
    /// of the `Policy` type reference the same instance of the shared state.
    ///
    /// [`Arc`]: std::sync::Arc
```

tower/src/retry/mod.rs
```diff
@@ -17,9 +17,20 @@ pin_project! {
     /// Configure retrying requests of "failed" responses.
     ///
     /// A [`Policy`] classifies what is a "failed" response.
+    ///
+    /// # Clone
+    ///
+    /// This middleware requires that the `Service` is `Clone` due to requirements
+    /// of `'static` futures. To easily add `Clone` to your service you can
+    /// use the `Buffer` middleware.
```
hawkw, line 25: some minor docs nits:
 - "due to requirements of `'static` futures" feels kind of vague to me --- will the reader understand what that means here?
 - in general, we tend not to use second-person ("you") language in API docs.

how about something like this:
```suggestion
    /// This middleware requires that the inner `Service` implements [`Clone`],
    /// because the `Service` must be stored in each [`ResponseFuture`] in
    /// order to retry the request in the event of a failure. If the inner
    /// `Service` type does not implement `Clone`, the [`Buffer`] middleware
    /// can be added to make any `Service` cloneable.
    ///
    /// [`Buffer`]: crate::buffer::Buffer
```

tower/src/retry/policy.rs
```diff
@@ -37,14 +38,14 @@ use std::future::Future;
...
+///     fn clone_request(&mut self, req: &Req) -> Option<Req> {
 ///         Some(req.clone())
 ///     }
 /// }
 /// ```
-pub trait Policy<Req, Res, E>: Sized {
+pub trait Policy<Req, Res, E> {
     /// The [`Future`] type returned by [`Policy::retry`].
-    type Future: Future<Output = Self>;
+    type Future: Future<Output = ()>;
```
hawkw, line 48: changing this to return a `Future<Output = ()>` and duplicating the `Policy` using `Clone` rather than having the future return a new policy definitely simplifies things, but it occurs to me that asynchronous work can no longer be performed in order to update the `Policy`. i can't _immediately_ come up with a reason that it would be necessary to generate the next `Policy` asynchronously, but it seems like it could be worth thinking about before we commit to this design...?

tower/src/retry/policy.rs
```diff
@@ -53,8 +54,9 @@ pub trait Policy<Req, Res, E>: Sized {
     ///
     /// If the request should **not** be retried, return `None`.
     ///
-    /// If the request *should* be retried, return `Some` future of a new
-    /// policy that would apply for the next request attempt.
+    /// If the request *should* be retried, return `Some` future that will delay
+    /// the next retry of the request. This can be used to sleep for a certain
+    /// duration or resolve right away.
```
hawkw, line 59: maybe:
```suggestion
    /// the next retry of the request. This can be used to sleep for a certain
    /// duration, to wait for some external condition to be met before retrying,
    /// or resolve right away, if the request should be retried immediately.
```

tower/src/retry/future.rs
```diff
@@ -88,11 +88,8 @@ where
                     }
                 }
                 StateProj::Checking { checking } => {
-                    this.retry
-                        .as_mut()
-                        .project()
-                        .policy
-                        .set(ready!(checking.poll(cx)));
+                    ready!(checking.poll(cx));
```
hawkw, line 91: nit, take it or leave it: is this state really "checking" if we can retry?  it seems like if we are in this state, we have already decided that the request can be retried, but we are _waiting_ before actually sending the retry request --- perhaps this should be renamed to `State::Waiting`?

not a blocker for this branch, since that naming was there before, but it might be worth changing...

hawkw (commented):
overall, I think this new design makes sense; &mdash; it's certainly simpler than the previous one. my one big question is whether there's a valid use case for generating the next `Policy` itself asynchronously that can't be satisfied with another approach. i don't really think there is, but perhaps others have input on that?

tower/src/retry/policy.rs
hawkw: thinking about it, i suppose that if an implementation needed to modify the `Policy` after the future's completion, it could always clone an `Arc`ed shared state into its `Future`...this may introduce *some* overhead over the current approach, but I don't really think that use case is common enough that it's particularly important to worry about...

LucioFranco: Yeah, I don't think that use case is something we should support with that level and anyways taking a lock on an uncontested arc/mutex should be very cheap and much cheaper than request io etc.

tower/src/retry/future.rs
LucioFranco: Good point, updated.

seanmonstar (approved):

PR #684, opened 2022-08-16
Moved docs from private `tower::util::boxed` module

### LINKED ISSUE #683: Docs for `BoxService` point at module level documentation which is missing
The [docs for `tower::util::BoxService`](https://docs.rs/tower/0.4.13/tower/util/struct.BoxService.html) state, in part, "See module level documentation for more details." But there is [no additional documentation of `BoxService` in that module](https://docs.rs/tower/0.4.13/tower/util/index.html).

Digging a little, I can see that this probably meant [the documentation in `tower::util::boxed`](https://github.com/tower-rs/tower/blob/tower-0.4.13/tower/src/util/boxed/mod.rs#L1-L33), but that module is not public.

dcormier:
Now they're on the public `BoxService` struct that referenced them.
    
Fixes #683.

### DIFF
--- a/tower/src/util/boxed/sync.rs
+++ b/tower/src/util/boxed/sync.rs
@@ -18,7 +18,31 @@ use std::{
 /// If you need a boxed [`Service`] that implements [`Clone`] consider using
 /// [`BoxCloneService`](crate::util::BoxCloneService).
 ///
-/// See module level documentation for more details.
+/// Dynamically dispatched [`Service`] objects allow for erasing the underlying
+/// [`Service`] type and using the `Service` instances as opaque handles. This can
+/// be useful when the service instance cannot be explicitly named for whatever
+/// reason.
+///
+/// # Examples
+///
+/// ```
+/// use futures_util::future::ready;
+/// # use tower_service::Service;
+/// # use tower::util::{BoxService, service_fn};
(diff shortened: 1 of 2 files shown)


tower/src/util/boxed/sync.rs
```diff
@@ -18,7 +18,7 @@ use std::{
 /// If you need a boxed [`Service`] that implements [`Clone`] consider using
 /// [`BoxCloneService`](crate::util::BoxCloneService).
 ///
-/// See module level documentation for more details.
+/// See [module level documentation](super) for more details.
```
dcormier, line 21: `BoxService` is available as `tower::util::BoxService` as well as `tower::util::boxed::BoxService`, but the referenced documentation is at `tower::util::boxed`. Linking like this goes to `tower::util::boxed` regardless of where the user found `BoxService`.

LucioFranco:
Hi, can you give more motivation on why this change? We don't normally expose the module but we make the types within it public at the `util` level.

LucioFranco:
Oh I think you linked the wrong issue #683

davidpdrsn:
Yeah it's to expose the module docs.

davidpdrsn (approved):

PR #685, opened 2022-08-19
retry: Add generic backoff utilities

LucioFranco:
This adds a new `Backoff` trait and a `ExponentialBackoff`
implementation borrwoed from `linkerd2-proxy`. This provides the initial
building blocks for a more fully batteries included retry policy.

cc @rcoh


tower/src/retry/backoff.rs
```diff
@@ -0,0 +1,228 @@
...
+//! included exponential backoff and jitter strategy.
+
+use rand::Rng;
+use rand::{rngs::SmallRng, thread_rng, SeedableRng};
+use std::fmt::Display;
+use std::future::Future;
+use std::time::Duration;
+use tokio::time;
+
+/// A backoff trait where a single mutable refernces represents a single
```
seanmonstar, line 17:
```suggestion
/// A backoff trait where a single mutable reference represents a single
```

tower/src/retry/backoff.rs
```diff
@@ -0,0 +1,228 @@
...
+/// backoff session. Implementors must also implement [`Clone`] which will
+/// reset the backoff back to the default state for the next session.
```
seanmonstar, line 19: `Clone` resets it to the default? That seems different than normal `Clone` behavior, and easy to get wrong. If I use a backoff a few times, and it's up to 2150ms, a normal clone with just give me the same thing at the same time.

tower/src/retry/backoff.rs
```diff
@@ -0,0 +1,228 @@
...
+    /// Must be greater than or equal to 0.0.
+    jitter: f64,
+
+    rng: R,
+    iterations: u32,
+}
+
+impl<R> ExponentialBackoff<R>
+where
+    R: Rng,
```
seanmonstar, line 49: I'd suggest people not depend on `rand::Rng`. It's fairly straightforward to make a simple internal RNG by making use of [`RandomState`](https://doc.rust-lang.org/std/collections/hash_map/struct.RandomState.html) and a counter.

tower/src/retry/backoff.rs
LucioFranco: Yeah, I originally had a `reset(&self) -> Self` which I could add back instead of  using clone. What do you think? or we could have a separate kinda factory trait. What do you think?

seanmonstar: To stay object-safe, a reset function could just mutate self to 0.

LucioFranco: The issue is that we need to basically have a factory to get owned copies of a backoff type that we can pass into the future. So we need some sort of way to share that. We could go with some sort of factory type but that is two traits which is more a pita. We could try `Stream` for this though.

seanmonstar: Could still ask that backoffs be clone...

Another option is to make the `next_backoff` return `Self`. It well encodes the meaning, but makes the trait object safe.

tower/src/retry/backoff.rs
```diff
@@ -0,0 +1,228 @@
...
+//! [`ExponentialBackoff`] implements [`Backoff`] and provides a batteries
+//! included exponential backoff and jitter strategy.
+
+use rand::Rng;
+use rand::{rngs::SmallRng, thread_rng, SeedableRng};
+use std::fmt::Display;
+use std::future::Future;
+use std::time::Duration;
+use tokio::time;
+
```
hawkw, line 16: the doc comment either for the module or the `Backoff` trait should _probably_ at least mention what the term "backoff" means, for readers who've never seen it before?

tower/src/retry/backoff.rs
hawkw: IMO, a `Backoff` trait which returns a `Stream` for each retry session is a cleaner design than one where cloning the `Backoff` type resets the backoff.

This would also allow more flexible user implementations; for example, a `tokio::time::Interval` could be returned as a backoff using the `tokio-stream` wrapper. Also, stream combinators could be used to modify an existing backoff: for example, if I wanted to do something every time a backoff completes (such as incrementing metrics), I could wrap an existing backoff implementation's return value in `Stream::inspect` or similar.

tower/src/retry/backoff.rs
```diff
@@ -0,0 +1,228 @@
...
+pub trait Backoff: Clone {
+    /// The future associated with each backoff. This usually will be some sort
+    /// of timer.
+    type Future: Future<Output = ()>;
+
+    /// Initiate the next backoff in the sequence.
+    fn next_backoff(&mut self) -> Self::Future;
+}
+
+/// A jittered exponential backoff strategy.
```
hawkw, line 29: let's at least briefly define what this actually does and why you would want it? maybe something like this:
```suggestion
/// A jittered [exponential backoff] strategy.
///
/// The backoff duration will increase exponentially for every subsequent
/// backoff, up to a maximum duration. A small amount of [random jitter] is
/// added to each backoff duration, in order to avoid retry spikes.
///
/// [exponential backoff]: https://en.wikipedia.org/wiki/Exponential_backoff
/// [random jitter]: https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/
```

tower/src/retry/backoff.rs
```diff
@@ -0,0 +1,228 @@
...
+    }
+}
+
+/// Backoff validation error.
+#[derive(Clone, Debug)]
+pub struct InvalidBackoff(&'static str);
+
+impl Display for InvalidBackoff {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{:?}", self)
```
hawkw, line 174: hmm, i don't think the `fmt::Display` implementation should just call the `fmt::Debug` implementation? maybe something like this:
```suggestion
        write!(f, "invalid backoff: {}", self.0)
```

tower/src/retry/backoff.rs
seanmonstar: I agree that's cleaner. But then we have to depend on `Stream`, which... I don't really want...

LucioFranco: My issue with stream is the requirement to implement stuff with `poll` which I want to avoid when you just need to work with Delay + Duration types which are much nicer to use than `poll` etc.

LucioFranco: Here is a proposal I have:

```rust

/// A backoff maker that can produce backoff types for retries.
pub trait MakeBackoff {
    /// Backoff returned from this maker.
    type Backoff: Backoff;

    /// Make a new backoff type that can be used for a single request
    /// retry strategy.
    fn make_backoff(&mut self) -> Self::Backoff;
}

/// A backoff trait where a single mutable reference represents a single
/// backoff session. Implementors must also implement [`Clone`] which will
/// reset the backoff back to the default state for the next session.
pub trait Backoff {
    /// The future associated with each backoff. This usually will be some sort
    /// of timer.
    type Future: Future<Output = ()>;

    /// Initiate the next backoff in the sequence.
    fn next_backoff(&mut self) -> Self::Future;
}
```

tower/src/retry/backoff.rs
LucioFranco: So one thing is how can we have deterministic testing of retries if `RandomState` is buried in this type?

tower/src/retry/backoff.rs
LucioFranco: https://github.com/tower-rs/tower/pull/685/commits/26fd2f8b7554aa9993216c874cd34c663af02eb2 adds this new trait

LucioFranco:
I think this is ready for another round of reviews!

tower/src/retry/backoff.rs
```diff
@@ -0,0 +1,263 @@
...
+            Duration::from_millis(u64::MAX),
+            0.99,
+            HasherRng::default(),
+        )
+        .expect("Unable to create ExponentialBackoff")
+    }
+}
+
+/// Backoff validation error.
+#[derive(Clone, Debug)]
```
seanmonstar, line 199: An error that implements Clone? Does it need to?

tower/src/retry/backoff.rs
```diff
@@ -0,0 +1,263 @@
...
+
+use std::fmt::Display;
+use std::future::Future;
+use std::time::Duration;
+use tokio::time;
+
+use crate::util::rng::{HasherRng, Rng};
+
+/// Trait used to construct [`Backoff`] trait implementors.
+pub trait MakeBackoff {
```
seanmonstar, line 21: I don't know why, something about having a second trait feels less ideal, but I admit I can't think of a better way (other than making Backoff return a new Backoff each time, but that makes it no longer object-safe...). :shrug:

seanmonstar (approved):
I think this looks good!

tower/src/retry/backoff.rs
LucioFranco: Yeah, I agree, I have the same feeling but this feels cleaner than any other approach I could find.

PR #686, opened 2022-08-23
util: Add `rng` utilities

LucioFranco:
This adds new PRNG utilities that only use libstd and not the external
`rand` crate. This change's motivation are that in tower middleware that
need PRNG don't need the complexity and vast utilities of the `rand`
crate.

This adds a `Rng` trait which abstracts the simple PRNG features tower
needs. This also provides a `HasherRng` which uses the `RandomState`
type from libstd to generate random `u64` values. In addition, there is
an internal only `sample_inplace` which is used within the balance p2c
middleware to randomly pick a ready service. This implementation is
crate private since its quite specific to the balance implementation.

The goal of this in addition to the balance middlware getting `rand`
removed is for the upcoming `Retry` changes. The `next_f64` will be used
in the jitter portion of the backoff utilities in https://github.com/tower-rs/tower/pull/685.

### DIFF
--- a/tower/src/util/rng.rs
+++ b/tower/src/util/rng.rs
@@ -1,4 +1,4 @@
-//! Utilities for generating random numbers.
+//! [PRNG] utilities for tower middleware.
 //!
 //! This module provides a generic [`Rng`] trait and a [`HasherRng`] that
 //! implements the trait based on [`RandomState`] or any other [`Hasher`].
@@ -6,6 +6,8 @@
 //! These utlities replace tower's internal usage of `rand` with these smaller
 //! more light weight methods. Most of the implemenations are extracted from
 //! their corresponding `rand` implementations.
+//! 
+//! [PRNG]: https://en.wikipedia.org/wiki/Pseudorandom_number_generator
 
 use std::{
     collections::hash_map::RandomState,
--- a/tower/src/util/rng.rs
+++ b/tower/src/util/rng.rs
@@ -3,8 +3,8 @@
 //! This module provides a generic [`Rng`] trait and a [`HasherRng`] that
 //! implements the trait based on [`RandomState`] or any other [`Hasher`].
 //!
-//! These utlities replace tower's internal usage of `rand` with these smaller
-//! more light weight methods. Most of the implemenations are extracted from
+//! These utlities replace tower's internal usage of `rand` with these smaller,
+//! more lightweight methods. Most of the implementations are extracted from
 //! their corresponding `rand` implementations.
 //!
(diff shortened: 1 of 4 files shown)


tower/src/util/rng.rs
```diff
@@ -0,0 +1,198 @@
...
+        HasherRng { hasher, counter: 0 }
+    }
+}
+
+impl<H> Rng for HasherRng<H>
+where
+    H: BuildHasher,
+{
+    fn next_u64(&mut self) -> u64 {
+        let mut hasher = self.hasher.build_hasher();
```
seanmonstar, line 105: I don't think you want to `build_hasher` for each number. That will restart the hasher back to the beginning. I'd expect you'd build a hasher once, and then feed the counter into it over and over. Or am I wrong?

tower/src/util/rng.rs
```diff
@@ -0,0 +1,198 @@
+//! [PRNG] utilities for tower middleware.
+//!
+//! This module provides a generic [`Rng`] trait and a [`HasherRng`] that
+//! implements the trait based on [`RandomState`] or any other [`Hasher`].
+//!
+//! These utlities replace tower's internal usage of `rand` with these smaller
+//! more light weight methods. Most of the implemenations are extracted from
```
hawkw, line 7:
```suggestion
//! more lightweight methods. Most of the implementations are extracted from
```

tower/src/util/rng.rs
```diff
@@ -0,0 +1,198 @@
+//! [PRNG] utilities for tower middleware.
+//!
+//! This module provides a generic [`Rng`] trait and a [`HasherRng`] that
+//! implements the trait based on [`RandomState`] or any other [`Hasher`].
+//!
+//! These utlities replace tower's internal usage of `rand` with these smaller
```
hawkw, line 6:
```suggestion
//! These utlities replace tower's internal usage of `rand` with these smaller,
```

tower/src/util/rng.rs
LucioFranco: I thought finished consumed it, but I guess finish just takes a &self. Let me see if that makes any meaningful difference.

seanmonstar: Yea, thinking more, I could be much more wrong than right.

seanmonstar (approved):

tower/src/util/rng.rs
LucioFranco: Follow up here HashMap creates a hasher for each hash so we are reverting back to that setup.

PR #690, opened 2022-09-12
retry: Extend `Policy` trait documentation

wutchzone:
Last week I was debugging while my `Retry` service was not behaving as expected. I found out in the source code that if my `Policy` implementation returns `None` in `clone_request`, then the retry logic will not be called. I would like to mention this in the interface to save time for the folks that could probably debug the same behavior.


LucioFranco (approved):
Thanks!

PR #693, opened 2022-09-25
chore: Use doc_auto_cfg

jplatte:
… so extra doc(cfg) attributes aren't needed in most places.


davidpdrsn (approved):
We've done this in axum for a while and its working great. So LGTM!

PR #694, opened 2022-09-28
layer: Implement Layer for tuples of up to 16 elements

jplatte:
cc @davidpdrsn @LucioFranco

### DIFF
--- a/tower-layer/src/lib.rs
+++ b/tower-layer/src/lib.rs
@@ -19,6 +19,7 @@
 mod identity;
 mod layer_fn;
 mod stack;
+mod tuple;
 
 pub use self::{
     identity::Identity,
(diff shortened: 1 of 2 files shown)


LucioFranco:
Seems really cool! I would love to seem some examples with it.

jplatte:
Okay, I updated the first commit to be more consistent (I had a few `TNs` instead of `LNs`) and then added a new commit for the reversed order. Like that, it's easy to reverse-apply the second commit to change the order later, if wanted 🙂

davidpdrsn:
Sweet! I can confirm that it matches `ServerBuilder` now.

jplatte:
Anything that still needs to happen before this can be merged?

davidpdrsn (approved):
Nope I think it's good to go :)

PR #701, opened 2022-10-13
discover: Implement Clone for Change

samvrlewis:
Implements Clone for discover::Change, if the underlying key and value both implement clone.

This is convenient for use-cases where a single change is duplicated so that it can be sent to multiple discover streams.

### DIFF
--- a/tower/src/discover/mod.rs
+++ b/tower/src/discover/mod.rs
@@ -97,7 +97,7 @@ where
 }
 
 /// A change in the service set.
-#[derive(Debug)]
+#[derive(Debug, Clone)]
 pub enum Change<K, V> {
     /// A new service identified by key `K` was identified.
     Insert(K, V),


tower/src/discover/mod.rs
```diff
@@ -104,3 +104,16 @@ pub enum Change<K, V> {
...
+    K: Clone,
+    V: Clone,
+{
+    fn clone(&self) -> Self {
+        match self {
+            Change::Insert(k, v) => Change::Insert(k.clone(), v.clone()),
+            Change::Remove(k) => Change::Remove(k.clone()),
+        }
+    }
+}
```
hawkw, line 119: This implementation seems like it could be derived?

hawkw (commented):
I think implementing `Clone` for this seems like a reasonable addition to the API.

Is there a reason that `Clone` was manually implemented rather than derived?

tower/src/discover/mod.rs
samvrlewis: Yep, you're right - I had some silly notions around deriving clone on enums with generics. Have fixed now, thanks!

hawkw (approved):
looks good to me, thanks!

PR #703, opened 2022-10-17
retry: Add `Budget` trait

boraarslan:
This PR adds `BudgetTrait` and `Bucket` traits to allow users to implement their budget and bucket implementations. The current `Budget` struct is changed to be generic over `Bucket`. The current `Bucket` implementation is also renamed as `TpsBucket`.

A new `SimpleBucket` struct is added. This bucket works by increasing the cost of the next retry on each retry request. Cost is then decreased gradually on each successful request.


tower/src/retry/budget.rs
```diff
@@ -89,21 +89,51 @@ use tokio::time::Instant;
 /// For more info about [`Budget`], please see the [module-level documentation].
 ///
 /// [module-level documentation]: self
-pub struct Budget {
-    bucket: Bucket,
+pub struct Budget<B = TpsBucket> {
```
LucioFranco, line 92: So I would not add a generic here and instead move this whole file to a new submodule and then have the main trait in the mod file.

tower/src/retry/budget.rs
```diff
@@ -89,21 +89,51 @@ use tokio::time::Instant;
...
+/// 
+/// For more info about [`Budget`], please see the [module-level documentation].
+/// For more info about [`Bucket`], see [bucket trait]
+/// 
+/// [module-level documentation]: self
+/// [bucket trait]: self::Bucket
 #[derive(Debug)]
-pub struct Overdrawn {
-    _inner: (),
+pub struct SimpleBucket {
```
LucioFranco, line 111: Instead of adding the new simple bucket here lets start smaller and just do the moving around for now. Then we can add a new impl in a future PR.

tower/src/retry/budget.rs
```diff
@@ -124,9 +154,93 @@ struct Generation {
...
+    /// and true is returned.
+    fn try_get(&self, amt: isize) -> bool;
+    /// Returns the amount of tokens in the reserve.
+    fn reserve(&self) -> isize;
+}
+
 // ===== impl Budget =====
 
 impl Budget {
+    /// Create a [`Budget`] that allows for a certain percent of the total
```
LucioFranco, line 196: nice!

LucioFranco (changes requested):
I think this is a great start, I left some feedback. I think we should break this into some smaller PRs so the diffs are easier to review and there is less changing.

tower/src/retry/budget/mod.rs
```diff
@@ -0,0 +1,116 @@
...
+#[allow(clippy::module_inception)]
+pub mod budget;
+
+pub use budget::Budget;
+pub use budget::TpsBucket;
+
+/// For more info about [`Budget`], please see the [module-level documentation].
+///
+/// [module-level documentation]: self
+pub trait BudgetTrait {
```
LucioFranco, line 83: I think this can just be Budget with the prev type renamed to `TpsBudget`. what do you think?

tower/src/retry/budget/mod.rs
```diff
@@ -0,0 +1,116 @@
...
+///
+/// A token bucket manages a reserve of tokens to decide if a retry for the request is
+/// possible. Successful requests put tokens into the reserve. Before a request is retried,
+/// bucket is checked to ensure there are sufficient amount of tokens available. If there are,
+/// specified amount of tokens are withdrawn.
+///
+/// For more info about [`Budget`], please see the [module-level documentation].
+///
+/// [module-level documentation]: self
+pub trait Bucket {
```
LucioFranco, line 105: I am not sure why we need a bucket trait? I imagine we only need the budget trait to abstract these two types we want to have? I don' think we need a second trait to represent internal implementation details? Happy to chat on discord about this if it doesn't make sense what I am saying.

tower/src/retry/budget/budget.rs
```diff
@@ -1,75 +1,4 @@
-//! A retry "budget" for allowing only a certain amount of retries over time.
```
LucioFranco, line 1: Why is this file not getting highlighted....

LucioFranco (changes requested):

tower/src/retry/budget/mod.rs
```diff
@@ -0,0 +1,116 @@
...
+//!         }
+//!     }
+//!
+//!     fn clone_request(&mut self, req: &Req) -> Option<Req> {
+//!         Some(req.clone())
+//!     }
+//! }
+//! ```
+
+#[allow(clippy::module_inception)]
```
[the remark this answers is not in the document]
LucioFranco: wait ya why do we have a mod budget inside the budget module already? shouldn't this mod be in lib.rs only?

tower/src/retry/budget/tps_budget.rs
```diff
@@ -20,25 +20,23 @@ use super::{Bucket, BudgetTrait};
 /// For more info about [`Budget`], please see the [module-level documentation].
 ///
 /// [module-level documentation]: super
-pub struct Budget<B = TpsBucket> {
-    bucket: B,
+pub struct TpsBudget {
+    bucket: TpsBucket,
```
LucioFranco, line 24: I would remove this extra struct and just merge everything into the TpsBucket and rename that TpsBudget. That should simplify some of this code and also make it less confusing since TpsBucket looks like TpsBudget lol

tower/src/retry/budget/tps_budget.rs
```diff
@@ -127,7 +125,7 @@ impl Budget {
     }
 }
 
-impl<B: Bucket> BudgetTrait for Budget<B> {
+impl Budget for TpsBudget {
```
LucioFranco, line 128: Just to check we should remove the deposit/withdraw methods from the base impl TpsBudget and only have the trait impl ones. I can't see it in the diff but just want to check.

boraarslan: Yes, I already removed them from the base impl.

LucioFranco (approved):
LGTM!

PR #708, opened 2022-11-01
util: Adds a BoxCloneServiceLayer

### LINKED ISSUE #707: Sharing Collections of Layers: BoxCloneLayer?
Building a big stack of layers sometimes results in pretty long and unwieldily types, even with only a few layers, e.g. `Stack<Either<TimeoutLayer, Identity>, Stack<Either<GlobalConcurrencyLimitLayer, Identity>, Stack<Either<LoadShedLayer, Identity>, Identity>>>`

When working with such layers, `BoxLayer` can be useful, but only if you need a `BoxService`. If you want to produce a `BoxCloneService`, you are out of luck. Should tower add a `BoxCloneLayer`? Am I approaching this all wrong? I want to be able to combine several layers and pass them to a `ServiceBuilder` (e.g. Layers with a common configuration source) without having to write & edit a huge type every time.

alexrudy:
This layer is similar to a BoxLayer, but produces a BoxCloneService instead, so can be used when the underlying layers must be clone.

Resolves #707

### DIFF
--- a/tower/src/util/boxed/layer_clone.rs
+++ b/tower/src/util/boxed/layer_clone.rs
@@ -1,0 +1,128 @@
+use crate::util::BoxCloneService;
+use std::{fmt, sync::Arc};
+use tower_layer::{layer_fn, Layer};
+use tower_service::Service;
+
+/// A [`Clone`] + [`Send`] boxed [`Layer`].
+///
+/// [`BoxCloneLayer`] turns a layer into a trait object, allowing both the [`Layer`] itself
+/// and the output [`Service`] to be dynamic, while having consistent types.
+///
+/// This [`Layer`] produces [`BoxCloneService`] instances erasing the type of the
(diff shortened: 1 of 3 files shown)


davidpdrsn (commented):
I agree this seems useful but I'm not a fan of the name `BoxCloneLayer`. It sounds like the layer is `Clone` but in fact it's the service it produces that is `Clone`.

`BoxLayer` is already `Clone`.

LucioFranco:
Should we name it CloneBoxLayer?

alexrudy:
Happy for whatever name you want. `BoxCloneServiceLayer` is another option, but that's a mouthful.

I will clarify in the docs for this type that BoxLayer is already clone, but that the _services_ it produces are not clone.

LucioFranco (approved):

LucioFranco:
Thanks!

PR #709, opened 2022-11-01
util: Two call_all bug fixes

### LINKED ISSUE #706: call_all should stop after poll_ready returns an error
The [docs](https://tower-rs.github.io/tower/tower/trait.Service.html#tymethod.poll_ready) say that a service should be discarded after an error is returned from `poll_ready`. But `CallAll` will currently treat a `poll_ready` error in the same way as a request error and continues to call the service. It seems to me that `CallAll` should instead return the error once and end the stream even if there are more requests to be made.

leoyvens:
One is handling poll_ready errors, resolves #706.

The other is fixing the TODO about disarming poll_ready, since there is no disarm this makes sure `poll_ready` is only called if `call` will immediately follow. This one actually hit me in production.

Each bug fix has a corresponding test case. This should not be breaking any public API.

### DIFF
--- a/tower/src/util/call_all/common.rs
+++ b/tower/src/util/call_all/common.rs
@@ -1,6 +1,7 @@
 use futures_core::{ready, Stream};
 use pin_project_lite::pin_project;
 use std::{
+    fmt,
     future::Future,
     pin::Pin,
     task::{Context, Poll},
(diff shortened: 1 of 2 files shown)


LucioFranco (approved):

leoyvens:
@LucioFranco Thank you for the review! Fast as a 🥷

Issue #648, opened 2022-03-07
Question: best practice for timing out poll_ready?


myagley:
The timeout service allows a request to be timed out. However, it appears that the timeout timer begins only after the inner service is ready, and only applies to the `call`.  (i.e., the timer is setup in the `call` and only affects the returned future.)

Is there a recommended way to timeout starting with the `poll_ready`? An example scenario is a stack using Reconnect, where the endpoint can't be reached/connected. Reconnect will continue to return `poll_ready` `Pending` until connected, and a caller will wait indefinitely for the service to become ready.

Is there support in tower to timeout the `poll_ready`, or should this be handled by the caller? Or am I missing something obvious?

olix0r:
I think this is generally handled by the caller with something like:

```rust
let svc = tokio::time::timeout(timeout, svc.ready()).await?;
```
or
```rust
let rsp = tokio::time::timeout(timeout, svc.oneshot(req)).await?;
```

While it's definitely pretty easy to write a tower middleware that times out when a service that becomes ready, I'm not sure how useful it really is for a readiness timeout to be encoded into the stack -- all it can really do is cause the whole stack to become defunct (by returning an error in poll_ready), so it's generally up to the caller to figure out how long it wants to wait. I could probably be convinced otherwise, though :)

myagley:
Thanks for the response. This makes sense. I figured that timing out the `poll_ready` should leave the service unusable (which is not usually what you want). Thanks for confirming.

PR #715, opened 2023-01-01
Add a timeout for service discovery

ThomWright:
## Context

As part of looking into timeouts in various parts of the stack in TrueLayer/ginepro#37 I found that we have no timeout for DNS resolution. As currently implemented, it's possible for requests to hang forever if DNS never resolves and no services are discovered.

Applications can add their own timeouts around requests, but I think it can be helpful for underlying libraries to offer timeouts for things like TCP connection and DNS resolution (or general service discovery), especially as fallbacks for when higher level timeouts are forgotten!

## Changes

I've taken a shot at solving this, but I'm not very familiar with writing lower level async code using `poll` etc. rather than async/await. I can't say I'm confident that this is the best approach or that I haven't missed something.

With that in mind:

1. Is there a better way to solve this problem?
2. Is the change in the type of `Balance::Future` a problematic breaking change?
3. Are there any more docs or test cases I should consider adding?

I should probably have raised an issue first before diving into a solution, but it's Christmas and I got carried away 😅

### DIFF
--- a/tower/src/balance/p2c/future.rs
+++ b/tower/src/balance/p2c/future.rs
@@ -1,0 +1,64 @@
+//! Future types
+
+use std::future::Future;
+use std::pin::Pin;
+use std::task::{Context, Poll};
+
+use futures_core::ready;
+use pin_project_lite::pin_project;
+
+use crate::timeout::error::Elapsed;
+
+pin_project! {
+    /// Future for the [`Balance`] service.
+    ///
+    /// [`Balance`]: crate::balance::p2c::Balance
+    #[derive(Debug)]
+    pub struct ResponseFuture<F> {
+        #[pin]
+        state: ResponseState<F>,
+    }
+}
+
+pin_project! {
+    #[project = ResponseStateProj]
+    #[derive(Debug)]
+    enum ResponseState<F> {
+        Called {
+            #[pin]
+            fut: F
+        },
+        TimedOut,
+    }
+}
+
+impl<F> ResponseFuture<F> {
+    pub(crate) fn called(fut: F) -> Self {
+        ResponseFuture {
+            state: ResponseState::Called { fut },
+        }
+    }
+
+    pub(crate) fn timed_out() -> Self {
+        ResponseFuture {
+            state: ResponseState::TimedOut,
+        }
+    }
+}
+
+impl<F, T, E> Future for ResponseFuture<F>
+where
+    F: Future<Output = Result<T, E>>,
+    E: Into<crate::BoxError>,
+{
+    type Output = Result<T, crate::BoxError>;
+
+    fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+        match self.project().state.project() {
+            ResponseStateProj::Called { fut } => {
+                Poll::Ready(ready!(fut.poll(cx)).map_err(Into::into))
+            }
+            ResponseStateProj::TimedOut => Poll::Ready(Err(Elapsed::new().into())),
+        }
+    }
+}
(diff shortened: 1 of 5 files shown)


seanmonstar:
I've not had the time to review this, but I'm case it helps, I'd expect that the balance layer wouldn't need any special knowledge of timeouts, and instead you could fit it at a layer above or below.

ThomWright:
> I've not had the time to review this, but I'm case it helps, I'd expect that the balance layer wouldn't need any special knowledge of timeouts, and instead you could fit it at a layer above or below.

Thanks, I was kind of wondering if the timeout logic could be extracted somehow (it would be nice), but wasn't sure how to do it.

At a layer above I could imagine a generic timeout on `poll_ready()`, but 1. it was discussed in #648 and there didn't seem much appetite and 2. I don't think it would discriminate between service discovery having completed (e.g. DNS taking a long time) and some other cause of non-readiness (e.g. being rate limited at a lower layer). But maybe there's no good reason to discriminate and I got a bit too fixated on DNS timeouts!

I'm not sure how it would be applied at a layer below.

On another note: I just noticed that `UnreadyService` in `tower/src/balance/p2c/service.rs` looks like dead code. From I can see that concept now lives in `ReadyCache` as `Pending`. I can remove that in a separate PR, unless I'm missing something.

olix0r:
In linkerd, we use composable middlewares like [failfast](https://github.com/linkerd/linkerd2-proxy/blob/e8fca296797da76e7ccbb3ce60828f2ac883167c/linkerd/stack/src/failfast.rs) to enforce readiness timeouts.

I agree that it's not really appropriate to add this functionality to the balancer directly. It may be worth upstreaming our failfast/loadshed modules, though.

ThomWright:
> I think dns timeouts should probably exist within the dns layer itself. So if its hyper we should add timeouts there which should then trickle back up. This would also help users that don't use tower.

Yeah, I agree. Ideally this timeout would be elsewhere. I wonder if a better description of the problem I'm working around would be helpful in explaining why I ended up here.

[Ginepro](https://github.com/TrueLayer/ginepro) has two [resolution strategies](https://docs.rs/ginepro/latest/ginepro/enum.ResolutionStrategy.html): Eager and Lazy. The Eager variant will resolve a set of initial IPs with a timeout, which works nicely for the initial DNS resolution. However, both strategies run a background task to do periodic DNS probes and send updates to the balancer. When the balancer ends up in a state with no services, there's currently no way to time out on DNS itself since it's outside of the request flow. 

I first tried solving this problem within ginepro itself and wasn't happy with my solution. Since the consensus seems to be that the balancer isn't an appropriate place for this timeout (fair), I reckon I'll head back to ginepro and give it a second attempt. I learned a few things while writing this which I think will help.

> In linkerd, we use composable middlewares like [failfast](https://github.com/linkerd/linkerd2-proxy/blob/e8fca296797da76e7ccbb3ce60828f2ac883167c/linkerd/stack/src/failfast.rs) to enforce readiness timeouts.
> 
> I agree that it's not really appropriate to add this functionality to the balancer directly. It may be worth upstreaming our failfast/loadshed modules, though.

The FailFast middleware looks like a good candidate for upstreaming to me.

Linkerd's [LoadShed](https://github.com/linkerd/linkerd2-proxy/blob/e8fca296797da76e7ccbb3ce60828f2ac883167c/linkerd/stack/src/loadshed.rs) looks pretty similar to tower's own [LoadShed](https://github.com/tower-rs/tower/blob/master/tower/src/load_shed/mod.rs).

Anyway, thank you all for the feedback. I'm happy to close this PR given this doesn't seem to be the right approach.

PR #716, opened 2023-01-10
use a simpler random 2-sampler

conradludgate:
While this only saves 40ns (70ns down to 30ns - probably more if there are other threads causing allocator contention)
per sample, I still think it makes sense to not allocate here.

This was the only use of `sample_inplace` in this codebase. I also have an impl that scales to higher `amount`s, but it uses const generics which won't fit in the current MSRV

### DIFF
--- a/tower/src/util/rng.rs
+++ b/tower/src/util/rng.rs
@@ -159,13 +159,22 @@ mod tests {
             let mut rng = HasherRng::default();
             rng.counter = counter;
 
-            let [a, b] = super::sample_floyd2(&mut rng, length, amount);
+            let [a, b] = super::sample_floyd2(&mut rng, length);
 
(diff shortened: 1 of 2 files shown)


tower/src/util/rng.rs
```diff
@@ -167,32 +151,21 @@ mod tests {
...
+            if a >= length || b >= length || a == b {
+                return TestResult::failed();
             }
 
             TestResult::passed()
         }
     }
-
-    #[test]
-    fn sample_inplace_boundaries() {
```
LucioFranco, line 191: does this test not apply to this anymore? or is this case covered in the quickcheck test?

LucioFranco (approved):
Thanks!

PR #719, opened 2023-02-22
Add "make_infallible_service"

DaAitch:
`axum::routing::Router` and other server impls using `tower`, often want `Infallible` services.

This PR brings `.make_infallible(err_to_ok_fn)` to any `Service` to fit `Service<.., Error = Infallible>`.
I often need this boilerplate when gluing together a server so I thought opening a PR.

Please let me know, if this is a good idea and if so, I will work on the documentation.

### DIFF
--- a/tower/src/make/make_infallible_service.rs
+++ b/tower/src/make/make_infallible_service.rs
@@ -1,0 +1,174 @@
+use crate::sealed::Sealed;
+use pin_project_lite::pin_project;
+use std::future::Future;
+use std::{
+    convert::Infallible,
+    pin::Pin,
+    task::{Context, Poll},
+};
+use tower_service::Service;
+
+// TODO: documentation :)
+
+pub trait MakeInfallibleService<S, F, Request>: Sealed<(Request,)> {
+    fn make_infallible(self, err_into_ok: F) -> InfallibleService<S, F>;
+}
+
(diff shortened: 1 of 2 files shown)


tower/src/make/make_infallible_service.rs
```diff
@@ -0,0 +1,174 @@
...
+{
+    type Response = S::Response;
+    type Error = Infallible;
+    type Future = InfallibleCall<F, S::Future>;
+
+    fn poll_ready(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
+        match self.service.poll_ready(cx) {
+            Poll::Pending => Poll::Pending,
+            Poll::Ready(Ok(())) => Poll::Ready(Ok(())),
+            Poll::Ready(Err(_)) => panic!("poll_ready must not return an error!"),
```
davidpdrsn, line 63: Hm panicking here doesn't feel right 😕

axum already has https://docs.rs/axum/0.6.7/axum/error_handling/struct.HandleErrorLayer.html. I don't think it's appropriate to put in tower though. It doesn't discard services when poll_ready returns an error which ideally you should.

DaAitch: I thought there is nothing more reasonable to do here than panicking, but yes I also struggled here.

So great, didn't notice `HandleErrorLayer`. Will check it out.
I'm going to close the PR.

PR #728, opened 2023-04-11
Add `#[must_use]` warning annotation to `Service::call()`

mishazharov:
While doing some unrelated work I noticed that the `Service::call()` method didn't provide a warning when its future wasn't used or awaited on. If this isn't intentional, I've added the annotation to the trait.

While patching my repo for the fix, I noticed some `Cargo.toml` version numbers were out of date as well, so I bumped them

### DIFF
--- a/tower-service/src/lib.rs
+++ b/tower-service/src/lib.rs
@@ -351,6 +351,7 @@ pub trait Service<Request> {
     ///
     /// Implementations are permitted to panic if `call` is invoked without
     /// obtaining `Poll::Ready(Ok(()))` from `poll_ready`.
+    #[must_use = "futures do nothing unless you `.await` or poll them"]
     fn call(&mut self, req: Request) -> Self::Future;
 }
 


tower-layer/Cargo.toml
```diff
@@ -6,7 +6,7 @@ name = "tower-layer"
 #   - README.md
 # - Update CHANGELOG.md.
 # - Create "v0.1.x" git tag.
-version = "0.3.1"
+version = "0.3.2"
```
davidpdrsn, line 9: Please split these changes into a separate PR, also how?! 😅 did someone publish a new release and forget to push to GitHub?

seanmonstar:
All futures have this warning automatically, by being on the traits in libstd, no?

davidpdrsn:
> All futures have this warning automatically, by being on the traits in libstd, no?

I thought so too but doesn't seem like it:

```rust
fn foo(svc: impl Service) {
    svc.call();
}

trait Service {
    type Future: Future;

    fn call(self) -> Self::Future;
}
```

That generates no warnings for me. Adding `#[must_use]` to call does make a warning appear.

tower-layer/Cargo.toml
mishazharov: > Please split these changes into a separate PR, also how?!

Done! https://github.com/tower-rs/tower/pull/729

> did someone publish a new release and forget to push to GitHub?

and no idea haha

LucioFranco:
Is this considered a breaking change? Since it adds a warning it might require a code change?

davidpdrsn:
> Is this considered a breaking change? Since it adds a warning it might require a code change?

I wouldn't say so. If people have `#[deny(warnings)]` then they're opting in to breakage like this, otherwise it should be fine. In axum I also recently added `#[must_use]` to a bunch of things, in a patch version, and no one has complained.

mishazharov:
Is there any remaining work to do here?

davidpdrsn (approved):

davidpdrsn:
> Is there any remaining work to do here?

Nope! Sorry for the delay.

PR #732, opened 2023-05-02
feat: enable non-`'static` where possible

Daniel-Bloom-dfinity:
## Motivation

Currently, many tower service wrappers and layers are limited to working with `'static` types, which limits its usefulness when working with non-`static`ally borrowed things to and from scoped tasks (e.g. `async-scoped`).

## Solution

Remove the `'static` restrictions where possible, mostly by introducing lifetime parameters.

This change excludes `Buffer` and `BufferLayer` which are more difficult to make non-`'static`. It is possible to do by abstracting `tokio::spawn` into a trait (which `async_scoped::Scope` can also implement), this is also pending the publish of the recently merged tokio-rs/tokio#5665.


GlenDC (commented):
Looks like a welcome change, but it is breaking, no? So probably has to wait for a next minor version updatae?

tobz:
Closing in favor in #765.

PR #741, opened 2023-06-27
Upgrade to indexmap v2 (MSRV 1.63)

cuviper:


tower/src/ready_cache/cache.rs
```diff
@@ -193,7 +193,7 @@ where
     }
 
     /// Obtains a mutable reference to a service in the ready set by index.
-    pub fn get_ready_index_mut(&mut self, idx: usize) -> Option<(&mut K, &mut S)> {
+    pub fn get_ready_index_mut(&mut self, idx: usize) -> Option<(&K, &mut S)> {
```
cuviper, line 196: This is a breaking change, so I suppose it should go on your 0.5 milestone. But if you *really* want the mutable key, there's an opt-in by using the `MutableKeys` trait and its [`get_index_mut2`](https://docs.rs/indexmap/2.0.0/indexmap/map/trait.MutableKeys.html#tymethod.get_index_mut2) method.

cuviper:
We were later able to relax `hashbrown` and `indexmap` to MSRV 1.63, so I've updated that here. Either way, this is well beyond `tower`'s stated 6-month MSRV policy.

Ameobea:
`ahash`, a dependency of v1 indexmap, no longer compiles on new Rust nightly versions: https://github.com/tkaitchuck/aHash/issues/200

As a result, the latest version of tower no longer builds on Rust nightly

This PR should fix that problem

cuviper:
> `ahash`, a dependency of v1 indexmap

I'm not sure what you're seeing, but `indexmap` has never had a dependency on `ahash`. Even its `hashbrown` dependency has always been with `default-features = false` to avoid `ahash` there, unless something else enabled that.

Ameobea:
> > `ahash`, a dependency of v1 indexmap
> 
> I'm not sure what you're seeing, but `indexmap` has never had a dependency on `ahash`. Even its `hashbrown` dependency has always been with `default-features = false` to avoid `ahash` there, unless something else enabled that.

You're right, I wasn't precise.  `ahash` is a dependency on more level down as you pointed out: `tower` -> `indexmap` -> `hashbrown` -> `ahash`.

Pulling in tower by itself with no other crates, even with `features = ["full"]` set, still compiles on nightly because `hashbrown` has `default-features = false` set by default.

It seems some other crate in our workspace is turning on the `ahash` feature of `hashbrown`.  The output of `cargo tree` was showing `ahash` in the tree underneath `tower` even though the feature got turned on elsewhere.

So in that case, bumping indexmap to v2 on tower probably wouldn't help out since there is some other crate somewhere depending on it and turning on the feature.

Apologies for the confusion here.

cuviper:
OK -- `cargo tree -i ahash -e features` may help you track that down!

tower/src/ready_cache/cache.rs
tobz: It's feels like more of a footgun to have the key be mutable, given the warnings mentioned in `MutableKeys::get_index_mut2`... and there's really nothing in the original PR or docs that speaks specifically to the _key_ needing to be mutable... so I think this is a worthwhile change to make given that changing the MSRV is breaking anyways.

tobz (approved):

tobz:
This has been released as part of `tower@v0.5.0`.

Thanks again for your contribution!

Issue #751, opened 2023-11-05
Extending `Building a middleware from scratch` guide


samueljoli:
The [Building a middleware from scratch](https://github.com/tower-rs/tower/blob/master/guides/building-a-middleware-from-scratch.md#building-a-middleware-from-scratch) is a great guide, I think it would be an improvement if the following steps were added:

- Creating a layer
```rs
impl TimeoutLayer {
    /// Create a timeout from a duration
    pub fn new(timeout: Duration) -> Self {
        TimeoutLayer { timeout }
    }
}

impl<S> Layer<S> for TimeoutLayer {
    type Service = Timeout<S>;

    fn layer(&self, service: S) -> Self::Service {
        Timeout::new(service, self.timeout)
    }
}

```
- [Error handling for middleware](https://docs.rs/axum/latest/axum/middleware/index.html#error-handling-for-middleware)

- Using the layer (integration with `ServiceBuilder`

PR #752, opened 2023-11-05
docs: extend building a middleware from scratch guide

samueljoli:
add steps:
- create a layer
- integration with service builder

addresses: #751

### DIFF
--- a/guides/building-a-middleware-from-scratch.md
+++ b/guides/building-a-middleware-from-scratch.md
@@ -487,7 +487,7 @@ where
 }
 ```
 
-Finally we have to revisit our `Service` implementation and update it to also
+Revisit our `Service` implementation and update it to also
 use `BoxError`:
 
 ```rust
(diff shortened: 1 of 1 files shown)


guides/building-a-middleware-from-scratch.md
```diff
@@ -519,6 +519,94 @@ where
...
+    |
+123 |     Router::new().layer(ServiceBuilder::new().layer(TimeoutLayer::new(Duration::from_secs(5))))
+    |                   ----- ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `From<Box<(dyn std::error::Error + Send + Sync + 'static)>>` is not implemented for `Infallible`
+    |                   |
+    |                   required by a bound introduced by this call
+    |
+```
+
+That's because there's an attempt to convert a Box<dyn StdError + Send + Sync + 'static> into Infallible. The Infallible type is used to signify that a function will never return an error.
+> Axum’s error handling model requires handlers to always return a response. However middleware is one possible way to introduce errors into an application. If hyper receives an error the connection will be closed without sending a response. Thus axum requires those errors to be handled gracefully. - [doc](https://docs.rs/axum/latest/axum/middleware/index.html#error-handling-for-middleware)
```
GlenDC, line 570: Should we really add specific framework mentions to these fundamental docs? Is that done elsewhere? Just asking, as I'm not sure of the value of it, given not everyone might be familiar with hyper and/or axum.

samueljoli: @GlenDC That's a valid point. Apologies, I'm kinda new to the tokio ecosystem and didn't fully realize that tower is incredibly framework agnostic. The motivation was to provide an example of usage and what issues might come up when integrating with Axum. But you're right in that:
> given not everyone might be familiar with hyper and/or axum.

Maybe this can find a home in one of Axum's examples

I can close out this PR and the accompanying Issue

PR #754, opened 2023-12-04
ensure that re-exported symbols show feature label in root

### LINKED ISSUE #743: `tower::service_fn` docs don't say that you need the `util` flag
See the docs for it: https://docs.rs/tower/latest/tower/fn.service_fn.html.
It doesn't mention that you need to enable `util` which is a bit confusing.

GlenDC:
closes #743

### DIFF
--- a/tower/src/lib.rs
+++ b/tower/src/lib.rs
@@ -202,15 +202,20 @@ pub mod layer;
 
 #[cfg(feature = "util")]
 #[doc(inline)]
+#[cfg_attr(docsrs, doc(cfg(feature = "util")))]
 pub use self::util::{service_fn, ServiceExt};
 
 #[doc(inline)]
 pub use crate::builder::ServiceBuilder;
+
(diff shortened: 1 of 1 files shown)


jplatte (commented):
Have you tested this? We are already using `doc_auto_cfg` so I don't  think these `doc(cfg)` attributes should make a difference. If my memory serves me right, it might instead help to put the `doc(cfg)` on the definition of those items, rather than the re-exports.

GlenDC:
@jplatte in my test it didn't show up before and it did show after (locally using the right --cfg flags etc).

Whether it's the right solution I do not know. Feel free to propose or guide me into doing something different if you think I approached it wrong.

jplatte:
Okay, if you did test it then I have no concerns.

tobz (approved):

tobz:
This has been released as part of `tower@v0.5.0`.

Thanks again for your contribution!

PR #760, opened 2024-01-07
feat: Make new functions const when possible

kerkmann:
The main reason was to allow to initialize the RateLimitLayer in a const context. So why not making ever new function const (wherever it's possible). :P

### DIFF
--- a/tower-layer/src/identity.rs
+++ b/tower-layer/src/identity.rs
@@ -14,7 +14,7 @@ pub struct Identity {
 
 impl Identity {
     /// Create a new [`Identity`] value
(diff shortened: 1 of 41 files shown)


tobz (approved):

tobz:
This is blocked on the MSRV, which would need to be at least 1.61 to support generic trait bounds on const functions.

Parking for now.

tobz:
This has been released as part of `tower@v0.5.0`, `tower-layer@v0.3.3`, and `tower-service@v0.3.3`.

Thanksd again for your contribution!

PR #758, opened 2023-12-27
Update indexmap dependency & fix compiler warnings

serprex:
#741 proposes changing `get_ready_index_mut`'s signature, but that would be a semver breaking change, so can wait for `0.5`

Also remove unused lazy_static dev dependency

PR #761, opened 2024-01-15
Upgrade indexmap to version 2.

pixlwave:
This PR upgrades the dependency on the indexmap crate to version 2.0.

Checking the [release notes](https://github.com/bluss/indexmap/blob/master/RELEASES.md) I don't see any breaking changes that would affect this crate, but I'm still relatively new to Rust, so please don't take my word for it.


pixlwave:
> I don't see any breaking changes that would affect this crate

Ah, and now I've opened the PR I realise I didn't compile the crate with the `ready-cache` feature enabled - it turns out there is indeed breaking changes to handle. 🤦‍♂️

serprex:
Duplicate of #758

pixlwave:
My bad! Closing in favour of #758 👍

PR #762, opened 2024-01-24
docs: provide an example to explain the cloning inner services

shengsheng:
Provide the example "ConcurrencyLimit" to explain why "Even though the original service is ready, the clone might not be".
https://docs.rs/tower/latest/tower/limit/struct.ConcurrencyLimit.html#impl-Clone

### DIFF
--- a/tower-service/src/lib.rs
+++ b/tower-service/src/lib.rs
@@ -236,7 +236,10 @@ use std::task::{Context, Poll};
 ///
 /// Services are permitted to panic if `call` is invoked without obtaining `Poll::Ready(Ok(()))`
 /// from `poll_ready`. You should therefore be careful when cloning services for example to move
-/// them into boxed futures. Even though the original service is ready, the clone might not be.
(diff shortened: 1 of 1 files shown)


tower-service/src/lib.rs
```diff
@@ -236,7 +236,10 @@ use std::task::{Context, Poll};
 ///
 /// Services are permitted to panic if `call` is invoked without obtaining `Poll::Ready(Ok(()))`
 /// from `poll_ready`. You should therefore be careful when cloning services for example to move
-/// them into boxed futures. Even though the original service is ready, the clone might not be.
+/// them into boxed futures. Even though the original service is ready, the clone might not be,
```
david-perez, line 239: I don't understand how `ConcurrencyLimit` exemplifies what this docs section is trying to convey. If you're referring to the fact that cloning an inner `ConcurrencyLimit` service resets the permit to unacquired and that `poll_ready` acquires it, that will only be understood by readers who have read `ConcurrencyLimit`'s source code.

PR #765, opened 2024-03-18
feat: enable non-'static where possible

Daniel-Aaron-Bloom:
## Motivation

Currently, many tower service wrappers and layers are limited to working with `'static` types, which limits its usefulness when working with non-`static`ally borrowed things to and from scoped tasks (e.g. `async-scoped`).

## Solution

Remove the `'static` restrictions where possible, mostly by introducing lifetime parameters.

This change excludes `Buffer` and `BufferLayer` which are more difficult to make non-`'static`.

This PR replaces #732


ripytide (approved):

ripytide:
It's good progress at least though if you maintainers are happy with the change given the requirement that it is merged as part of the next major breaking release and not merged until then.

Daniel-Aaron-Bloom:
Are there any labels that could be applied to this PR to ensure it doesn't get forgotten in vNext?

PR #777, opened 2024-07-22
Add `BoxCloneSyncService`

### LINKED ISSUE #770: Add `BoxCloneSyncService`
Currently `BoxCloneService` does not implement `Sync` even though the inner Service implements `Sync`. This can be solved by adding a `Sync` bound to the inner service bounds. 

Other projects such as axum also have [this](https://github.com/tokio-rs/axum/blob/5b316ebc64fed4df4d9153b57d7fe2e58cbadb4b/axum/src/box_clone_service.rs) just internally. Might be worth adding it upstream. 

I would be willing to submit a PR for this.

tthebst:
resolves #770

### DIFF
--- a/tower/src/util/boxed_clone_sync.rs
+++ b/tower/src/util/boxed_clone_sync.rs
@@ -18,7 +18,8 @@ use tower_service::Service;
 pub struct BoxCloneSyncService<T, U, E>(
     Box<
         dyn CloneService<T, Response = U, Error = E, Future = BoxFuture<'static, Result<U, E>>>
-            + Send,
+            + Send
+            + Sync,
     >,
 );
 
@@ -26,7 +27,7 @@ impl<T, U, E> BoxCloneSyncService<T, U, E> {
     /// Create a new `BoxCloneService`.
     pub fn new<S>(inner: S) -> Self
     where
-        S: Service<T, Response = U, Error = E> + Clone + Send + 'static,
+        S: Service<T, Response = U, Error = E> + Clone + Send + Sync + 'static,
         S::Future: Send + 'static,
     {
         let inner = inner.map_future(|f| Box::pin(f) as _);
@@ -39,7 +40,7 @@ impl<T, U, E> BoxCloneSyncService<T, U, E> {
     /// [`Layer`]: crate::Layer
     pub fn layer<S>() -> LayerFn<fn(S) -> Self>
     where
-        S: Service<T, Response = U, Error = E> + Clone + Send + 'static,
+        S: Service<T, Response = U, Error = E> + Clone + Send + Sync + 'static,
         S::Future: Send + 'static,
     {
         layer_fn(Self::new)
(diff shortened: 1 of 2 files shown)


jlizen:
Thanks for cutting this! This is useful in an issue i'm working on for reqwest, to expose the tower stack:
https://github.com/seanmonstar/reqwest/issues/2490#issuecomment-2525347550

Will ping in the tower discord to try to help find a reviewer.

jlizen:
@tthebst  do you see a reason we couldn't add `BoxCloneSyncServiceLayer` similar to [BoxCloneServiceLayer](https://github.com/tower-rs/tower/blob/6283f3aff1895c234341cc0766cbd7db233ba1ba/tower/src/util/boxed/layer_clone.r)?

tower/src/util/boxed_clone_sync.rs
```diff
@@ -0,0 +1,101 @@
...
+    {
+        let inner = inner.map_future(|f| Box::pin(f) as _);
+        BoxCloneSyncService(Box::new(inner))
+    }
+
+    /// Returns a [`Layer`] for wrapping a [`Service`] in a [`BoxCloneSyncService`]
+    /// middleware.
+    ///
+    /// [`Layer`]: crate::Layer
+    pub fn layer<S>() -> LayerFn<fn(S) -> Self>
```
seanmonstar, line 41: Part of me is a little skeptical of naming this util type here. It would mean we couldn't change to a dedicated layer type without it being "breaking". Changing this to `impl Layer` could work, but people eventually complain.

Is this what you were referring to @jlizen?

jlizen: I'm not sure I understand the concern. This maps to how `BoxCloneService` and `BoxService` implement it as well. The service has function that returns an unnamed layer function, or else there is a separate layer module. 

Both started out without the dedicated layer and added it in, so perhaps the unnamed function is more vestigial.

Anyway, I think it's probably fine to have both, and libraries can use the named one if they want to expose it to the caller. Probably more confusing to have the API diverge from the other type erasure wrappers.

Fwiw, for my purposes in that reqwest commit, I only need the service wrapper anyway. I've switched over to composing with generics, and then if custom layers are provided, wrapping the final thing in a `BoxCloneSyncService` to avoid polluting the built client with long generic bounds everywhere.

seanmonstar: Ah, I didn't realize other types have two separate layer constructors. Hm, that seems like an eventual place to improve, but oh well.

seanmonstar (approved):
Excellent work, thank you!

yyaroshevich:
Should this one also be updated to be considered for merging? https://github.com/tower-rs/tower/pull/746

yyaroshevich:
Does it make sense to extend `ServiceExt` to include helper method to construct new service wrapper? The same seem to be also relevant for an existing `UnsyncBoxService` wrapper.

jlizen:
Good call, was thinking the same. I can cut that if you are busy, let me
know.

jlizen:
> Does it make sense to extend `ServiceExt` to include helper method to construct new service wrapper? The same seem to be also relevant for an existing `UnsyncBoxService` wrapper.

Sure, I can cut that tomorrow unless you have time to do it now. Was meaning to but slipped my mind. 

Regarding #746 , you could probably ping for a reviewer in the tokio discord's tower-dev channel.

yyaroshevich:
> Sure, I can cut that tomorrow unless you have time to do it now. Was meaning to but slipped my mind.

No rush, just was curious maybe there was a reason no to include.

> Regarding https://github.com/tower-rs/tower/pull/746 , you could probably ping for a reviewer in the tokio discord's tower-dev channel.

I'm not the author, just today wanted this to be presented in tower. I'll need to take closer look at existing MR, it look abandoned from the first glance, so maybe it's already need to be adjusted/re-done.

PR #779, opened 2024-07-22
fix warnings found when running check/doc commands

GlenDC:


GlenDC:
Due to our MSRV check it does require a lower bound on the minor version of `futures` dep, FYI @tobz .

examples/Cargo.toml
```diff
@@ -8,11 +8,11 @@ edition = "2018"
 # [dependencies] instead.
 [dev-dependencies]
 tower = { version = "0.4", path = "../tower", features = ["full"] }
-tower-service = "0.3" 
+tower-service = "0.3"
 tokio = { version = "1.0", features = ["full"] }
 rand = "0.8"
 pin-project = "1.0"
-futures = "0.3"
+futures = "^0.3.22"
```
tobz, line 15: You can drop the leading caret. That's already the implicit behavior of Cargo's version resolver.

tobz (changes requested):

tobz (approved):

PR #788, opened 2024-08-18
Upgrade http and sync_wrapper dependencies

elichai:
Upgrade these dependencies to allow downstream users to reduce duplicate dependencies
(specifically this makes `axum` CI fail for duplicate dependencies when upgrading to tower 0.5: https://github.com/tokio-rs/axum/actions/runs/10440315723/job/28910071844?pr=2880 )


tottoto:
I think this is not breaking change as `http` crate is dev-dependency, and `sync_wrapper`'s type is not exposed and the behaviour is not changed from 0.1.

elichai:
@jplatte Any chance we can get this in 5.1?

jplatte:
Sorry, I maintain `tower-http` but not `tower`. I don't think I even have the right to merge things here.

jplatte (approved):
I can check though...

elichai:
@seanmonstar ping maybe?

kanpov:
@seanmonstar any update on this? Would be good to eliminate sync_wrapper duplication

Issue #682, opened 2022-08-11
Retry middleware improvements


LucioFranco:
This issue scopes out the changes we are proposing to the retry middleware to improve its ergonomics. Currently, the retry middleware is quite hard to use and requires implementing a custom `Policy`. Writing this policy is non-trivial and is more work than it should be.

These improvements are aimed at setting up retries with tower to be easier and more user friendly. As well as providing good defaults that work out of the box. 

List of improvements to `tower` and` tower-http`:

- [ ] Simplify `Policy` (https://github.com/tower-rs/tower/pull/681).
  - Change trait fn to take `&mut self`.
  - Change `Future` output to `()`.
  - Alllow `Policy` to be object safe.
- [ ] Add generic backoff utilities.
  - [ ] Add some `trait Backoff` that has an associated future type that allows others to use this utility without being tied to `tokio::time`.
  - [ ] Add a `ExponentialBackoff` type that implements `Backoff`, ported from `linkerd2-proxy`.
- [ ] Budget improvements.
  - [x] Merge budget docs PR (https://github.com/tower-rs/tower/pull/613).
  - [ ] Implement simpler non-time based token bucket budget.
  - [ ] Add some `Budget` trait to allow users to choose which implementation to use.
- [ ] Add a new batteries included standard retry policy
  - [ ] New `StandardRetryPolicy` combining `impl Backoff` and `impl Budget`.
  - [ ] Add `StandardRetryPolicyBuilder` that accept closures (?) for `is_retryable(&mut Req, &mut Result<Res, E>) -> bool` and a `clone_request(&Req) -> Option<Req>`.
- [ ] `tower-http` improvements.
  - [ ] Add new `retry` module
  - [ ] Implement `ReplayBody` similar to the one implemented in `linkerd2-proxy`.
  - [ ] Add new `HttpRetry` layer that accepts higher level constructs for retrying, like `ClassifyResponse`, and will wrap http request bodies with `ReplayBody`.
- [ ] Documentation
  - [ ] Blog post on how to setup retries with `tower` and `tower-http`.
  - [ ] Examples for thick clients with retries in both `tower` and `tower-http`.

## Example code

`tower` examples with no `http`:

```rust
let policy = StandardRetryPolicy::builder()
    .should_retry(|res| match res {
        Ok(_) => false,
        Err(_) => true,
    })
    .clone_request(|r| Some(*r))
    .build();

let mut svc = ServiceBuilder::new()
    .retry(policy)
    .buffer(10)
[...]

PR #790, opened 2024-08-21
Retry middleware improvements for non-cloneable requests

fraillt:
Hello,

I wasn't sure how to contribute to this project, so I hope this is a right place to start discussing.

Basically this revision solves real problem that I had with [tonic](https://github.com/hyperium/tonic), using [retry](https://github.com/tower-rs/tower/tree/b2c48b46a3110af844b667ace7e14caae70a8a33/tower/src/retry) middleware.
The core of the problem is that [tonic](https://github.com/hyperium/tonic) (a library that implements gRPC) uses HTTP2, which is by design is streaming protocol. For this reason it is not possible to "clone" tonic request by having just a reference to it.
In order to make request clone-able we need to fully consume it, and then reconstruct original object from consumed bytes.

This revision does exactly that.
1. consumes original request and stores it's content into new user defined `CloneableRequest`.
2. later, it recreates back original request from the contents of `CloneableRequest`.

There's some more discussion related to this:
* https://github.com/hyperium/tonic/issues/733
* #682

I think that it doesn't make sense to solve this issue at [tonic](https://github.com/hyperium/tonic) side, because it's doesn't make sense for tonic to have cloneable request (image situation where client is streaming multi-gibabyte file,  how would you consume it? where would you store it? it's not responsibility of tonic).
On the other hand, if you want to have retry mechanism, you must be aware that a request will be copied in order to be able to retry it.

Any feedback is welcome :)

### DIFF
--- a/tower/src/retry/future.rs
+++ b/tower/src/retry/future.rs
@@ -1,5 +1,6 @@
 //! Future types
 
+use super::policy::Outcome;
 use super::{Policy, Retry};
 use futures_core::ready;
 use pin_project_lite::pin_project;
@@ -16,7 +17,7 @@ pin_project! {
         P: Policy<Request, S::Response, S::Error>,
         S: Service<Request>,
     {
-        request: Option<Request>,
+        request: P::CloneableRequest,
         #[pin]
         retry: Retry<P, S>,
         #[pin]
@@ -49,7 +50,7 @@ where
     S: Service<Request>,
 {
     pub(crate) fn new(
-        request: Option<Request>,
+        request: P::CloneableRequest,
         retry: Retry<P, S>,
         future: S::Future,
     ) -> ResponseFuture<P, S, Request> {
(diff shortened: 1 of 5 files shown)


Kayanski:
Hello ! This was very interesting to me and I implemented something very similar inspired off of your work in my application. 

However, instead of adding another method in the trait, I simply changed the `clone_request` definition to the following: 
```rust
fn clone_request(&mut self, req: Req) -> (Req, Option<Req>);
``` 

This allows for less methods in the trait that serve the same purpose and is more straightforward. The doc comment could be: 
```rust
  /// Tries to clone a request before being passed to the inner service.
  ///
  /// If the request can be cloned, it should return the original request and its clone wrapped in [`Some`]
  ///
  /// If the request cannot be cloned simply, it is allowed to drop and reconstruct the original request twice
  ///
  /// If the request cannot be cloned, returns (original_request, [`None`]). Moreover, the retry
  /// function will not be called if the [`None`] is returned.
```

PR #778, opened 2024-07-22
use workspace dependencies for tower

GlenDC:
Closes #776

PR #792, opened 2024-08-24
fix: relax MSRV to 1.63

### LINKED ISSUE #785: Changelog says that MSRV is 1.63 but it's 1.64
Encountered in https://github.com/hyperium/hyper-util/pull/144

tottoto:
Resolves #785. Relaxes MSRV to 1.63 from 1.64.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -7,7 +7,7 @@ on:
   pull_request: {}
 
 env:
-  MSRV: 1.64.0
+  MSRV: 1.63.0
 
 jobs:
   check-stable:
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -6,23 +6,3 @@ members = [
   "tower-service",
   "tower-test",
 ]
-
-[workspace.dependencies]
-futures = "0.3.22"
-futures-core = "0.3.22"
-futures-util = { version = "0.3.22", default-features = false }
-hdrhistogram = { version = "7.0", default-features = false }
-http = "0.2"
-indexmap = "2.0.2"
-lazy_static = "1.4.0"
-pin-project-lite = "0.2.7"
-quickcheck = "1"
-rand = "0.8"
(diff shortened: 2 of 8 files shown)


paolobarbolini:
Considering that the published crate gets the `Cargo.toml` flattened I wander if the MSRV is already 1.63 without this PR :thinking:. See https://docs.rs/crate/tower/0.5.1/source/Cargo.toml

jplatte:
This effectively reverts https://github.com/tower-rs/tower/pull/778. Is there a reason you chose to do that instead of fixing #785 the "obvious" way, by updating the changelog?

tottoto:
I'm not sure what is the obvious way, but the reason is `hyper-util`'s MSRV is 1.63 and `tower` doesn't use any 1.64 features other than workspace inheritance. As I think maintaining support for older versions of Rust also incurs a certain cost, if it does not conform to tower's policy, I will withdraw the proposal.

tottoto:
I'm not sure what is the obvious way, but the reason is `hyper-util`'s MSRV is 1.63 and `tower` doesn't use any 1.64 features other than workspace inheritance. As I think maintaining support for older versions of Rust incurs a certain cost, if this does not conform to tower's policy, I will withdraw this proposal.

jplatte:
I don't have a strong opinion on this one way or the other, let's see what Sean says in https://github.com/hyperium/hyper-util/pull/144.

paolobarbolini:
Isn't the MSRV effectively 1.63 once the crate is published? I'd expect a `cargo` version not aware of workspace dependencies to still be able to parse this https://docs.rs/crate/tower/0.5.1/source/Cargo.toml

jplatte:
That's a good point...

seanmonstar:
At least for hyper-util, this isn't needed, since I removed the dependency on `tower`.

tottoto:
Closed as the main motivation has been resolved.

PR #796, opened 2024-09-27
fix trait bounds on make `AsService` & `IntoService`

Boog900:
The trait bounds on `AsService` & `IntoService` both require that `M: Service`, however this defeats the purpose of the types as if we already require that `M: Service` then all the `ServiceExt` methods would already be reachable on `M`.

This PR changes the bounds to `M: MakeService`.

### DIFF
--- a/tower/src/make/make_service.rs
+++ b/tower/src/make/make_service.rs
@@ -189,13 +189,12 @@ where
     }
 }
 
-impl<M, S, Target, Request> Service<Target> for IntoService<M, Request>
+impl<M, Target, Request> Service<Target> for IntoService<M, Request>
 where
-    M: Service<Target, Response = S>,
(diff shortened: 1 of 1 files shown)


ripytide (approved):
Yep, this makes sense. I tried coming up with what the trait bounds would look like if you make `M: MakeService` and ended up with exactly the same results.

ripytide:
Actually, after a bit more thought, while I still stand by that this PR fixes a accidental mistake in the blanket trait for implementing `Service` for all `MakeService`s. I don't actually see the point in the `MakeService` trait, in the docs it says its a trait alias, and some of the associated types are different, some are named the same (like `Future`). Overall, I don't see it reducing the complexity of trait bounds for functions that take services returning other services and would probably choose to go without it in my own code.

PR #799, opened 2024-10-26
fix: fix dead code warning for 'Sealed' trait and 'sample_floyd2' func

### LINKED ISSUE #783: Fix dead code warning when default features are turned off.
There's a warning about `Sealed` (in `tower`) not being used when default features are disabled, as seen here: https://github.com/tower-rs/tower/actions/runs/10373855535/job/28719943265#step:5:72

We should fix that.

bp7968h:
### Related Issue

Fixes #783 

### Overview

This PR fixes the dead code warning using the conditional compilation flag.


tower/src/lib.rs
```diff
@@ -219,6 +219,7 @@ pub use tower_layer::Layer;
 pub use tower_service::Service;
 
 #[allow(unreachable_pub)]
+#[cfg(any(feature = "balance", feature = "discover", feature = "full", feature = "make"))]
```
tobz, line 222: No need to check for `full`, since `full` implies the other three features you're checking for here.

tower/src/util/rng.rs
```diff
@@ -116,6 +116,7 @@ where
 ///
 /// ref: This was borrowed and modified from the following Rand implementation
 /// https://github.com/rust-random/rand/blob/b73640705d6714509f8ceccc49e8df996fa19f51/src/seq/index.rs#L375-L411
+#[cfg(any(feature = "full", not(any(feature = "hedge", feature = "retry", feature = "spawn-ready", feature = "util"))))]
```
tobz, line 119: It looks like this method is only used in `Balance`, so it seems like the stanza should just be:

```suggestion
#[cfg(feature = "balance")]
```

tobz (changes requested):

tower/src/lib.rs
bp7968h: Noted, thank you for the advice.

tower/src/util/rng.rs
bp7968h: @tobz, thank you for pointing that out but when I checked the build logs, the warning for `sample_floyd2` appearded on `hedge, retry, spawn-ready and util` and there are 14 features and did not appear on other features doesn't that imply it is being used by other features.

Really appreciate your help.

tower/src/lib.rs
bp7968h: Removed the `full` feature flag.

tower/src/util/rng.rs
bp7968h: @tobz , thank you very much, got it, `tower/lib.rs` controls the `balance` module, and is indeed being only used there, having only `#[cfg(feature = "balance")]` does the job.

seanmonstar (approved):

PR #804, opened 2024-12-19
add ServiceBuilder::boxed_clone_sync helper

jlizen:
As requested in: https://github.com/tower-rs/tower/pull/777

Adding the `ServiceBuilder::boxed_clone_sync()` method to match `boxed_clone` and `boxed()`.

Note that in this case, I used the named, `BoxCloneSyncServiceLayer`, rather than the layer fn returned by `BoxedCloneSyncService::layer()`. I don't really see a good reason to prefer the latter. I think the reason that other helpers use that, is historical (named layer didn't exist yet).

Probably we should migrate everything over, but that seems likely to be a breaking change so I held off. Glad to cut an issue on that if you concur.

The only testing was via a doctest. This matches the other helper.

### DIFF
--- a/tower/src/builder/mod.rs
+++ b/tower/src/builder/mod.rs
@@ -789,6 +789,68 @@ impl<L> ServiceBuilder<L> {
     {
         self.layer(crate::util::BoxCloneService::layer())
     }
+
+    /// This wraps the inner service with the [`Layer`] returned by [`BoxCloneSyncServiceLayer`].
+    ///
+    /// This is similar to the [`boxed_clone`] method, but it requires that `Self` implement
+    /// [`Sync`], and the returned boxed service implements [`Sync`].
+    ///
+    /// See [`BoxCloneSyncService`] for more details.
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// use tower::{Service, ServiceBuilder, BoxError, util::BoxCloneSyncService};
+    /// use std::time::Duration;
+    /// #
+    /// # struct Request;
+    /// # struct Response;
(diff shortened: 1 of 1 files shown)


jlizen:
This is a quickie if you have a minute @seanmonstar , you have context on this from https://github.com/tower-rs/tower/pull/777

tower/src/builder/mod.rs
```diff
@@ -789,6 +789,70 @@ impl<L> ServiceBuilder<L> {
...
+            >,
+            Identity,
+        >,
+    >
+    where
+        L: Layer<S> + Send + Sync + 'static,
+        L::Service: Service<R> + Clone + Send + Sync + 'static,
+        <L::Service as Service<R>>::Response: Send + Sync + 'static,
+        <L::Service as Service<R>>::Error: Send + Sync + 'static,
+        <L::Service as Service<R>>::Future: Send + Sync + 'static,
```
seanmonstar, line 850: I notice this method had several more bounds on associated types than the other method above, are they necessary? (Besides the `Sync` addition).

jlizen: Yeah, I believe I was getting compiler errors without them. It might be related to using the named layer. 

Let me double check though.

jlizen: Ah, indeed the response and error GAT bounds were not needed. Sorry about that. Good catch.

jlizen: CI is passing still, resolving. Thanks for the careful read @seanmonstar

seanmonstar (approved):

PR #811, opened 2025-01-27
Bump rand to v0.9

paolobarbolini:


tower/tests/balance/main.rs
```diff
@@ -28,7 +29,9 @@ impl Service<Req> for Mock {
 impl tower::load::Load for Mock {
     type Metric = usize;
     fn load(&self) -> Self::Metric {
-        rand::random()
```
ripytide, line 31: Why change this from `rand::random()` to the longer, more explicit version? From my reading of the `rand` docs `rand::random` already picks a uniformly distributed random value from the whole range of `usize` values?

ripytide (changes requested):
Could do with some minor changes but no glaring issues. I'm happy to created an improved PR if that's easier than waiting for the original author to make the changes.

tower/tests/balance/main.rs
```diff
@@ -129,7 +134,9 @@ fn stress() {
             } else {
                 // fail a service
                 while !services.is_empty() {
-                    let k = rand::random::<usize>() % (services.iter().last().unwrap().0 + 1);
+                    let k = Uniform::new_inclusive(0, services.iter().last().unwrap().0)
+                        .unwrap()
+                        .sample(&mut rand::rng());
```
ripytide, line 139: Same here

tower/tests/balance/main.rs
paolobarbolini: I've was able to the initial ugly `Uniform::new_inclusive`, but we won't be able to use `rand::random()` because the necessary traits are not implemented on usize anymore.

tower/tests/balance/main.rs
```diff
@@ -114,7 +117,9 @@ fn stress() {
                 } else {
                     // remove
                     while !services.is_empty() {
-                        let k = rand::random::<usize>() % (services.iter().last().unwrap().0 + 1);
+                        let k = Uniform::new_inclusive(0, services.iter().last().unwrap().0)
+                            .unwrap()
+                            .sample(&mut rand::rng());
```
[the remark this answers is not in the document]
paolobarbolini: Changed. I'd missed that `rand::random` and `rand::random_range` take different traits. The first one doesn't accept usize, the latter does.

PR #817, opened 2025-03-11
chore: note `Budget` breaking change in changelog

cratelyn:
`Budget` is now a trait in the 0.5 release. this is a breaking change relative to the 0.4 release, where it was a concrete [struct](https://docs.rs/tower/0.4.13/tower/retry/budget/struct.Budget.html).

this commit updates the changelog to characterize this as a breaking change, rather than an additive change.

### DIFF
--- a/tower/CHANGELOG.md
+++ b/tower/CHANGELOG.md
@@ -37,7 +37,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   increases the flexibility of the retry policy. To update, update your method signature to include `mut` for both parameters. ([#584])
 - **retry**: **Breaking Change** Change Policy to accept &mut self ([#681])
 - **retry**: Add generic backoff utilities ([#685])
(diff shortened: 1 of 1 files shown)


seanmonstar:
The struct still exists...

cratelyn:
> The struct still exists...

i am not sure i follow. do you mean [`TpsBudget`](https://docs.rs/tower/0.5.2/tower/retry/budget/tps_budget/struct.TpsBudget.html)?

even if the struct still exists, in 0.4.13 there is a https://docs.rs/tower/0.4.13/tower/retry/budget/struct.Budget.html struct, and in 0.5 this is now a trait. https://docs.rs/tower/0.5.2/tower/retry/budget/trait.Budget.html.

that's a breaking change, even if the structure still exists under a different name, right?

seanmonstar:
Um, oh right. I don't love that really at all. But whatever, you are indeed right, this was breaking.

seanmonstar (approved):

PR #822, opened 2025-04-30
chore: Remove unused dependency

tottoto:
These dependencies seem not to be used.

### DIFF
--- a/tower/Cargo.toml
+++ b/tower/Cargo.toml
@@ -56,6 +56,7 @@ spawn-ready = ["futures-util", "tokio/sync", "tokio/rt", "util", "tracing"]
 steer = []
 timeout = ["pin-project-lite", "tokio/time"]
 util = ["futures-core", "futures-util", "pin-project-lite", "sync_wrapper"]
+tokio-stream = [] # TODO: Remove this feature at the next breaking release.
 
 [dependencies]
 tower-layer = { version = "0.3.3", path = "../tower-layer" }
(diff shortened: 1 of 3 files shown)


tower/Cargo.toml
```diff
@@ -67,7 +67,6 @@ hdrhistogram = { workspace = true, optional = true }
 indexmap = { workspace = true, optional = true }
 slab = { workspace = true, optional = true }
 tokio = { workspace = true, features = ["sync"], optional = true }
-tokio-stream = { workspace = true, optional = true }
```
jplatte, line 70: This is a breaking change, since optional dependencies that aren't mentioned as `dep:that-dependency` in a feature create an implicit feature. Can you add a feature of the same name with a comment that it's a no-op and will be removed in the future?

jplatte (commented):
Not sure why CI is failing..

jplatte:
CI is broken due to the minimum version for `lazy_static` now being 1.0.0 (for non-dev dependencies).

This should fix it:

```patch
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -90,7 +90,9 @@ jobs:
       - name: "install Rust nightly"
         uses: dtolnay/rust-toolchain@nightly
       - name: Select minimal versions
-        run: cargo update -Z minimal-versions
+        run: |
+          cargo update -Z minimal-versions
+          cargo update -p lazy_static
```

jplatte (approved):
Thanks!

Issue #824, opened 2025-05-05
tower::Balance does not work


c98:
tower-0.5.2

Platform: linux

In the process of using tonic, I found that its dynamic endpoint uses tower. I found a problem in actual use. When `<Balance as Service>::poll_ready` is triggered, `Balance::promote_pending_to_ready`=>`ReadyCache::poll_pending`=>`futures_util::stream::futures_unordered::FuturesUnordered::poll_next` has an issue: assuming there are multiple endpoints, only one `Poll::Ready(Some(Ok((key, svc, cancel_rx))))` will be returned here, and the others are `Poll::Pending`, which makes the balance function invalid.

c98:
It seems that When the Change::Insert(Key, Svc) svc poll_ready phase contains some I/O or polling op, `FuturesUnordered::poll_next` does not work as expected.

c98:
I think I know the reason now.

I'm using Tonic, and when the client initiates a req,
`Buffer::Worker::poll` `poll_next_msg` will be followed by `self.service.poll_ready` calling `balance::poll_ready`. Let's say there are multiple endpoints, one of which is ready, and the other is in pending, and `self.service.poll_ready` still returns `poll:: Ready(Ok(()))`, the problem here is that if no next req arrives, the endpoint that is being pending will not have a chance to be polled even if it becomes ready.

This is fine most of the time, but if there is complex IO logic in the SVC's poll_ready (similar to a server-side active disconnection), the ready endpoint will not respond for a long time and will be disconnected. Of course, this can be compatible with this kind of case by implementing the corresponding logic in the SVC `poll_ready`, but this may not be the part of Tower that is of interest.

From my point of view, there are two solutions to fix this issue:
1. Tower handles the wake-up logic of multiple endpoint pending in buffer/worker or balance, and is not affected by `worker::p oll_next_msg`;
2. The user of the tower (i.e. tonic) identifies the error in the `State::Connecting` poll stage in the 'poll_ready' of svc (corresponding to the tonic should be `Reconnect`), here I have actually tested it and unfortunately it does not return a Err in the case, maybe it is difficult to do this;

PR #825, opened 2025-05-13
Adjust Worker::poll logic to fix pending wake failure in balance service.

c98:
#824 
This commit aims to solve the problem that `Worker::poll_next_msg` may block `service.poll_ready`. Especially when the service is a `Balance` service, while multiple pending endpoints become ready, the task is blocked on `Worker::poll_next_msg` if there is no other message come which would cause these endpoints to be disconnected.

### DIFF
--- a/tower/src/buffer/worker.rs
+++ b/tower/src/buffer/worker.rs
@@ -24,7 +24,6 @@ pin_project_lite::pin_project! {
     where
         T: Service<Request>,
     {
-        current_message: Option<Message<Request, T::Future>>,
         rx: mpsc::Receiver<Message<Request, T::Future>>,
         service: T,
         finish: bool,
@@ -53,7 +52,6 @@ where
         };
 
         let worker = Worker {
-            current_message: None,
             finish: false,
             failed: None,
             rx,
@@ -71,30 +69,18 @@ where
     fn poll_next_msg(
         &mut self,
         cx: &mut Context<'_>,
-    ) -> Poll<Option<(Message<Request, T::Future>, bool)>> {
+    ) -> Poll<Option<Message<Request, T::Future>>> {
         if self.finish {
             // We've already received None and are shutting down
             return Poll::Ready(None);
         }
 
         tracing::trace!("worker polling for next message");
-        if let Some(msg) = self.current_message.take() {
-            // If the oneshot sender is closed, then the receiver is dropped,
-            // and nobody cares about the response. If this is the case, we
-            // should continue to the next request.
-            if !msg.tx.is_closed() {
-                tracing::trace!("resuming buffered request");
-                return Poll::Ready(Some((msg, false)));
-            }
-
-            tracing::trace!("dropping cancelled buffered request");
-        }
-
         // Get the next request
         while let Some(msg) = ready!(Pin::new(&mut self.rx).poll_recv(cx)) {
             if !msg.tx.is_closed() {
                 tracing::trace!("processing new request");
-                return Poll::Ready(Some((msg, true)));
+                return Poll::Ready(Some(msg));
             }
             // Otherwise, request is canceled, so pop the next one.
             tracing::trace!("dropping cancelled request");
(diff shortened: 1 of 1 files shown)


seanmonstar:
Thanks! Looks like CI failed...

Also, is there a way to update the tests to check this change?

c98:
Thanks for your attention, I have updated the tests to check this change.

seanmonstar:
@cratelyn do yall have a more extensive suite that includes balancing that can test this change out?

cratelyn:
> @cratelyn do yall have a more extensive suite that includes balancing that can test this change out?

we do! i'll try running the linkerd2-proxy tests with this patch applied, pardon the wait.

tower/tests/buffer/main.rs
```diff
@@ -375,25 +351,6 @@ async fn wakes_pending_waiters_on_failure() {
...
-
-    let service = support::AssertSpanSvc::new(span.clone());
-    let (service, worker) = Buffer::pair(service, 5);
-    let worker = tokio::spawn(worker);
-
-    let result = tokio::spawn(service.oneshot(()).instrument(span));
-
-    result.await.expect("service panicked").expect("failed");
-    worker.await.expect("worker panicked");
-}
```
cratelyn, line 395: may i ask why this test was deleted?

tower/tests/buffer/main.rs
```diff
@@ -345,12 +327,6 @@ async fn wakes_pending_waiters_on_failure() {
         "response should fail with a ServiceError, got: {:?}",
         err
     );
-    let err = assert_ready_err!(response2.poll());
-    assert!(
-        err.is::<error::ServiceError>(),
-        "response should fail with a ServiceError, got: {:?}",
-        err
-    );
```
cratelyn, line 353: why is this no longer the case?

tower/tests/buffer/main.rs
```diff
@@ -216,16 +215,6 @@ async fn waits_for_channel_capacity() {
...
-
-    assert_pending!(worker.poll());
-    handle
-        .next_request()
-        .await
-        .unwrap()
-        .1
-        .send_response("world");
-    assert_pending!(worker.poll());
-    assert_ready_ok!(response4.poll());
```
cratelyn, line 228: why are these assertions no longer needed?

cratelyn (commented):
i left some notes below. my main questions and concerns about this change are about how many of the test suite's assertions (_and a test case_) are removed in this change.

if changes to the test suite are required here, is this a breaking change to this middleware's behavior?

cratelyn:
if this change is focused on solving how this middleware interacts with a `Balance` service, i might suggest adding a test case that exercises the buffer with a `Balance` service.

cratelyn:
> > @cratelyn do yall have a more extensive suite that includes balancing that can test this change out?
> 
> we do! i'll try running the linkerd2-proxy tests with this patch applied, pardon the wait.

@seanmonstar i wasn't able to get a patched version of the linkerd2-proxy building with this patch. we're on tonic v0.12 at the moment, which depends on tower 0.4, and thus builds fail if this branch is used.

i did leave some questions above, however. i'm concerned that trace contexts no longer seem to be propagated, and that this change seems to change the behavior of `Buffer`. having debugged changes in behavior introduced by other changes to `Buffer`, e.g. #635, i'm very hesitant about this pr.

tower/tests/buffer/main.rs
c98: because in this case there is no other request will be issued, so it is no need to do this assertion.

tower/tests/buffer/main.rs
c98: this commit change the worker exec order: 

1. service.poll_ready
2. poll_next_msg
3. service.call

after 2 poll_next_msg, the span in the msg can be retrieved, the service.call is now in the span scope, but not the servce.poll_ready, as you see, `AssertSpanSvc` does not fit the case, so i delete this test.

tower/tests/buffer/main.rs
c98: as mentioned above, the old exec order is 
1. poll_next_msg
2. service.poll_ready
3. service.call

the new exec order is
1. service.poll_ready
2. poll_next_msg
3. service.call

there is no 'preload' message, so i delete it.

PR #828, opened 2025-06-12
fix: use minimal tokio features in `make` and `reconnect` features

Icemic:
`make` or `reconnect` both features rely only on the basic tokio (tokio::io) and do not require the `io-std` feature, which instead leads to compilation errors on wasm32 target.

This PR changes it to the mininal `tokio/sync` to just enable `tokio`, improving compatibility with the wasm32 target.


tower/Cargo.toml
```diff
@@ -48,9 +48,9 @@ hedge = ["util", "filter", "futures-util", "hdrhistogram", "tokio/time", "tracin
 limit = ["tokio/time", "tokio/sync", "tokio-util", "tracing", "pin-project-lite"]
 load = ["tokio/time", "tracing", "pin-project-lite"]
 load-shed = ["pin-project-lite"]
-make = ["pin-project-lite", "tokio/io-std"]
+make = ["pin-project-lite", "tokio/sync"]
```
jplatte, line 51: I don't see any use of `tokio::sync` in `make/`, so I think we don't need this feature activation either.

```suggestion
make = ["pin-project-lite", "tokio"]
```

tower/Cargo.toml
```diff
@@ -48,9 +48,9 @@ hedge = ["util", "filter", "futures-util", "hdrhistogram", "tokio/time", "tracin
...
 ready-cache = ["futures-core", "futures-util", "indexmap", "tokio/sync", "tracing", "pin-project-lite"]
-reconnect = ["make", "tokio/io-std", "tracing"]
+reconnect = ["make", "tokio/sync", "tracing"]
```
jplatte, line 53: Same here, `reconnect` doesn't seem to use `tokio` so this seems like it can be removed entirely.

```suggestion
reconnect = ["make", "tracing"]
```

tower/Cargo.toml
jplatte: We seem to have `sync` as an enabled feature of `tokio` whenever that optional dependency is activated, further below. We should be able to drop that as well.

Icemic: Ah, indeed. Fixed.

jplatte:
You only partially implemented my suggestions. Here is what I meant:

```diff
--- a/tower/Cargo.toml
+++ b/tower/Cargo.toml
@@ -50,7 +50,7 @@ load = ["tokio/time", "tracing", "pin-project-lite"]
 load-shed = ["pin-project-lite"]
 make = ["pin-project-lite", "tokio"]
 ready-cache = ["futures-core", "futures-util", "indexmap", "tokio/sync", "tracing", "pin-project-lite"]
-reconnect = ["make", "tokio", "tracing"]
+reconnect = ["make", "tracing"]
 retry = ["tokio/time", "util"]
 spawn-ready = ["futures-util", "tokio/sync", "tokio/rt", "util", "tracing"]
 steer = []
@@ -66,7 +66,7 @@ futures-util = { workspace = true, features = ["alloc"], optional = true }
 hdrhistogram = { workspace = true, optional = true }
 indexmap = { workspace = true, optional = true }
 slab = { workspace = true, optional = true }
-tokio = { workspace = true, features = ["sync"], optional = true }
+tokio = { workspace = true, optional = true }
 tokio-stream = { workspace = true, optional = true }
 tokio-util = { workspace = true, optional = true }
 tracing = { workspace = true, features = ["std"], optional = true }
 ```

Icemic:
Done but I'm not sure if it is ok to remove `sync` from enabled features of tokio for all cases.

jplatte:
Thanks! I just realized this isn't merged yet. Not sure what happened there. GitHub is telling me there's merge conflicts and I can't resolve them myself because I'm not allowed to push to your branch. Mind rebasing or merging our `main` back into your branch?

PR #831, opened 2025-07-05
chore: Use tokio-stream UnboundedReceiverStream

tottoto:
tokio-stream's UnboundedReceiverStream could be used.


ripytide (approved):
Good simplification, and tests still pass.

PR #838, opened 2025-09-24
chore: Disable unused futures feature

tottoto:
Disables unused `futures` features.

### DIFF
--- a/tower-service/src/lib.rs
+++ b/tower-service/src/lib.rs
@@ -17,6 +17,9 @@
 
 extern crate alloc;
 
+#[cfg(doc)]
+extern crate std;
+
 use alloc::boxed::Box;
 
 use core::future::Future;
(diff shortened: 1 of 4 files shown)


tower-service/src/lib.rs
```diff
@@ -332,7 +331,7 @@ pub trait Service<Request> {
     /// If the service is at capacity, then `Poll::Pending` is returned and the task
     /// is notified when the service becomes ready again. This function is
     /// expected to be called while on a task. Generally, this can be done with
-    /// a simple `futures::future::poll_fn` call.
+    /// a simple [`core::future::poll_fn`] call.
```
jplatte: It may also be available in `core`, but I think usually you'd link to the `std` re-export.

```suggestion
    /// a simple [`std::future::poll_fn`] call.
```

PR #852, opened 2026-02-07
ci: Remove unused working-directory config

tottoto:
`working-directory` config doesn't seem to be used.


jplatte (approved):
Love how this was not an error from GHA previously 🤦🏼

PR #863, opened 2026-05-12
fix(retry): avoid truncating fractional retry_percent in TpsBudget

### LINKED ISSUE #857: retry: `TpsBudget` over-allocates retries for fractional `retry_percent < 1.0`
### Summary

`TpsBudget::new` truncates `1.0 / retry_percent` when `retry_percent <= 1.0`:

```rust
let (deposit_amount, withdraw_amount) = if retry_percent == 0.0 {
    (0, 1)
} else if retry_percent <= 1.0 {
    (1, (1.0 / retry_percent) as isize)
} else {
    (1000, (1000.0 / retry_percent) as isize)
};
```

For fractional values below `1.0` whose reciprocal is not an integer, this rounds the withdrawal cost down and permits **more retries than configured**.

For example, `retry_percent = 0.6` yields `(deposit_amount, withdraw_amount) = (1, 1)`, so every deposit authorizes one full retry (100%), not 0.6 retries (60%).

### Reproduction

```rust
use std::time::Duration;
use tower::retry::budget::{Budget, TpsBudget};

#[test]
fn fractional_retry_percent_below_one_overissues_budget() {
    let budget = TpsBudget::new(Duration::from_secs(1), 0, 0.6);

    for _ in 0..10 {
        budget.deposit();
    }

    let allowed = (0..10).filter(|_| budget.withdraw()).count();

    assert!(
        allowed <= 6,
        "10 deposits at retry_percent=0.6 should not allow more than 6 retries, got {allowed}"
    );
}
```

This silently over-allocates retry budget for many fractional values below `1.0`, which undermines the budget's role in limiting retry amplification during failures.

SAY-5:
`TpsBudget::new` used `(1, (1.0 / retry_percent) as isize)` for `retry_percent <= 1.0`, which truncates the reciprocal so e.g. `0.6` authorizes a full retry per deposit instead of 0.6. This reuses the existing 1000x deposit scaling for all positive percentages and adds a regression test. Closes #857.

### DIFF
--- a/tower/src/retry/budget/tps_budget.rs
+++ b/tower/src/retry/budget/tps_budget.rs
@@ -78,13 +78,12 @@ impl TpsBudget {
             // If there is no percent, then you gain nothing from deposits.
             // Withdrawals can only be made against the reserve, over time.
             (0, 1)
-        } else if retry_percent <= 1.0 {
-            (1, (1.0 / retry_percent) as isize)
-        } else {
-            // Support for when retry_percent is between 1.0 and 1000.0,
-            // meaning for every deposit D, D * retry_percent withdrawals
-            // can be made.
+        } else if retry_percent <= 0.5 {
+            // Small fractional percents lose precision when 1/retry_percent is
+            // truncated to an isize, so scale deposits by 1000 here.
             (1000, (1000.0 / retry_percent) as isize)
+        } else {
+            (1, (1.0 / retry_percent) as isize)
         };
         let reserve = (min_per_sec as isize)
             .saturating_mul(ttl.as_secs() as isize) // ttl is between 1 and 60 seconds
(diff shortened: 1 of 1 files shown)


tower/src/retry/budget/tps_budget.rs
```diff
@@ -78,12 +78,9 @@ impl TpsBudget {
             // If there is no percent, then you gain nothing from deposits.
             // Withdrawals can only be made against the reserve, over time.
             (0, 1)
-        } else if retry_percent <= 1.0 {
```
seanmonstar, line 81: I think most often the percent will be 10, 20, around there. So, to not need to deal with large numbers in the most common case, this could probably be edit to just `<= 0.5`. What do you think?

PR #867, opened 2026-05-21
fix(buffer): assert non-zero bound on construction

### LINKED ISSUE #861: buffer: `bound = 0` panics in `Buffer::new` / `Buffer::pair` (and via `BufferLayer::new`)
## Summary

`Buffer::new`, `Buffer::pair`, and `BufferLayer::new` accept `bound: usize`, but `0` is not actually supported. `Buffer::pair` forwards `bound` directly to `tokio::sync::mpsc::channel(bound)`, which panics when the capacity is `0`.

Because the public Tower docs describe `bound` as the maximum number of queued requests before backpressure, but do not document `0` as invalid, this currently turns an apparently-valid configuration into an immediate panic.

## Reproduction

```rust
use std::{
    convert::Infallible,
    future::{ready, Ready},
    task::{Context, Poll},
};
use tower::{buffer::Buffer, Service};

#[derive(Clone)]
struct Svc;

impl Service<()> for Svc {
    type Response = ();
    type Error = Infallible;
    type Future = Ready<Result<(), Infallible>>;

    fn poll_ready(&mut self, _cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
        Poll::Ready(Ok(()))
    }

    fn call(&mut self, _req: ()) -> Self::Future {
        ready(Ok(()))
    }
}

fn main() {
    let _ = Buffer::pair(Svc, 0);
}
```

This can crash applications during startup or reconfiguration when the buffer size is computed dynamically or loaded from configuration and reaches zero. But in practice, the chance that this can cause any denial-of-service (DoS) attack is almost zero, so it is not a security concern, more like a surprise.

Alpy16:
Closes #861

Description
Initializing Buffer or BufferLayer with a capacity bound of 0 creates an invalid configuration state. This bypasses initial validation and causes an unrecoverable runtime panic down the line when the underlying mpsc::channel allocation is executed.

This patch adds a defensive assert!(bound > 0, ...) boundary guard to both the BufferLayer::new and Buffer::pair constructors. This shifts a latent runtime crash into a predictable, fail-fast initialization panic.

Type of Change
Bug fix (non-breaking change enforcing defensive API boundaries)

Verification
Added #[should_panic] integration tests to tower/tests/buffer/main.rs to verify that both direct instantiation and layer application fail safely when configured with a bound of 0.

Verified workspace tests pass via cargo test --test buffer --all-features.

### DIFF
--- a/tower/src/buffer/layer.rs
+++ b/tower/src/buffer/layer.rs
@@ -34,6 +34,7 @@ impl<Request> BufferLayer<Request> {
     /// [`call`]: crate::Service::call
     /// [`poll_ready`]: crate::Service::poll_ready
     pub const fn new(bound: usize) -> Self {
+        assert!(bound > 0, "buffer bound must be greater than zero");
         BufferLayer {
             bound,
             _p: PhantomData,
--- a/tower/src/buffer/layer.rs
+++ b/tower/src/buffer/layer.rs
@@ -20,6 +20,12 @@ impl<Request> BufferLayer<Request> {
     /// `bound` gives the maximal number of requests that can be queued for the service before
     /// backpressure is applied to callers.
     ///
+    ///
+    /// # Panics
+    ///
+    /// Panics if `bound` is zero.
+    ///
+    ///
     /// # A note on choosing a `bound`
     ///
     /// When [`Buffer`]'s implementation of [`poll_ready`] returns [`Poll::Ready`], it reserves a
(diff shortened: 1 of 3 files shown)


seanmonstar:
Probably the docs on the constructor should also mention that the value must be greater than 0, maybe in a `# Panics` section?

Alpy16:
I'll add the # Panics section to the documentation for both constructors right now and get those pushed up. Thanks for the review

PR #869, opened 2026-05-21
chore: improve tower maintenance path

oab24413gmai:
Summary:
- Add unit tests for the Timeout middleware covering the elapsed-error branch: verify that when the inner future exceeds the configured duration, the service returns an Elapsed error (downcastable via the boxed error), and that requests completing before the deadline pass through unchanged. The timeout module files are small and self-contained; tests can be added inline with #[cfg(test)] using tokio::time::pause for deterministic timing.
- Keep the change narrow so it is straightforward to review.

Notes:
- I kept this scoped to the relevant implementation and tests.

### DIFF
--- a/tower/src/timeout/mod.rs
+++ b/tower/src/timeout/mod.rs
@@ -68,3 +68,72 @@ where
         ResponseFuture::new(response, sleep)
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use std::{
+        convert::Infallible,
+        future::Future,
+        pin::Pin,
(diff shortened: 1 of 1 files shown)


jlizen (changes requested):
I'm happy to have some regression test coverage for the timeout layer, but let's update it to match the structure of other tests in `tower/tests/<layer>/main.rs`.

Those use `tower_test::mock` and `tokio_test` macros, check `poll_ready`, and various other things. I would expect the timeout tests to look very similar.

PR #874, opened 2026-06-17
fix(balance): clear cached P2C ready index after a discovery removal

### LINKED ISSUE #856: balance/p2c: stale cached ready index after discovery removal can route requests to a different endpoint
## Summary
`p2c::Balance` caches the selected ready endpoint as `ready_index: Option<usize>`. If `poll_ready()` is called again before `call()`, it reuses that cached numeric slot after first processing discovery updates.

However, `update_pending_from_discover()` handles `Change::Remove(key)` by calling `ReadyCache::evict(&key)`, and `ReadyCache::evict()` removes from the ready set with `swap_remove_full`. That can change which service lives at the cached index.

`poll_ready()` then calls `check_ready_index()` on the stale index, which validates whichever service currently occupies that slot, not the endpoint originally chosen by P2C. `call()` later forwards the request to that swapped-in service.

## Reproduction
A deterministic way to reproduce this:

1. Construct a `p2c::Balance` with a deterministic RNG and a discover stream.
2. Insert three ready endpoints:
   - `a` with load `0`
   - `b` with load `1`
   - `c` with load `10`
3. Call `poll_ready()` once so that P2C selects `b` and caches its slot in `ready_index`.
4. Before calling `call()`, emit `Change::Remove("b")` from discovery.
5. Call `poll_ready()` again.
6. `ReadyCache::evict("b")` performs a `swap_remove_full`, so `c` is moved into the slot previously occupied by `b`.
7. `Balance::poll_ready()` reuses the old index, `check_ready_index()` succeeds for `c`, and `call()` sends the request to `c`.

With the same deterministic RNG, a fresh P2C selection over the remaining ready set `[a, c]` would choose `a`, so routing to `c` demonstrates stale index reuse rather than a fresh selection.

This does not send traffic to a removed endpoint, but it can silently bypass the balancer's intended P2C choice during endpoint churn.

## Relevant code
```rust
// tower/src/balance/p2c/service.rs
Some(Change::Remove(key)) => {
    self.services.evict(&key);
}

if let Some(index) = self.ready_index.take() {
    match self.services.check_ready_index(cx, index) {
        Ok(true) => {
            self.ready_index = Some(index);
            return Poll::Ready(Ok(()));
        }
        // ...
    }
}

fn call(&mut self, request: Req) -> Self::Future {
    let index = self.ready_index.take().expect("called before ready");
    self.services.call_ready_index(index, request)
}
```

```rust
// tower/src/ready_cache/cache.rs
/// [`ReadyCache::poll_pending`] and [`ReadyCache::evict`] may perturb the order of
/// the ready set, so any cached indexes should be discarded after such a call.

pub fn evict<Q: Hash + Equivalent<K>>(&mut self, key: &Q) -> bool {
    // ...
    self.ready.swap_remove_full(key)
}
```

## Suggested fix
At minimum, clear `ready_index` whenever `update_pending_from_discover()` processes a removal that calls `evict()`.

A more robust fix would be to cache the selected key instead of the selected index (or cache both and verify the key still matches before `check_ready_index()` / `call_ready_index()`). That avoids coupling correctness to ready-set ordering.

ameyypawar:
## Summary

Fixes #856.

`p2c::Balance` caches the P2C-selected endpoint as a numeric `ready_index`. When a subsequent `poll_ready` processes a `Change::Remove`, `update_pending_from_discover` calls `ReadyCache::evict`, which removes from the ready set with `swap_remove_full`. That can move a *different* endpoint into the slot the cached `ready_index` points at. `poll_ready` then revalidates the stale index with `check_ready_index` (which succeeds for whatever now occupies that slot) and `call` dispatches the request there — silently bypassing the P2C choice during endpoint churn.

This matches the warning already in `ready_cache`'s own docs:

> calls to `ReadyCache::poll_pending` and `ReadyCache::evict` may perturb the order of the ready set, so any cached indexes should be discarded after such a call.

## Fix

Clear `ready_index` whenever a removal evicts a service, so the next `poll_ready` re-runs P2C over the current ready set:

```rust
Some(Change::Remove(key)) => {
    trace!("remove");
    self.services.evict(&key);
    // `evict` removes from the ready set with `swap_remove`, which can move a
    // different endpoint into the slot a previously-selected `ready_index`
    // points at. Discard the cached selection so `poll_ready` re-runs P2C.
    self.ready_index = None;
}
```

I also corrected the `poll_ready` comment that claimed *"These updates cannot disturb the order of existing ready services"* — they can, via `swap_remove`.

## Why this minimal fix is sufficient

The only operation that reorders the *existing* ready set between caching `ready_index` and reusing it is `evict`'s `swap_remove_full`:

- `ReadyCache::poll_pending` only `insert`s into the ready `IndexMap` (appends new keys / updates existing in place — existing indices stay put).
- `Change::Insert` pushes to the *pending* set, never the ready set.
- `check_ready_index`'s not-ready/error arms and `call_ready_index` also `swap_remove`, but only *after* `ready_index` has been `.take()`n within the same `poll_ready` loop, each immediately followed by a fresh P2C selection — so no stale index survives across calls.

The issue notes a more robust alternative (cache the key instead of the index); that's a larger change and isn't required to close this bug, but happy to go that route if you'd prefer it.

## Tests

Not included in this commit. A deterministic regression test needs a dynamic `Discover` (to emit `Change::Remove` between `poll_ready` calls) plus a seeded `Rng`, which the current `p2c` tests don't have scaffolding for. Glad to follow up with that test (building the small test `Discover` + mock `Rng`) if you'd like it in this PR.

### DIFF
--- a/tower/src/balance/p2c/service.rs
+++ b/tower/src/balance/p2c/service.rs
@@ -116,6 +116,13 @@ where
                 Some(Change::Remove(key)) => {
                     trace!("remove");
                     self.services.evict(&key);
+                    // `evict` removes from the ready set with `swap_remove`,
+                    // which can move a different endpoint into the slot a
+                    // previously-selected `ready_index` points at. Discard the
+                    // cached selection so `poll_ready` re-runs P2C over the
+                    // current ready set instead of dispatching to the
+                    // swapped-in endpoint.
+                    self.ready_index = None;
                 }
                 Some(Change::Insert(key, svc)) => {
                     trace!("insert");
@@ -206,8 +213,11 @@ where
     >;
 
     fn poll_ready(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
-        // `ready_index` may have already been set by a prior invocation. These
-        // updates cannot disturb the order of existing ready services.
+        // `ready_index` may have already been set by a prior invocation. If a
+        // discovered removal evicts a service, `update_pending_from_discover`
+        // clears it, since eviction can reorder the ready set (via
+        // `swap_remove`) and leave the cached index pointing at a different
+        // endpoint.
         let _ = self.update_pending_from_discover(cx)?;
         self.promote_pending_to_ready(cx);
 
--- a/tower/src/balance/p2c/test.rs
+++ b/tower/src/balance/p2c/test.rs
@@ -193,11 +193,17 @@ async fn ready_index_reset_after_removal() {
     // Promote endpoints one per poll so the ready set is deterministically
     // ordered [a@0, b@1, c@2]. P2C selects `a` first (the only ready service)
     // and keeps it cached while it stays ready.
-    changes.borrow_mut().push_back(Change::Insert(0, endpoint(0, 0)));
+    changes
+        .borrow_mut()
+        .push_back(Change::Insert(0, endpoint(0, 0)));
     assert_ready_ok!(svc.poll_ready());
-    changes.borrow_mut().push_back(Change::Insert(1, endpoint(1, 1)));
+    changes
+        .borrow_mut()
+        .push_back(Change::Insert(1, endpoint(1, 1)));
     assert_ready_ok!(svc.poll_ready());
-    changes.borrow_mut().push_back(Change::Insert(2, endpoint(2, 10)));
+    changes
+        .borrow_mut()
+        .push_back(Change::Insert(2, endpoint(2, 10)));
     assert_ready_ok!(svc.poll_ready());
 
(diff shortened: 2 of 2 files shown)


seanmonstar (approved):
Logically this makes sense to me, thanks!

@cratelyn any desire to try this on linkerd before I merge? Not expected, just pausing briefly in case.

cratelyn:
> Logically this makes sense to me, thanks!
> 
> @cratelyn any desire to try this on linkerd before I merge? Not expected, just pausing briefly in case.

i am alright, but thank you for asking!

